PR #323, opened 2020-02-16
[WIP] Import migen's genlib.misc into nmigen.compat

awygle:


awygle:
Need to figure out whether to add corresponding functionality to nmigen, and add depreciation warnings, for all items in genlib.misc.

awygle:
I'm gonna chop this up into much smaller bits at some point, this seems unlikely to be mergeable as is.

Issue #328, opened 2020-03-05
Add `r_level` and `w_level` to AsyncFIFO


awygle:
It's often useful to have a "high watermark/low watermark" signal from a FIFO, particularly when processing any kind of fixed-length frames. This could be implemented as a wrapper around `SyncFIFO` pretty easily but probably wants to be more deeply integrated with `AsyncFIFO` since there's already a clock-domain-crossing counter to use (which is not currently exposed to consumers).

whitequark:
I'm not sure how an async FIFO with water marks works, CDC-wise. Do you have any references?

awygle:
As section 5.6 of http://www.sunburst-design.com/papers/CummingsSNUG2002SJ_FIFO1.pdf says, you compare against (ptr+N) instead of (ptr) as you would for empty/full. That section appears to be discussing FIFO style 1, but the same technique should be adaptable to FIFO style 2 with minor modifications.

whitequark:
That means we have to Gray decode the resynchronized counters, right? Then it seems like we should rather have `r_level` and `w_level` for consistency with `SyncFIFO`, and the consumer of the FIFO interface can then write a comparator with HWM/LWM. We can even make it a part of the general FIFO interface, with the SyncFIFO case aliasing `level` as both.

awygle:
That makes sense, then we can always create a HWM/LWM wrapper as a utility module if we want to and it can work with any `FIFOInterface` instance.

whitequark:
In particular I think having a level is strictly more useful than watermarks (even configurable watermarks), if only for debugging. It's true that comparators are cheaper on architectures which have lookahead carry primitives, but I would say that it's the toolchain's job to simplify subtract-then-compare, since that's a fairly simple local transformation.

awygle:
I'll be honest, I missed `level` on SyncFIFO when I was looking at the existing FIFO code 😅

whitequark:
Fixed in https://github.com/nmigen/nmigen/pull/472, https://github.com/nmigen/nmigen/pull/512, https://github.com/nmigen/nmigen/pull/513, https://github.com/nmigen/nmigen/pull/524.

Issue #185, opened 2019-08-22
ASIC support tracking issue


nmigen-issue-migration:
<a href="https://github.com/whitequark"><img src="https://avatars1.githubusercontent.com/u/54771?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Thursday Aug 22, 2019 at 21:49 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/184_

----

A number of people have expressed a desire to use nMigen for ASIC design. This issue tracks missing aspects of that workflow.

 - [x] On FPGAs, asynchronous resets inhibit inference of BRAMs, DSPs, etc. On ASICs, asynchronous resets are required to make sure the chip is reset properly even with no clock.
 - [ ] On FPGAs, resets are generally positive polarity, and using the wrong polarity can have severe consequences for timing. On ASICs, resets are generally negative polarity, such that the chip is held in reset during power rail ramp-up.
 - [ ] On ASICs and in some cases on FPGAs, memories may not be initialized.
 - [ ] On ASICs, `reset_less` signals are not initialized at power-on.

(There are likely more.)

whitequark:
@Fatsie you're interested in active low resets for ASIC work, right? I was wondering if you have any opinion on handling `ResetSignal` in that case; unlike `ClockSignal`, a lot of generic code depends on `ResetSignal` polarity being logic level (i.e. active high), so if we add clock domains with negative polarity resets that have inverted levels of `ResetSignal`, all such logic will break. There are a few solutions I can think of, but none are ideal.

Fatsie:
@whitequark I don't think it is a problem to have nmigen only have active high reset support in it's clock domain.
For ASICs you typically want the external reset pin active low so you can keep the reset low during power ramp up by a simple RC circuit. But you can handle that by PinsN definition for the platform. If the standard cell library would have flip-flops with active low set and reset signals yosys should optimize away the two inverters in the logic path of the set/reset signal.

hofstee:
I've seen lots of inverters get added to a design since tools will often synthesize things to use active-low resets (sometimes due to what's available in technology libraries). That ends up leading to a bunch of extra inverters in the design, especially at black-box boundaries where you've synthesized something to be used as a hard macro elsewhere (back to back inverters in timing reports confused the heck out of me for a while).

I can't say to what degree it actually affected the quality of the design, I don't remember the numbers at the moment.

[...]

PR #329, opened 2020-03-05
Implement solution to AsyncFIFO resets, as in #328

awygle:
Mark both read domain counters as reset-less, then synchronize write
domain reset into read domain, On write domain reset, force empty flag
and set read counters == write counters.

---

Marking this as WIP because the reset functionality isn't really tested yet. Interested in thoughts on  how that might be done.

### DIFF
--- a/nmigen/hdl/cd.py
+++ b/nmigen/hdl/cd.py
@@ -21,7 +21,7 @@ class ClockDomain:
         If ``True``, the domain does not use a reset signal. Registers within this domain are
         still all initialized to their reset state once, e.g. through Verilog `"initial"`
         statements.
-    clock_edge : str
+    clk_edge : str
         The edge of the clock signal on which signals are sampled. Must be one of "pos" or "neg".
     async_reset : bool
         If ``True``, the domain uses an asynchronous reset, and registers within this domain
--- a/nmigen/hdl/cd.py
+++ b/nmigen/hdl/cd.py
@@ -21,6 +21,8 @@ class ClockDomain:
         If ``True``, the domain does not use a reset signal. Registers within this domain are
         still all initialized to their reset state once, e.g. through Verilog `"initial"`
         statements.
+    clock_edge : str
+        The edge of the clock signal on which signals are sampled. Must be one of "pos" or "neg".
     async_reset : bool
         If ``True``, the domain uses an asynchronous reset, and registers within this domain
         are initialized to their reset state when reset level changes. Otherwise, registers
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -321,7 +321,7 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         m.d.comb += produce_w_nxt.eq(produce_w_bin + do_write)
         m.d[self._w_domain] += produce_w_bin.eq(produce_w_nxt)
 
-        # Note: Both read-domain counters must be reset_less
+        # Note: Both read-domain counters must be reset_less (see comments below)
         consume_r_bin = Signal(self._ctr_bits, reset_less=True)
         consume_r_nxt = Signal(self._ctr_bits)
         m.d.comb += consume_r_nxt.eq(consume_r_bin + do_read)
@@ -372,13 +372,22 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         ]
 
         # Reset handling to maintain FIFO and CDC invariants in the presence of a write-domain
-        # reset. For further discussion, see https://github.com/nmigen/nmigen/issues/181.
+        # reset.
+        # There is a CDC hazard associated with resetting an async FIFO - Gray code counters which
+        # are reset to 0 violate their Gray code invariant. One way to handle this is to ensure
+        # that both sides of the FIFO are asynchronously reset by the same signal. We adopt a
+        # slight variation on this approach - reset control rests entirely with the write domain.
+        # The write domain's reset signal is used to asynchronously reset the read domain's
+        # counters and force the FIFO to be empty when the write domain's reset is asserted.
+        # This requires the two read domain counters to be marked as "reset_less", as they are
+        # reset through another mechanism. See https://github.com/nmigen/nmigen/issues/181 for the
+        # full discussion.
         w_rst = ResetSignal(domain=self._w_domain, allow_reset_less=True)
         r_rst = Signal()
         
         # Async-set-sync-release synchronizer avoids CDC hazards
         rst_cdc = m.submodules.rst_cdc = \
-            AsyncFFSynchronizer(w_rst, r_rst, domain="rst_cdc")
+            AsyncFFSynchronizer(w_rst, r_rst, domain=self._r_domain)
 
         # Decode Gray code counter synchronized from write domain to overwrite binary
         # counter in read domain.
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -3,8 +3,8 @@
 from .. import *
 from ..asserts import *
 from .._utils import log2_int, deprecated
-from .coding import GrayEncoder
-from .cdc import FFSynchronizer
+from .coding import GrayEncoder, GrayDecoder
+from .cdc import FFSynchronizer, ResetSynchronizer
 
 
 __all__ = ["FIFOInterface", "SyncFIFO", "SyncFIFOBuffered", "AsyncFIFO", "AsyncFIFOBuffered"]
@@ -317,7 +317,8 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         m.d.comb += produce_w_nxt.eq(produce_w_bin + do_write)
         m.d[self._w_domain] += produce_w_bin.eq(produce_w_nxt)
 
-        consume_r_bin = Signal(self._ctr_bits)
+        # Note: Both read-domain counters must be reset_less
+        consume_r_bin = Signal(self._ctr_bits, reset_less=True)
         consume_r_nxt = Signal(self._ctr_bits)
         m.d.comb += consume_r_nxt.eq(consume_r_bin + do_read)
         m.d[self._r_domain] += consume_r_bin.eq(consume_r_nxt)
@@ -331,7 +332,7 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         m.d.comb += produce_enc.i.eq(produce_w_nxt),
         m.d[self._w_domain] += produce_w_gry.eq(produce_enc.o)
 
-        consume_r_gry = Signal(self._ctr_bits)
+        consume_r_gry = Signal(self._ctr_bits, reset_less = True)
         consume_w_gry = Signal(self._ctr_bits)
         consume_enc = m.submodules.consume_enc = \
             GrayEncoder(self._ctr_bits)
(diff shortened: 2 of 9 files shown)


nmigen/lib/fifo.py
```diff
@@ -366,6 +367,31 @@ def elaborate(self, platform):
             self.r_rdy.eq(~r_empty),
         ]
 
+        # Reset handling to maintain FIFO and CDC invariants in the presence of a write-domain
+        # reset. For further discussion, see https://github.com/nmigen/nmigen/issues/181.
+        w_rst = ResetSignal(domain=self._w_domain, allow_reset_less=True)
+        r_rst = Signal()
+        
+        # Create synthetic clock domain to work around ResetSynchronizer expecting a domain
+        fake_domain = ClockDomain(name="rst_cdc", async_reset=True, local=True)
```
whitequark: I think this characterization isn't correct. You are adding some FFs with a different control set (one that includes w_rst), which in fact form a (very small) new clock domain. This doesn't inherently have anything to do with the way `ResetSynchronizer` works; nMigen requires you to create a clock domain for every distinct async control set.

nmigen/lib/fifo.py
```diff
@@ -366,6 +367,31 @@ def elaborate(self, platform):
...
+        m.domains += fake_domain
+        rst_cdc = m.submodules.rst_cdc = \
+                ResetSynchronizer(w_rst, domain = "rst_cdc")
+        m.d.comb += [
+                fake_domain.clk.eq(ClockSignal(domain = self._r_domain)),
+                r_rst.eq(ResetSignal(domain = "rst_cdc", allow_reset_less=True)),
```
whitequark: The domain you create here is never reset-less.

nmigen/lib/fifo.py
awygle: But `ResetSynchronizer` creates such a clock domain internally (right?), so this one is unnecessary.

whitequark:
(There's a few style nits I skipped for now.)

> Interested in thoughts on how that might be done.

I don't have any especially good ideas. Like you I find testing HDL (especially multiclock) to be pretty frustrating and I don't have silver bullets.

nmigen/lib/fifo.py
whitequark: You're totally right. I was thinking of `ResetSynchronizer` delegating to platform code, in which case it generally won't use an nMigen clock domain, but that's a different problem.

Do you feel like fixing it?

awygle: Happy to but the right solution wasn't obvious to me, other than making something like a `AsyncAssertSyncDeassertSynchronizer` module which `ResetSynchronizer` would use under the hood, and I didn't want to try to create a new cdc module without checking with you.

awygle:
Are you willing to merge this without tests of that functionality? If not, it seems like the most logical thing would be to add some properties to `FIFOContractSpec`, which would also hit the Sync cases, so I'd have to spend time understanding that before I could add the necessary asserts. Totally willing to, it'll just delay this landing (which is also fine with me).

nmigen/lib/fifo.py
whitequark: What about (a) adding a `rst` attribute (output), alongside `arst` (input)`, and allowing (b) passing `domain=None` to the constructor? Then the platform code would assert `reset_sync.rst`, which in the case when domain is specified will be assigned to the domain.

awygle: That should work. It seems a bit more complex to me, and it doesn't address a case where you might want to use this functionality for something besides a reset (admittedly I don't know what use case that would be), but it's a nice way to enhance `ResetSynchronizer` without (I think) having to touch all the platform code.

I lean towards creating the module I described (a more reasonable name might be `ReleaseSynchronizer`), but I'm open to the solution you described as well. I just think a generic module that gets replaced with platform-specific modules if available is a cleaner mental model.

whitequark: > without (I think) having to touch all the platform code.

You have to touch all the platform code with my suggestion.

> I just think a generic module that gets replaced with platform-specific modules if available is a cleaner mental model.

I'm a bit confused here. `ResetSynchronizer` *does* get replaced with platform-specific modules; that's half the reason for its existence and being in `nmigen.lib`.

If you mean that you want to make `ResetSynchronizer` a thin wrapper around some other module that just does `ResetSignal(self._domain).eq(...)` then I'm not really opposed to it, but it seems like the overhead (having two modules that do *almost* the same thing, duplicate constructor arguments, documentation, etc) is not worth the benefit (maybe slightly cleaner separation of concerns).

> admittedly I don't know what use case that would be

Yes. "Chain of FFs with asynchronous preset and synchronous clear" is a very specific use case. Also, sometime in the future we'll have to extend it to handle negative polarity resets (#185).

whitequark:
> Totally willing to, it'll just delay this landing (which is also fine with me).

There's no rush to support this, so I would prefer if it landed with tests.

nmigen/lib/fifo.py
awygle: Let me just push what I'm talking about and we can discuss it directly instead of each talking about the picture in our head of the picture in the other person's head :) can always roll back.

awygle:
In the process of trying to write what I was describing I realized it wouldn't work. I'm not sure what you described works either, because you still need a domain in order to get at the clock for the synchronous release. So I added a `reset_less` parameter (~which... I forgot to document, oops~) whose only function is to prevent hooking up the output of the synchronizer to the clock domain's actual reset. I also added a `self.rst` signal which gets assigned to the value of the output regardless of the value of `reset_less` (~which I need to hook up in the platform case, and document, oops again~). This makes the change backwards-compatible, since it'll just get ignored by all existing users. How does this look to you?

whitequark:
> I'm not sure what you described works either, because you still need a domain in order to get at the clock for the synchronous release.

Argh, you're right, it doesn't. But I still think it's really confusing to have it called `reset_less`. I feel like we need a slightly different approach here then.

whitequark:
What about adding something like `AsyncFFSynchronizer` that has a `clk`, `i`, `o` ports and a `async_edge` parameter that can be `"pos"` or `"neg"`, like in a ClockDomain itself? That would serve the needs of `AsyncFIFO` here and we could make `ResetSynchronizer` a thin wrapper around it.

I did say earlier that adding a new entity here has a cost, but it seems like overloading the functionality of `ResetSynchronizer` like this has an even higher cost.

awygle:
That's basically what I did except I still took a `domain` instead of a `clk` port, I didn't implement `async_edge`, and I called it `ReleaseSynchronizer`. (I didn't see your comment until I was almost done making this change, hence the naming differences)

whitequark:
> I do think it should remain a `domain` though, for consistency with `FFSynchronizer` and friends.

Yes, that make sense. Could you do it as a part of another PR, please?

awygle:
Sure. You want me to pull all the `ResetSynchronizer` changes out from this PR into a separate one, yes?

whitequark:
Sure.

nmigen/hdl/cd.py
```diff
@@ -21,6 +21,8 @@ class ClockDomain:
         If ``True``, the domain does not use a reset signal. Registers within this domain are
         still all initialized to their reset state once, e.g. through Verilog `"initial"`
         statements.
+    clock_edge : str
```
whitequark: The actual parameter is called `clk_edge`.

whitequark: (But maybe it should be called `clock_edge` instead? We have a `rst` signal and `reset_less` parameter.)

awygle: derp. i fix. that's what i get for adding extras :p

Reset polarity could be `rst_edge` for consistency... but it's not really an edge-active signal, so `rst_pol`?

whitequark: I was thinking `rst_active="pos"`/`="neg"`, since resets can be both synchronous and asynchronous.

nmigen/lib/fifo.py
```diff
@@ -331,7 +332,7 @@ def elaborate(self, platform):
         m.d.comb += produce_enc.i.eq(produce_w_nxt),
         m.d[self._w_domain] += produce_w_gry.eq(produce_enc.o)
 
-        consume_r_gry = Signal(self._ctr_bits)
+        consume_r_gry = Signal(self._ctr_bits, reset_less = True)
```
whitequark:
```suggestion
        consume_r_gry = Signal(self._ctr_bits, reset_less=True)
```

nmigen/lib/fifo.py
```diff
@@ -366,6 +367,25 @@ def elaborate(self, platform):
...
         ]
 
+        # Reset handling to maintain FIFO and CDC invariants in the presence of a write-domain
+        # reset. For further discussion, see https://github.com/nmigen/nmigen/issues/181.
+        w_rst = ResetSignal(domain=self._w_domain, allow_reset_less=True)
+        r_rst = Signal()
+        
+        # Async-set-sync-release synchronizer avoids CDC hazards
+        rst_cdc = m.submodules.rst_cdc = \
+                AsyncFFSynchronizer(w_rst, r_rst, domain = "rst_cdc")
```
whitequark:
```suggestion
            AsyncFFSynchronizer(w_rst, r_rst, domain="rst_cdc")
```

awygle:
Oh oops, I didn't realize what Github was asking me - I made the requested formatting changes in the new commit.

nmigen/hdl/cd.py
awygle: Sure, that's reasonable. I don't have too strong of opinions on this subject.

whitequark:
> Oh oops, I didn't realize what Github was asking me - I made the requested formatting changes in the new commit.

I've never used it before. Wasn't sure if it was more polite to use that or just edit the PR; the latter seems like it'd be a problem if you force pushed, but is less busywork...

nmigen/lib/fifo.py
```diff
@@ -366,6 +371,25 @@ def elaborate(self, platform):
...
         ]
 
+        # Reset handling to maintain FIFO and CDC invariants in the presence of a write-domain
+        # reset. For further discussion, see https://github.com/nmigen/nmigen/issues/181.
+        w_rst = ResetSignal(domain=self._w_domain, allow_reset_less=True)
+        r_rst = Signal()
+        
+        # Async-set-sync-release synchronizer avoids CDC hazards
+        rst_cdc = m.submodules.rst_cdc = \
+            AsyncFFSynchronizer(w_rst, r_rst, domain="rst_cdc")
```
whitequark: Shouldn't this be:
```suggestion
            AsyncFFSynchronizer(w_rst, r_rst, domain=self._r_domain)
```

nmigen/lib/fifo.py
```diff
@@ -317,7 +321,8 @@ def elaborate(self, platform):
         m.d.comb += produce_w_nxt.eq(produce_w_bin + do_write)
         m.d[self._w_domain] += produce_w_bin.eq(produce_w_nxt)
 
-        consume_r_bin = Signal(self._ctr_bits)
+        # Note: Both read-domain counters must be reset_less
```
whitequark: Could you expand this to explain *why* the counters must be `reset_less`?

nmigen/lib/fifo.py
```diff
@@ -366,6 +371,25 @@ def elaborate(self, platform):
             self.r_rdy.eq(~r_empty),
         ]
 
+        # Reset handling to maintain FIFO and CDC invariants in the presence of a write-domain
+        # reset. For further discussion, see https://github.com/nmigen/nmigen/issues/181.
```
whitequark: Do you think you could summarize it here? You have all the relevant context right now, whereas that issue is fairly long and meandering. In cases like this I normally describe the problem, the possible solution, and the rationale for choosing the one I implemented, in a comment near the code. This way, anyone who reads the source will know exactly why it works the way it does, without having to parse an entire discussion.

whitequark:
Thanks!

PR #330, opened 2020-03-05
hdl.ir: Pick up the variable name as a default name for submodules

galibert:
The current generated rtlil uses name like U$$<number> for module names, which is not good for discoverability, implementing debuggers, etc.  This patch picks up the variable name as default name for modules, in a similar way than is done for Signals.

I'm not sure what the rules for unicity are, if any.  Maybe the name should include the hierarchy in it to avoid collisions.  I'm not sure.

### DIFF
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -9,7 +9,7 @@ from .._utils import *
 from .._unused import *
 from .ast import *
 from .cd import *
-
+from .. import tracer
 
 __all__ = ["UnusedElaboratable", "Elaboratable", "DriverConflict", "Fragment", "Instance"]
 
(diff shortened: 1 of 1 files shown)


whitequark:
Unfortunately the current behavior exists for a reason: module names are semantically meaningful, and having two modules with the same name is not legal. Also, I think it's important that `Elaboratable` involve as little magic as possible, and your solution makes it considerably more involved.

Have you considered, instead of using:
```python
submod = Submodule()
m.submodules += submod
```
doing:
```python
m.submodules.submod = submod = Submodule()
```
That's the suggested syntax and it even has the same amount of typing.

galibert:
You mean that submodules should not be instanciated in __init__ ?

whitequark:
They can be if it makes sense, then you'd do:
```python
m.submodules.submod = self.submod
```
later.

PR #332, opened 2020-03-06
Create AsyncFFSynchronizer CDC Primitive

awygle:
Per discussion on PR #329, implement `AsyncFFSynchronizer` CDC primitive for async-set-sync-release synchronization and make ResetSynchronizer a wrapper around it if platform-specific support does not exist.

### DIFF
--- a/nmigen/lib/cdc.py
+++ b/nmigen/lib/cdc.py
@@ -115,7 +115,7 @@ class AsyncFFSynchronizer(Elaboratable):
         Number of synchronization stages between input and output. The lowest safe number is 2,
         with higher numbers reducing MTBF further, at the cost of increased deassertion latency.
     """
-    def __init__(self, i, o, *, domain="sync", stages=2, async_edge="pos"):
+    def __init__(self, i, o, *, domain="sync", stages=2, async_edge="pos", max_input_delay=None):
         _check_stages(stages)
 
         self.i = i
@@ -129,21 +129,31 @@ class AsyncFFSynchronizer(Elaboratable):
                              .format(async_edge))
         self._edge = async_edge
 
+        self._max_input_delay = max_input_delay
+
     def elaborate(self, platform):
+        if hasattr(platform, "get_async_ff_sync"):
+            return platform.get_async_ff_sync(self)
+
+        if self._max_input_delay is not None:
+            raise NotImplementedError("Platform '{}' does not support constraining input delay "
+                                      "for AsyncFFSynchronizer"
+                                      .format(type(platform).__name__))
+
         m = Module()
-        m.domains += ClockDomain("reset_sync", async_reset=True, local=True)
+        m.domains += ClockDomain("async_ff", async_reset=True, local=True)
         flops = [Signal(1, name="stage{}".format(index), reset=1)
                  for index in range(self._stages)]
         for i, o in zip((0, *flops), flops):
-            m.d.reset_sync += o.eq(i)
+            m.d.async_ff += o.eq(i)
 
         if self._edge == "pos":
-            m.d.comb += ResetSignal("reset_sync").eq(self.i)
+            m.d.comb += ResetSignal("async_ff").eq(self.i)
         else:
-            m.d.comb += ResetSignal("reset_sync").eq(~self.i)
+            m.d.comb += ResetSignal("async_ff").eq(~self.i)
 
         m.d.comb += [
-            ClockSignal("reset_sync").eq(ClockSignal(self._domain)),
+            ClockSignal("async_ff").eq(ClockSignal(self._domain)),
             self.o.eq(flops[-1])
         ]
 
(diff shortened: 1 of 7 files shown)


nmigen/lib/cdc.py
```diff
@@ -95,6 +95,61 @@ def elaborate(self, platform):
...
+        self._domain = domain
+        self._stages = stages
+
+        if async_edge not in ("pos", "neg"):
+            raise ValueError("AsyncFFSynchronizer async edge must be one of 'pos' or 'neg', not {!r}"
+                             .format(async_edge))
+        self._edge = async_edge
+
+    def elaborate(self, platform):
+        m = Module()
```
awygle: Do we need the ability to swap in a platform-specific module here?

whitequark: Yes, absolutely. In fact we *only* need it here and not in `ResetSynchronizer`; the reason platform overrides exist is to make sure CDC constraints are correct, and `ResetSynchronizer` itself doesn't do any CDC anymore.

nmigen/lib/cdc.py
```diff
@@ -95,6 +95,61 @@ def elaborate(self, platform):
...
+        m.domains += ClockDomain("reset_sync", async_reset=True, local=True)
```
whitequark: This domain should be called something like `async_ff`.

nmigen/lib/cdc.py
```diff
@@ -125,13 +180,14 @@ class ResetSynchronizer(Elaboratable):
     Define the ``get_reset_sync`` platform method to override the implementation of
     :class:`ResetSynchronizer`, e.g. to instantiate library cells directly.
     """
-    def __init__(self, arst, *, domain="sync", stages=2, max_input_delay=None):
+    def __init__(self, arst, *, domain="sync", stages=2, max_input_delay=None, reset_less=False):
```
whitequark: This isn't used anymore, right?

nmigen/lib/cdc.py
awygle: Ah, yes. Hm... It doesn't really want to be called `get_reset_sync` anymore then does it? So much for not touching the platform code.  `get_async_ff_sync`?

nmigen/lib/cdc.py
awygle: No it's not, it snuck in when I was cherry-picking. Oops...

nmigen/lib/cdc.py
whitequark: > It doesn't really want to be called `get_reset_sync` anymore then does it? `get_async_ff_sync`?

Sounds good to me.

> So much for not touching the platform code.

That was inevitable because of the dual edge support I think. (We can postpone that; I do have every supported toolchain installed though.)

awygle: Mm this also means `_max_input_delay` needs to move to `AsyncFFSynchronizer`.

> That was inevitable because of the dual edge support

Should I add an edge parameter to `ResetSynchronizer` while I'm  in here?

whitequark: After reviewing the current code I suggest postponing platform support for configurable edge. This is because only Intel and Xilinx override CDC currently, and the three Xilinx platforms would really benefit from having clock domains with negative polarity resets for implementing this. You can insert an assertion in every platform override that the polarity is what it expects.

whitequark: > Mm this also means `_max_input_delay` needs to move to `AsyncFFSynchronizer`.

Yes.

> Should I add an edge parameter to `ResetSynchronizer` while I'm in here?

No, not yet. I'm actually not sure yet how configurable polarity resets should work, given that `ResetSignal` is late-bound. Not being able to write something like `m.If(ResetSignal()):` to check if you are being reset, and having to branch on polarity, would be really awful. Having `cd.rst` always have the logical polarity is confusing when you're driving it. Inverting the polarity of `ResetSignal()` that is being driven on LHS is technically complex. It's a can of worms... unfortunately one that is required for ASICs.

awygle:
I do not understand what codecov is complaining about - the details point to blank lines and comments?

nmigen/lib/cdc.py
```diff
@@ -95,6 +95,71 @@ def elaborate(self, platform):
...
+    ----------
+    i : Signal(1), in
+        Asynchronous input signal, to be synchronized.
+    o : Signal(1), out
+        Synchronously released output signal.
+    domain : str
+        Name of clock domain to reset.
+    stages : int, >=2
+        Number of synchronization stages between input and output. The lowest safe number is 2,
+        with higher numbers reducing MTBF further, at the cost of increased deassertion latency.
```
whitequark: Please document `async_edge`.

nmigen/vendor/intel.py
```diff
@@ -400,15 +400,22 @@ def get_ff_sync(self, ff_sync):
...
+                i_din=Const(1),
+                o_dout=async_ff_sync.o,
+            )
+        else:
+            m.submodules += Instance("altera_std_synchronizer",
+                p_depth=async_ff_sync._stages,
+                i_clk=ClockSignal(async_ff_sync._domain),
+                i_reset_n=async_ff_sync.i,
+                i_din=Const(1),
+                o_dout=async_ff_sync.o,
```
whitequark: I think you're missing an inverter here, Altera synchronizers are always reset to 0.

whitequark:
> I do not understand what codecov is complaining about - the details point to blank lines and comments?

Ignore it. It's only slightly more useful than annoying. We should probably make it less noisy.

nmigen/vendor/xilinx_spartan_3_6.py
```diff
@@ -437,24 +437,25 @@ def get_ff_sync(self, ff_sync):
...
-        if reset_sync._max_input_delay is not None:
-            raise NotImplementedError("Platform '{}' does not support constraining input delay "
-                                      "for ResetSynchronizer"
-                                      .format(type(self).__name__))
-
+    def get_async_ff_sync(self, async_ff_sync):
         m = Module()
-        m.domains += ClockDomain("reset_sync", async_reset=True, local=True)
+        m.domains += ClockDomain("async_ff", async_reset=True, local=True)
         flops = [Signal(1, name="stage{}".format(index), reset=1,
```
awygle: Looking at this, doesn't this platform need to throw if input delay is set? Since it's not supported?

whitequark: Yep, looks like this fell through the cracks.

whitequark:
Thank you!

PR #333, opened 2020-03-15
back.pysim: Implement modulus operator

sjolsen:
It looks to me like the rtlil backend produces zero when the right-hand side of the modulus operation is zero, so that's what I did here.


whitequark:
Thanks for the PR! Could you add a test?

sjolsen:
I copied the division test, replaced the division results with what the modulus results should be, and added a case for 2 % 0 = 0. The test passes if I run it with `python -m unittest nmigen.test.test_sim`.

whitequark:
Thanks!

PR #338, opened 2020-03-20
Add support for iCE40 internal oscillators.

WRansohoff:
Thanks to #nmigen on Freenode IRC; they basically wrote this for me.

### DIFF
--- a/nmigen/vendor/lattice_ice40.py
+++ b/nmigen/vendor/lattice_ice40.py
@@ -336,10 +336,10 @@ class LatticeICE40Platform(TemplatedPlatform):
 
     @property
     def default_clk_constraint(self):
-        # Internal high-speed oscillator: 48MHz / (2 ^ div)
+        # Internal high-speed oscillator: 48 MHz / (2 ^ div)
         if self.default_clk == "SB_HFOSC":
-            return Clock(48e6 / pow(2, self.hfosc_div))
-        # Internal low-speed oscillator: 10KHz
+            return Clock(48e6 / 2 ** self.hfosc_div)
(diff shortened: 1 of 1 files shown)


nmigen/vendor/lattice_ice40.py
```diff
@@ -350,14 +361,39 @@ def create_missing_domain(self, name):
...
             if self.default_rst is not None:
                 rst_i = self.request(self.default_rst).i
             else:
                 rst_i = Const(0)
 
-            m = Module()
-
             # Power-on-reset domain
             m.domains += ClockDomain("por", reset_less=True, local=True)
             delay = int(15e-6 * self.default_clk_frequency)
```
whitequark: Both SB_HFOSC and SB_LFOSC output is unstable for 100 µs after enabling them. So if one of these oscillators are selected, this power-on reset delay should be increased to 100 µs here.

WRansohoff: Huh, you learn something new every day. Should it have a 5x margin like the BRAM delay and be 500us, or is 100us okay?

whitequark: Nope, 100 us is OK. The reason BRAM delay has a 5x margin is that the 3 us figure is experimental (i.e. typical case) rather than provided by the vendor (i.e. worst case). Since we have no idea where that delay comes from it seems wise to play it safe, but that's not needed here.

WRansohoff:
Okay, I think that should add a longer delay if an internal oscillator is selected.

whitequark:
Thanks for the PR! I've merged it with minor changes.

PR #346, opened 2020-04-02
Add support for using non-compat Elaboratable instances with some compat functions

### LINKED ISSUE #344: run_simulation doesn't call elaborate before trying to set domains
It crashes:

```
  File "src/test_run_simulation_bug.py", line 28, in <module>
    test1()
  File "src/test_run_simulation_bug.py", line 24, in test1
    vcd_name="test_run_simulation_bug.vcd")
  File "/home/jacob/projects/nmigen/nmigen/compat/sim/__init__.py", line 22, in run_simulation
    fragment.domains += ClockDomain("sync")
AttributeError: 'MyModule' object has no attribute 'domains'
```

nMigen version: 2d1e12d00ca13ce6ad92fe145beeaa74925de6ef

Failing code:

```python
from nmigen import Signal, Module, Elaboratable
from nmigen.compat.sim import run_simulation


class MyModule(Elaboratable):
    def __init__(self):
        self.a = Signal()

        m = Module()
        m.d.sync += self.a.eq(~self.a)
        return m


def test1():
    dut = MyModule()

    def generator():
        for _i in range(10):
            print((yield dut.a))
            yield

    run_simulation(dut, generator(),
                   vcd_name="test_run_simulation_bug.vcd")


if __name__ == '__main__':
    test1()
```

programmerjake:
Add support for using non-compat `Elaboratable` instances with `compat.fhdl.verilog.convert` and `compat.run_simulation`

Fixes #344

### DIFF
--- a/nmigen/compat/fhdl/verilog.py
+++ b/nmigen/compat/fhdl/verilog.py
@@ -4,6 +4,7 @@ from ...hdl.ir import Fragment
 from ...hdl.cd import ClockDomain
 from ...back import verilog
 from .conv_output import ConvOutput
+from .module import Module
 
 
 def convert(fi, ios=None, name="top", special_overrides=dict(),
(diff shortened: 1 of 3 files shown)


whitequark:
Thanks for the fix!

programmerjake:
Wow, that's fast!

Issue #228, opened 2019-09-23
Reconsider simulator interface


nmigen-issue-migration:
<a href="https://github.com/whitequark"><img src="https://avatars1.githubusercontent.com/u/54771?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Monday Sep 23, 2019 at 08:52 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/228_

----

The oMigen simulator had a strange interface where user processes would run after a clock edge, but before synchronous signals assume their new value. Not only it is very hard to use and bug-prone (my simulation testing workflow, for example, involves adding `yield` until tests pass...), but it also makes certain patterns impossible! For example, let's consider the `FIFOInterface.write()` simulation function:

```python
def write(self, data):
    """Write method for simulation."""
    assert (yield self.w_rdy)
    yield self.w_data.eq(data)
    yield self.w_en.eq(1)
    yield
    yield self.w_en.eq(0)
    yield
```

It *has* to take two clock cycles. This is because, after the first `yield`, it is not yet possible to read the new value of `self.w_rdy`; running `(yield self.w_rdy)` would return the value from the previous cycle. To make sure it's updated, it is necessary to deassert `self.w_en`, and waste another cycle doing nothing.

Because of this, I've removed these methods from nMigen FIFOInterface in a57b76fb.

I think we should look at prior art for cosimulation (cocotb?) and adopt a usage pattern that is easier to use. This is complicated by the fact that it is desirable to run oMigen testbenches unmodified.

andresdemski:
Hello,
I made a simple cxxrtl wrapper where i implemented an interface similar to COCOTB for nMigen.

https://github.com/andresdemski/nmigen-yosim

I know @whitequark is going to add an official cxxrtl support, take this just as an API proposal and don't look at the implementation please (i'm not C++ developer).

Some interesting features of "yosim":
* rising_edge, falling_edge and timer triggers are implemented
* configurable sim step
* vcd support (you can choose which signals you want to dump)
* fork and join capabilities
* the scheduller and triggers are in C++

whitequark:
> take this just as an API proposal

[...]

PR #348, opened 2020-04-07
back.pysim performance improvements

sjolsen:
I'm not sure how high a priority the Python simulation backend is or whether it's worth making the structural changes to hdl.ast to optimize hashing, but I figured I would share what I've figured out so far in case it's useful. Please let me know what you think.

The changes here fall roughly into the following categories:

- Fixing what I believe to be the intent behind curr/next comparisons, and avoiding doing work where (I think) it's known there is no work to be done
- Optimizing the hashing of Value/Signal/Statement key objects, since this is done in a few very frequently traversed code paths
- Caching compilation results (from hdl.ast to what I believe is Python bytecode)
- Micro-optimizations. I think the only one that made it through was the VCD hash table lookup, and it's pretty marginal at this point

I used the following simulation to make performance measurements: [sim_perf_test.py](https://github.com/sjolsen/nmigen-nexys/blob/master/test/sim_perf_test.py). I have perf data for each commit in the chain, both with and without VCD output enabled. I can see about uploading them somewhere if there's interest, but in the meantime here's a summary:

With no VCD output:

Commit | Description | Time elapsed (s) | Reduction (incr.) | Reduction (cum.)
-- | -- | -- | -- | --
bb1bbcc5 | hdl.mem: fix source location of ReadPort.en. | 116.8 | -- | --
52e30bc8 | back.pysim: Fix curr=next optimizations | 90.77 | 22.29% | 22.29%
6902c99a | hdl.ast: Intern value/statement keys off the hot path | 80.74 | 11.05% | 30.87%
642e4774 | back.pysim: Cache compilation results | 32.17 | 60.16% | 72.46%
d1fe3c63 | back.pysim: Reuse clock simulation commands | 25.29 | 21.39% | 78.35%
65405683 | back.pysim: Eliminate duplicate dict lookup in VCD update | 24.88 | 1.62% | 78.70%

With VCD output enabled:

Commit | Description | Time elapsed (s) | Reduction (incr.) | Reduction (cum.)
-- | -- | -- | -- | --
bb1bbcc5 | hdl.mem: fix source location of ReadPort.en. | 152.6 | -- | --
52e30bc8 | back.pysim: Fix curr=next optimizations | 127.4 | 16.51% | 16.51%
6902c99a | hdl.ast: Intern value/statement keys off the hot path | 118 | 7.38% | 22.67%
642e4774 | back.pysim: Cache compilation results | 66.94 | 43.27% | 56.13%
d1fe3c63 | back.pysim: Reuse clock simulation commands | 57.16 | 14.61% | 62.54%
65405683 | back.pysim: Eliminate duplicate dict lookup in VCD update | 53.25 | 6.84% | 65.10%

I did run the nMigen unit tests, but I haven't figured out yet how to set up sby on Windows so I don't know whether those tests are broken by my changes.

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -108,7 +108,69 @@ def signed(width):
     return Shape(width, signed=True)
 
 
-class Value(metaclass=ABCMeta):
+class _Internable(object):
+
+    class _Intern(object):
+        """Interned object representation suitable for fast hash lookups."""
+        def __init__(self, obj, properties, ordered=False):
+            super().__init__()
+            self.obj = obj
+            self.properties = properties
+            self.ordered = ordered
+            self._hash = hash(self.properties)
+
+        def __hash__(self):
+            return self._hash
+
+        def __eq__(self, other):
+            if type(self.obj) is not type(other.obj):
+                return False
+            return self.properties == other.properties
+
+        def __lt__(self, other):
+            if type(self.obj) is not type(other.obj):
+                raise TypeError(f"No ordering defined for objects of differing "
+                                f"type: {self!r} and {other!r}")
+            if not self.ordered:
+                raise TypeError(f"No ordering defined for object {self!r}")
+            if not other.ordered:
+                raise TypeError(f"No ordering defined for object {other!r}")
+            return self.properties < other.properties
+
+        def __repr__(self):
+            return "<{}._Internable._Intern {!r}>".format(__name__, self.obj)
+
+    class _Unhashable(object):
+        """Non-key for objects discovered at run time to be unhashable."""
+        def __init__(self, obj, reason):
+            super().__init__()
+            self.obj = obj
+            self.reason = reason
+
+        def __hash__(self):
+            raise TypeError(self.reason)
+
+        def __eq__(self, other):
+            raise TypeError(self.reason)
+
+        def __lt__(self, other):
+            raise TypeError(self.reason)
+
+        def __repr__(self):
+            return "<{}._Internable._Unhashable {!r}>".format(__name__, self.obj)
+
+    @property
+    def _key(self):
+        # Should be provided by concrete classes during __init__. This is not
+        # done by Value itself because the subclass instance data needed to
+        # create the key isn't available until after Value.__init__ returns. We
+        # create the key once per Value/Signal/etc. at initialization because
+        # hashing is on a VERY hot code path.
+        return self._Unhashable(
+            self, f"No interning information provided for object {self!r}")
+
+
+class Value(_Internable, metaclass=ABCMeta):
     @staticmethod
     def cast(obj):
         """Converts ``obj`` to an nMigen value.
@@ -123,6 +185,14 @@ class Value(metaclass=ABCMeta):
         if isinstance(obj, Enum):
             return Const(obj.value, Shape.cast(type(obj)))
         raise TypeError("Object {!r} cannot be converted to an nMigen value".format(obj))
+
+    @staticmethod
+    def _castable(obj):
+        try:
+            _ = Value.cast(obj)
+            return True
+        except TypeError:
+            return False
 
     def __init__(self, *, src_loc_at=0):
         super().__init__()
@@ -502,6 +572,11 @@ class Const(Value):
             shape = Shape.cast(shape, src_loc_at=1 + src_loc_at)
         self.width, self.signed = shape
         self.value = self.normalize(self.value, shape)
+        self.__key = self._Intern(self, (self.value,), ordered=True)
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
         return Shape(self.width, self.signed)
@@ -526,6 +601,11 @@ class AnyValue(Value, DUID):
         if not isinstance(self.width, int) or self.width < 0:
             raise TypeError("Width must be a non-negative integer, not {!r}"
                             .format(self.width))
+        self.__key = self._Intern(self, (self.duid,), ordered=True)
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
         return Shape(self.width, self.signed)
@@ -552,8 +632,19 @@ class Operator(Value):
         super().__init__(src_loc_at=1 + src_loc_at)
         self.operator = operator
         self.operands = [Value.cast(op) for op in operands]
+        self._shape = self._compute_shape()
+        self.__key = self._Intern(
+            self,
+            (self.operator, tuple(o._key.properties for o in self.operands)))
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
+        return self._shape
+
+    def _compute_shape(self):
         def _bitwise_binary_shape(a_shape, b_shape):
             a_bits, a_sign = a_shape
             b_bits, b_sign = b_shape
@@ -570,7 +661,7 @@ class Operator(Value):
                 # first signed, second operand unsigned (add sign bit)
                 return Shape(max(a_bits, b_bits + 1), True)
 
-        op_shapes = list(map(lambda x: x.shape(), self.operands))
+        op_shapes = [x.shape() for x in self.operands]
         if len(op_shapes) == 1:
             (a_width, a_signed), = op_shapes
             if self.operator in ("+", "~"):
@@ -669,6 +760,14 @@ class Slice(Value):
         self.value = Value.cast(value)
         self.start = start
         self.stop  = stop
+        self.__key = self._Intern(
+            self,
+            (self.value._key.properties, self.start, self.stop),
+            ordered=self.value._key.ordered)
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
         return Shape(self.stop - self.start)
@@ -696,6 +795,13 @@ class Part(Value):
         self.offset = Value.cast(offset)
         self.width  = width
         self.stride = stride
+        self.__key = self._Intern(
+            self,
+            (self.value._key.properties, self.offset._key.properties, self.width, self.stride))
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
         return Shape(self.width)
@@ -739,6 +845,11 @@ class Cat(Value):
     def __init__(self, *args, src_loc_at=0):
         super().__init__(src_loc_at=src_loc_at)
         self.parts = [Value.cast(v) for v in flatten(args)]
+        self.__key = self._Intern(self, tuple(p._key.properties for p in self.parts))
+
+    @property
+    def _key(self):
+        return self.__key
 
     def shape(self):
         return Shape(sum(len(part) for part in self.parts))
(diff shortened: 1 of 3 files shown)


whitequark:
> * Fixing what I believe to be the intent behind curr/next comparisons

Curr/next comparisons implement the two-phase simulation algorithm that is deterministic in presence of asynchronous code. I looked at your fix and it will lead to incorrect simulation results.

> * Optimizing the hashing of Value/Signal/Statement key objects, since this is done in a few very frequently traversed code paths

I looked at your changes and while the intent seems reasonable, the patch seems fairly invasive at first glance. I'll take another look later.

> * Caching compilation results (from hdl.ast to what I believe is Python bytecode)

Extremely good change, one I was going to make for some time but never got around to.

> * Micro-optimizations

Also looks all reasonable.

<hr>

To summarize, 642e477, d1fe3c6 and 6540568 are mergeable as-is, 6902c99 needs more discussion, and 52e30bc is unsound.

nmigen/hdl/ast.py
```diff
@@ -124,6 +124,14 @@ def cast(obj):
             return Const(obj.value, Shape.cast(type(obj)))
         raise TypeError("Object {!r} cannot be converted to an nMigen value".format(obj))
 
+    @staticmethod
+    def castable(obj):
```
whitequark, line 128: A patch dedicated to optimization should not introduce a new public API.

sjolsen:
Thanks!

>Curr/next comparisons implement the two-phase simulation algorithm that is deterministic in presence of asynchronous code. I looked at your fix and it will lead to incorrect simulation results. I can explain the algorithm if necessary.

I'm not sure I understand, but it's easy enough to revert. Would you prefer I modify the history to keep it linear or layer revert commits etc. on top of what's already here?

>I looked at your changes and while the intent seems reasonable, the patch seems fairly invasive at first glance. I'll take another look later.

Yeah, I wasn't entirely sure how to make this less invasive. A big part of the performance gain was from eliminating glue logic from what was previously ValueKey/SignalKey.

>To summarize, 642e477, d1fe3c6 and 6540568 are mergeable as-is, 6902c99 needs more discussion, and 52e30bc is unsound.

As it is, 642e477 depends on 6902c99 -- that was the motivation for the changes to ValueKey. I actually started with repr(value) as the key, but there was still a fair bit of overhead there.

>A patch dedicated to optimization should not introduce a new public API.

Should that specific method be named _castable? Or should it go somewhere else?

whitequark:
> I'm not sure I understand, but it's easy enough to revert.

The basic idea is that:
  1. `eval` only looks at `curr` and only modifies `next`;
  2. `commit` is the only one who can assign from `next` to `curr`;
  3. the `eval`/`commit` loop continues while there are any changes.

The reason it's arranged like that is because rules (1) and (2) ensure that no matter in which order `eval` and `commit` are run (i.e. no matter how any particular netlist is translated, how it's arranged into modules, etc), each full `eval`/`commit` cycle will have the exact same result. Rule (3), by iterating to fixed point, ensures that all updates in a netlist fully propagate before advancing time--for example, consider a case where you have two domains with asynchronous reset chained, and you reset the first one. This will require two delta cycles (`eval`/`commit` loops).

> Would you prefer I modify the history to keep it linear or layer revert commits etc. on top of what's already here?

I strongly prefer keeping the history linear.

> Yeah, I wasn't entirely sure how to make this less invasive. A big part of the performance gain was from eliminating glue logic from what was previously ValueKey/SignalKey.

> Should that specific method be named _castable? Or should it go somewhere else?

I'm not sure yet. I'll need to understand your proposed changes in detail to see if there is a less invasive way to do the same thing, or if there's really no other choice.

sjolsen:
I reverted the unsound curr/next change and rearranged the interning API to conserve the existing ValueKey and SignalKey classes. I think all the new abstractions should now be hidden.

whitequark:
Thanks. I've cherry-picked the second two commits. I expect it will take a while to convince myself that the proposed interning changes are the way forward. I appreciate that you rearranged them to have no impact to the public API, but the bulk of my concern has to do with the increased internal complexity as well as soundness in face of mutability of AST nodes: I am hesitant to introduce any sort of persistent hashing while properties of most AST nodes remain writable.

Have you considered adding fast paths to the simulator, for the case where bench code operates on simple signals alone? This removes the need for caching bytecode, is an easy to understand change with minimal impact, and should yield most of the performance benefit.

sjolsen:
>I expect it will take a while to convince myself that the proposed interning changes are the way forward

Understood. I'm not necessarily expecting to get this specific design upstreamed, it's just where I ended up after probing for performance improvements.

>the bulk of my concern has to do with the increased internal complexity as well as soundness in face of mutability of AST nodes: I am hesitant to introduce any sort of persistent hashing while properties of most AST nodes remain writable.

That makes sense. Do you anticipate the AST becoming immutable?

>Have you considered adding fast paths to the simulator, for the case where bench code operates on simple signals alone? This removes the need for caching bytecode, is an easy to understand change with minimal impact, and should yield most of the performance benefit.

I'm not sure I follow. Are you thinking along the lines of falling back to a simpler scheduler if only the default sync and comb domains are used? Or did you have something else in mind?

sjolsen:
>The basic idea is that [...]

Thank you for explaining this, BTW. The eval/commit/loop approach makes perfect sense to me. I think I see at least one way the unsoundness of 52e30bc could manifest: if one invocation sets next != curr, and a second invocation sets it back to curr, it will be left with the wrong value for next. That's important because a single fragment could set next repeatedly.

I retraced my steps and realized that the call to pending.clear was the only part of that commit that actually mattered for performance. The rest of the changes were a result of trying to micro-optimize _SignalState.commit. If I'm not (still) very much mistaken, calling pending.clear should be correct since at that point in _SimulatorState.commit, every pending update from the current eval/commit loop has been effected, and the next eval step will set pending if necessary. What do you think?

whitequark:
> Do you anticipate the AST becoming immutable?

This is a bit of a complex issue. At the very beginning, I designed nMigen so that it never mutates ASTs (and it still doesn't). When it needs to change an AST, it emits a new one with the necessary elements changed (the FragmentTransformer stuff). I knew this could be a performance issue when writing it, but I didn't realize how just *how* slow it would be.

This slowness could be fixed in two ways: a simpler but more flawed one--which is to copy the entire AST once, eliminating any sharing, and then mutating it; and a more complex but principled one--which is to not change ASTs in first place (with or without mutation), and instead treat `ClockSignal`, `ResetSignal`, domain names, etc similar to variable bindings. That is, instead of having `DomainRenamer` alter the actual AST, it would alter a binding table carried along with it, with this table specifying which domain becomes which after every transformation is applied.

The principled approach potentially has very significant benefits. It would dramatically improve performance on all designs but especially large generic ones. It would eliminate multiple complicated corner cases from the internal machinery. It would bring the internals of nMigen closer to well understood and well behaved compiler development principles, namely using variable bindings for late bound names (there's nothing wrong in principle with term rewriting it's currently doing, but this term rewriting is very ad-hoc and hard to analyze). It's also a lot of work.

There is a separate issue, which is the *user* mutability of ASTs, whether interior or exterior. For most expression nodes (everything except Signal, ArrayProxy, and UserValue), this is clearly undesirable. For ArrayProxy and UserValue, it's just interior mutability, and there's already code to cope with it, which should be largely sufficient. For Signal, the story is a bit complex--I believe it's never been "officially" supported, but you *can* mutate Signals and the results are currently correct, provided you never use them in expressions before you stop mutating them. This is used internally in nMigen in the FSM generator code, so it seems like a valuable capability and we might want to enforce that invariant in a future immutable AST.

To summarize, I think the AST *should* be immutable, but we're not quite there yet, and there is at least one corner case that currently presents an API with a soundness issue that would likely need to be handled explicitly.

whitequark:
> That's important because a single fragment could set next repeatedly.

Yup, and there are many cases where this happens--for example each time you assert reset.

> If I'm not (still) very much mistaken, calling pending.clear should be correct since at that point in _SimulatorState.commit, every pending update from the current eval/commit loop has been effected, and the next eval step will set pending if necessary. What do you think?

Could you show a patch that does what you want? I'm not quite sure exactly what change you're referring to, and it seems like preparing a patch isn't a lot of effort here.

whitequark:
> back.pysim: Clear pending updates after they are effected

Oh... did I just completely forget to remove elements from `self.pending` anywhere at all? Well that *would* make things a lot slower than they should be!!

whitequark:
I'm curious, could you measure the perf impact? I'm really glad you noticed this, it's a really silly mistake with severe consequences.

sjolsen:
I also cherry-picked it onto master:

```
* 555c23c8 back.pysim: Clear pending updates after they are effected
| * d891a794 back.pysim: Clear pending updates after they are effected
| * daa616f0 back.pysim: Eliminate duplicate dict lookup in VCD update
| * 1b60dbda ...
| * d698e6f4 ...
| * 3f466e89 ...
|/
* bb1bbcc5 hdl.mem: fix source location of ReadPort.en.
```

Here's what I got with VCD output disabled:

Commit | Description | Time elapsed (s) | Reduction (incr.) | Reduction (cum.)
-- | -- | -- | -- | --
bb1bbcc5 | hdl.mem: fix source location of ReadPort.en. | 152.6 | -- | --
... | ... | ... | ... | ...
daa616f0 | back.pysim: Eliminate duplicate dict lookup in VCD update | 47.24 | ... | 69.04%
d891a794 | back.pysim: Clear pending updates after they are effected | 25.05 | 46.97% | 83.58%

Commit | Description | Time elapsed (s) | Reduction (incr.) | Reduction (cum.)
-- | -- | -- | -- | --
bb1bbcc5 | hdl.mem: fix source location of ReadPort.en. | 152.6 | -- | --
555c23c8 | back.pysim: Clear pending updates after they are effected | 87.41 | 42.72% | 42.72%

whitequark:
I've cherry-picked commit d891a79, thank you!

> I'm not sure I follow. Are you thinking along the lines of falling back to a simpler scheduler if only the default sync and comb domains are used? Or did you have something else in mind?

Something else. You can notice that most simulator commands are along the lines of `yield sig` or `yield sig.eq(123)`. Which means that by special-casing just these two patterns we can dramatically speed up simulation without any of the interning complexity.

sjolsen:
I integrated the cherry-picked changes to clean up the PR in case you decide to do something with it. One thing that occurred to me that I haven't bothered to change is that the compilation cache should probably be scoped to the simulator rather than the coroutine process.

>Something else. You can notice that most simulator commands are along the lines of yield sig or yield sig.eq(123). Which means that by special-casing just these two patterns we can dramatically speed up simulation without any of the interning complexity.

That makes a lot of sense. Speaking of the simulator interface, something I've thought of requesting is a way to wait asynchronously for a signal change from a coroutine process. It sounds from #228 like you already have some idea of how you'd like to do that? FWIW I found Value hashing useful for hacking together (emphasis on "hack") an implementation using _FragmentCompiler.

whitequark:
> It sounds from #228 like you already have some idea of how you'd like to do that?

Internally, this is easy to do. The external API needs to be designed.

whitequark:
As per discussion above, I am closing this PR; I would encourage you (or anyone else) to submit another PR with a less invasive approach I suggested above. If it turns out that (contrary to my expectations) the simpler approach does not bring similar performance benefits, then I'm going to reevaluate the approach in this PR. Thanks for doing this work!

PR #352, opened 2020-04-11
Add rotate by constant

### LINKED ISSUE #339: Add rotate left/right by constant amount
While it's not hard to roll your own in a function, it seems a bit more ergonomic to have `Value.rotate_{left,right}()` instead.

ZirconiumX:
I think I got the signs right, anyway.

Closes #339.


nmigen/hdl/ast.py
```diff
@@ -417,6 +417,44 @@ def matches(self, *patterns):
...
+        offset : int, in
+            Amount to rotate by.
+
+        Returns
+        -------
+        Value, out
+            The input rotated left by offset if offset is positive, else the input rotated right by -offset.
+        """
+        if isinstance(offset, int):
+            if offset > len(self) or -offset > len(self):
```
whitequark, line 434: Is the condition necessary? It's just a nop if `offset < len(self)`.

nmigen/test/test_hdl_ast.py
```diff
@@ -201,6 +201,24 @@ def test_getitem_wrong(self):
                 msg="Cannot index value with 'str'"):
             Const(31)["str"]
 
+    def test_rotate_left(self):
+        self.assertRepr(Value.cast(256).rotate_left(1), "(cat (slice (const 9'd256) 8:9) (slice (const 9'd256) 0:8))")
+        self.assertRepr(Value.cast(256).rotate_left(7), "(cat (slice (const 9'd256) 2:9) (slice (const 9'd256) 0:2))")
+        self.assertRepr(Value.cast(256).rotate_left(-1), "(cat (slice (const 9'd256) 1:9) (slice (const 9'd256) 0:1))")
+        self.assertRepr(Value.cast(256).rotate_left(-7), "(cat (slice (const 9'd256) 7:9) (slice (const 9'd256) 0:7))")
+        with self.assertRaises(TypeError,
```
whitequark, line 209: Tests for API misuse should be separate from tests for correct API use (extract these into something like `test_rotate_left_wrong`).

nmigen/test/test_hdl_ast.py
```diff
@@ -201,6 +201,24 @@ def test_getitem_wrong(self):
...
+                msg="Argument to rotate_left must be an integer, not 'str'"):
```
whitequark, line 210: I would say "rotate amount" rather than "argument to rotate_left" since it gets the point across faster.

nmigen/hdl/ast.py
```diff
@@ -417,6 +417,42 @@ def matches(self, *patterns):
         else:
             return Cat(*matches).any()
 
+    def rotate_left(self, offset):
+        """Rotate left by constant modulo 2**len(self).
+        
+        Parameters
+        ----------
+        offset : int, in
```
whitequark, line 425: Non-`Value` parameters are always inputs, since Python does not have out parameters. Only `Value` parameters can, in principle, be outputs.

nmigen/hdl/ast.py
```diff
@@ -417,6 +417,42 @@ def matches(self, *patterns):
...
+            Amount to rotate by.
+
+        Returns
+        -------
+        Value, out
+            The input rotated left by offset if offset is positive, else the input rotated right by -offset.
+        """
+        if isinstance(offset, int):
```
whitequark, line 433: Could you invert this check for consistency with the rest of the code? If there are multiple precondition checks, the code is easier to understand that way; see `matches` for a good example of why.

Issue #355, opened 2020-04-14
[RFC] Redesign UserValue to avoid breaking code that inherits from it


whitequark:
`UserValue` fulfills an important role: it allows code downstream from nMigen to define objects that can be fluently used in nMigen code, i.e. on RHS and, possibly, LHS of nMigen assignments. It was originally created to reflect the non-primitive nature of Records, and also because there was some disagreement on the best way to handle Records. These disagreements have persisted, and I both acknowledge them, and see `UserValue` as an important tool to explore any solutions.

Unfortunately, as currently implemented, it has a fatal flaw. It pollutes the namespace of the classes that inherit from it. We add new methods to `Value` liberally (see #352 for the latest addition), and each of these methods has the potential of clashing with downstream code. As it is, any new method on `Value` is a breaking change, which is not acceptable.

Fortunately, there is a good solution. One can observe that `UserValue`s (for example, any `Record`s--at least post-#354) do not really need to implement any numeric operations because they are not numbers--indeed, if they were simple numbers, they would be normal `Value`s. They don't necessarily implement bit vector operations either--for example, `Record` overrides `__getitem__`. They only need to be *implicitly castable* to `Value` as opposed to be a *kind of* `Value`.

I propose that we change `UserValue` to not inherit from `Value` at all, and make sure it injects a small and well-defined set of attributes into derived classes. Its primary purpose would then to be a marker class that `Value.cast` uses to recognize that it should first lower its argument.

Unresolved questions:
  * Should we rename `lower` to `as_value`? The former is a fairly generic term, and therefore potentially useful to derived classes. The latter, in context of nMigen, is quite unambiguous, and very unlikely to clash with user-defined record fields.
  * Should we provide `eq`? Some `UserValue`s, such as `Record`s, will be commonly used on LHS, and writing `rec1.as_value().eq(rec2)` is a lot less convenient than `rec1.eq(rec2)`. However, one can implement a `UserValue` that lowers to a constant, and that would be never usable on LHS. 
     One option is to leave it to the specific derived class to implement `eq`, since the implementation is as simple as `return Assign(self, value, src_loc_at=1)`. Such an approach could also be useful because the derived class might be able to implement useful diagnostics that are not applicable to `Value`s in general.

anuejn:
[...]

PR #357, opened 2020-04-15
make Record inherit from UserValue

### LINKED ISSUE #354: Inherit `Record` from `UserValue`
For all intents and purposes, a `Record` is a kind of `UserValue`. However, it predates `UserValue` and for historical reasons is special-cased in a lot of places. It should just inherit from `UserValue`.

anuejn:
closes #354

### DIFF
--- a/nmigen/back/rtlil.py
+++ b/nmigen/back/rtlil.py
@@ -360,9 +360,6 @@ class _ValueCompiler(xfrm.ValueVisitor):
     def on_Initial(self, value):
         raise NotImplementedError # :nocov:
 
-    def on_Record(self, value):
-        return self(ast.Cat(value.fields.values()))
-
     def on_Cat(self, value):
         return "{{ {} }}".format(" ".join(reversed([self(o) for o in value.parts])))
 
@@ -373,7 +370,11 @@ class _ValueCompiler(xfrm.ValueVisitor):
         if value.start == 0 and value.stop == len(value.value):
             return self(value.value)
 
-        sigspec = self._prepare_value_for_Slice(value.value)
+        if isinstance(value.value, ast.UserValue):
+            sigspec = self._prepare_value_for_Slice(value.value._lazy_lower())
+        else:
+            sigspec = self._prepare_value_for_Slice(value.value)
+
         if value.start == value.stop:
             return "{}"
         elif value.start + 1 == value.stop:
@@ -639,7 +640,7 @@ class _LHSValueCompiler(_ValueCompiler):
         return wire_next or wire_curr
 
     def _prepare_value_for_Slice(self, value):
-        assert isinstance(value, (ast.Signal, ast.Slice, ast.Cat, rec.Record))
+        assert isinstance(value, (ast.Signal, ast.Slice, ast.Cat))
         return self(value)
 
     def on_Part(self, value):
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -1186,7 +1186,10 @@ class UserValue(Value):
 
     def _lazy_lower(self):
         if self.__lowered is None:
-            self.__lowered = Value.cast(self.lower())
+            lowered = self.lower()
+            if isinstance(lowered, UserValue):
+                lowered = lowered._lazy_lower()
+            self.__lowered = Value.cast(lowered)
         return self.__lowered
 
     def shape(self):
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -85,7 +85,7 @@ class Layout:
 
 
 # Unlike most Values, Record *can* be subclassed.
-class Record(Value):
+class Record(UserValue):
     @staticmethod
     def like(other, *, name=None, name_suffix=None, src_loc_at=0):
         if name is not None:
@@ -113,6 +113,8 @@ class Record(Value):
         return Record(other.layout, name=new_name, fields=fields, src_loc_at=1)
 
     def __init__(self, layout, *, name=None, fields=None, src_loc_at=0):
+        super().__init__(src_loc_at=src_loc_at)
+
         if name is None:
             name = tracer.get_var_name(depth=2 + src_loc_at, default=None)
 
@@ -165,8 +167,8 @@ class Record(Value):
         else:
             return super().__getitem__(item)
 
-    def shape(self):
-        return Shape(sum(len(f) for f in self.fields.values()))
+    def lower(self):
+        return Cat(self.fields.values())
 
     def _lhs_signals(self):
         return union((f._lhs_signals() for f in self.fields.values()), start=SignalSet())
--- a/nmigen/test/test_hdl_ast.py
+++ b/nmigen/test/test_hdl_ast.py
@@ -916,6 +916,14 @@ class UserValueTestCase(FHDLTestCase):
         self.assertEqual(uv.shape(), unsigned(1))
         self.assertEqual(uv.lower_count, 1)
 
+    def test_lower_to_user_value(self):
+        uv = MockUserValue(MockUserValue(1))
+        self.assertEqual(uv.shape(), unsigned(1))
+        self.assertIsInstance(uv.shape(), Shape)
+        uv.lowered = MockUserValue(2)
+        self.assertEqual(uv.shape(), unsigned(1))
+        self.assertEqual(uv.lower_count, 1)
+
 
 class SampleTestCase(FHDLTestCase):
     def test_const(self):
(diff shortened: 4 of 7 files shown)


nmigen/hdl/rec.py
```diff
@@ -113,6 +113,8 @@ def concat(a, b):
         return Record(other.layout, name=new_name, fields=fields, src_loc_at=1)
 
     def __init__(self, layout, *, name=None, fields=None, src_loc_at=0):
+        super().__init__(src_loc_at=src_loc_at)
```
whitequark, line 116: I think this needs a `+ 1`, but please check if the source locations are in fact correct.
```suggestion
        super().__init__(src_loc_at=src_loc_at + 1)
```

nmigen/back/rtlil.py
```diff
@@ -639,7 +636,7 @@ def on_Signal(self, value):
         return wire_next or wire_curr
 
     def _prepare_value_for_Slice(self, value):
-        assert isinstance(value, (ast.Signal, ast.Slice, ast.Cat, rec.Record))
+        assert isinstance(value, (ast.Signal, ast.Slice, ast.Cat, ast.UserValue))
```
whitequark, line 639: This change is actually quite subtle. Since `UserValue`s cannot by themselves be translated, and should be lowered first, this assertion would not work correctly (and neither would the other implementation of `_prepare_value_for_Slice`). I think what you need to do here is to lower the `UserValue` in `on_Slice` *before* passing it to `_prepare_value_for_Slice`.

anuejn: okay

nmigen/hdl/rec.py
anuejn: will check

anuejn:
Another thing that came to my mind just now is, that we maybe want to somehow mark "things that are somehow a bundle". This would help to implement genlib.record.iter_flat() without the assumption that Record is the only bundle. Do you have any ideas of how to achive that?

whitequark:
> This would help to implement genlib.record.iter_flat() without the assumption that Record is the only bundle. Do you have any ideas of how to achive that?

As described elsewhere, I don't think this is a good idea, and nMigen's built-in records will not support `iter_flat`. (They are still there in `nmigen.compat`, but only for compatibility.)

anuejn:
okay

whitequark:
To add to this, I think it is important to examine the use cases of `iter_flat` and see which other mechanism would suit them well. Of course, it is also important to have an escape hatch for those who *really* want to use their own kind of records in their own codebase, which is why `UserValue` is there.

But I don't currently see a convincing case for having a core interoperability mechanism for flat iteration.

nmigen/hdl/rec.py
anuejn: it works without +1. with +1 i get ValueError: call stack is not deep enough. without the source location is correct

whitequark: Ah, it looks like I already added this in the `UserValue` constructor.

nmigen/back/rtlil.py
anuejn: do we want to allow returning other UserValues from the lower method?
If i get it right this is currently not supported but if we do not allow it now we would break all code that lowers UserValues into Records

whitequark: Good question! That's something I did not consider. I think that for now, we can change `UserValue._lazy_lower` to recursively call `_lazy_lower` again if it gets another `UserValue`. Once #355 is solved, this logic should be a part of `Value.cast`.

anuejn:
I have some utility function to measure the size of a Module (/Elaboratable) by synthesizing it with Yosys and pulling some output from the stat command. To generate the rtlil for this, I use a similar function to extract all nMigen Value like things and add them to the io ports.

But maybe this usecase is a bit too unimportant / hacky to be supported ;).

nmigen/back/rtlil.py
anuejn: okay i will do that too in the rtlil backend

whitequark:
> To generate the rtlil for this, I use a similar function to extract all nMigen Value like things and add them to the io ports.

Take a look at the Yosys `expose` command. It will let you do this without any cooperation from nMigen at all.

anuejn:
ah nice thanks :)

whitequark:
Oh, and although the Yosys `expose` command is probably the most elegant way to do this, you're not limited to using it; you can convert the `Module` into a `Fragment`, and then iterate through `stmt._rhs_signals()` for every statement. Really, your use case has quite a few ways to solve it, none of which involve `iter_flat` as a core interoperability mechanism.

nmigen/back/rtlil.py
whitequark: You should call `_lazy_lower` in the RTLIL backend, since that is the nMigen-side API of `UserValue`s.

anuejn: bikeshed: isn't _lazy_lower() is not really lazy, its more caching?

whitequark: The "laziness" in case of `_lazy_lower` is because you can use a `UserValue` without committing to a specific shape upfront, only doing that on first use. The caching aspect is incidental, it is there to guarantee correctness (internal consistency of nMigen data structures).

nmigen/back/rtlil.py
```diff
@@ -373,6 +370,9 @@ def on_Slice(self, value):
         if value.start == 0 and value.stop == len(value.value):
             return self(value.value)
 
+        if isinstance(value.value, ast.UserValue):
+            value.value = value.value._lazy_lower()
```
whitequark, line 374: You should never mutate any AST node anywhere in nMigen internals. That's one of the core concepts; you can elaborate/synthesize a design as many times as you want and it'll always work the same.

nmigen/hdl/ast.py
```diff
@@ -1186,7 +1186,13 @@ def lower(self):
 
     def _lazy_lower(self):
         if self.__lowered is None:
-            self.__lowered = Value.cast(self.lower())
+            lowered = self.lower()
+            if isinstance(lowered, UserValue):
+                lowered = lowered._lazy_lower()
+            elif not isinstance(lowered, Value):
+                raise ValueError("UserValue.lower() needs to return a nMigen Value or UserValue, found {}"
```
whitequark, line 1193: Nit: our convention is to show a repr in error messages such as this, which should provide more information.
```suggestion
                raise ValueError("UserValue.lower() needs to return a nMigen Value or UserValue, not {!r}"
                                                 .format(lowered))
```

anuejn: ok this check was stupid anyway since it doesnt allow for returning e.g. integers and all other cases are cought by Value.cast()

nmigen/hdl/ast.py
```diff
@@ -1186,7 +1186,10 @@ def lower(self):
 
     def _lazy_lower(self):
         if self.__lowered is None:
-            self.__lowered = Value.cast(self.lower())
+            lowered = self.lower()
+            if isinstance(lowered, UserValue):
+                lowered = lowered._lazy_lower()
```
whitequark, line 1191: Could you please write a test for lowering of a `UserValue` that returns another `UserValue`?

nmigen/hdl/ast.py
whitequark: Good catch.

nmigen/test/test_hdl_ast.py
```diff
@@ -908,14 +908,17 @@ def lower(self):
...
-    def test_shape(self):
-        uv = MockUserValue(1)
+    def test_shape(self, lower_to=int):
+        uv = MockUserValue(lower_to(1))
         self.assertEqual(uv.shape(), unsigned(1))
         self.assertIsInstance(uv.shape(), Shape)
-        uv.lowered = 2
+        uv.lowered = lower_to(2)
         self.assertEqual(uv.shape(), unsigned(1))
         self.assertEqual(uv.lower_count, 1)
```
whitequark, line 917: This test is a bit hard to understand--could you duplicate the code? It's not much and it makes it much more clear what's going on.

nmigen/test/test_hdl_xfrm.py
```diff
@@ -620,3 +620,12 @@ def test_lower(self):
...
         """)
+
+
+class UserValueRecursiveTestCase(UserValueTestCase):
+    def setUp(self):
+        self.s = Signal()
+        self.c = Signal()
+        self.uv = MockUserValue(MockUserValue(self.s))
+
+    # inherit the test_lower method from UserValueTestCase because the checks are the same
```
whitequark, line 631: (This one seems reasonably straightforward though.)

nmigen/test/test_hdl_ast.py
anuejn: k

whitequark:
Thanks for the PR!

PR #358, opened 2020-04-15
use declarative setuptools config

graingert:


setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+    Jinja2~=2.11
+packages = find:
+
+[options.entry_points]
+console_scripts =
+    nmigen-rpc = nmigen.rpc:main
+
+[tools.setuptools_scm]
+version_scheme = guess-next-dev
+local_scheme = local_scheme.local_scheme
```
graingert, line 27: I don't think this is correct, might need to be `attr: local_scheme.local_scheme`

whitequark: Can we inline the Python code and get rid of yet another auxiliary file? It's not very important, but going from one file to four with the exact same info is annoying.

graingert: No - that's the point of setuptools declarative though. To avoid python code as much as possible

whitequark: OK, makes sense.

local_scheme.py
```diff
@@ -0,0 +1,6 @@
+def local_scheme(version):
+    return (
```
whitequark, line 2: Could you use the old code for this function please? The changed version is harder to understand IMO.

setup.cfg
```diff
@@ -0,0 +1,27 @@
...
+description = Python toolbox for building complex digital hardware
+license = BSD
+project_urls =
+    Source Code=https://github.com/nmigen/nmigen
+    Bug Tracker=https://github.com/nmigen/nmigen/issues
+
+[options]
+python_requires = ~=3.6
+install_requires =
+    pyvcd~=0.2.0
```
whitequark, line 17: Please keep the comments, they're useful to explain the scope where the dependency is required.

whitequark:
Is `setup.py` required after these changes, or is it only left for compatibility?

graingert:
it is still required

Thomas Grainger

local_scheme.py
graingert: This code has been moved into setuptools-scm-node-and-dirty-tag, it could be moved back inline once https://github.com/pypa/setuptools_scm/issues/418 or https://github.com/pypa/setuptools_scm/issues/419 land

whitequark:
Looks like the setuptools_scm maintainer isn't planning to implement the necessary functionality himself (which is perfectly fine!); I don't have time to do that either. Unless someone wants to implement https://github.com/pypa/setuptools_scm/issues/419 I propose we postpone migration to declarative setuptools configuration indefinitely, since there is very little downside to that.

whitequark:
Turns out it's actually enough to have a minimal `pyproject.toml` alone to fix the original issue with the `venv` command. I've done that in 7fca037f9c14974f499195171bb84c38ebddc1e1.

@graingert Thank you for your effort, even if in the end it was not used. If setuptools_scm ever gains configurable version formats I'll resurrect this PR.

PR #361, opened 2020-04-20
vendor: use nextpnr -12k for -12F devices; remove theoretical devices

ktemkin:
nextpnr now supports -12k; which replaces the use of -25k and --idcode
together to build bitstreams compatible with -12F devices. Use this.

This also removes the LFE5UM-12K and its 5G counterpart; as per Dave Shah
they're currently only theoretical FPGAs.


whitequark:
Thanks!

PR #364, opened 2020-04-22
Fix `_yosys_version()`

hofstee:
`re.match` only matches at the beginning of the string, not each line. This change changes it to do a `re.search` with the multiline flag. I think this is primarily of use when the user has compiled yosys with verific extensions.

Example of `yosys -V` output that failed with the prior code:

```
~/s/nmigen · yosys -V
-- (c) Copyright 1999 - 2020 Verific Design Automation Inc. All rights reserved
-- Compilation time was Fri Feb 28 11:19:43 2020

-- This program will expire on Sun Jun  7 12:19:43 2020

Yosys 0.9+2406 (git sha1 cf716e1f, gcc 8.3.0-6 -Os)
```

Verilog generation is still borked. `yosys -q -` doesn't seem to stop the Verific license from being printed, so it gets captured in the subprocess stdout when nmigen grabs the output from yosys. It might be a bit more foolproof to modify that to write the data to a tempfile instead and pass the file between nmigen/yosys.

```
-- (c) Copyright 1999 - 2020 Verific Design Automation Inc. All rights reserved
-- Compilation time was Fri Feb 28 11:19:43 2020

-- This program will expire on Sun Jun  7 12:19:43 2020

/* Generated by Yosys 0.9+2406 (git sha1 cf716e1f, gcc 8.3.0-6 -Os) */

(* generator = "nMigen" *)
(* top =  1  *)
(* \nmigen.hierarchy  = "top" *)
module top(clk, rst, o);
```

### DIFF
--- a/nmigen/back/verilog.py
+++ b/nmigen/back/verilog.py
@@ -16,6 +16,7 @@ class YosysError(Exception):
 def _yosys_version():
     yosys_path = require_tool("yosys")
     version = subprocess.check_output([yosys_path, "-V"], encoding="utf-8")
+    # If Yosys is built with Verific, then Verific license information is printed first.
     m = re.search(r"^Yosys ([\d.]+)(?:\+(\d+))?", version, flags=re.M)
     tag, offset = m[1], m[2] or 0
     return tuple(map(int, tag.split("."))), offset
--- a/nmigen/back/verilog.py
+++ b/nmigen/back/verilog.py
@@ -16,7 +16,7 @@ class YosysError(Exception):
 def _yosys_version():
     yosys_path = require_tool("yosys")
     version = subprocess.check_output([yosys_path, "-V"], encoding="utf-8")
-    m = re.match(r"^Yosys ([\d.]+)(?:\+(\d+))?", version)
+    m = re.search(r"^Yosys ([\d.]+)(?:\+(\d+))?", version, flags=re.M)
     tag, offset = m[1], m[2] or 0
     return tuple(map(int, tag.split("."))), offset
 


whitequark:
> `yosys -q -` doesn't seem to stop the Verific license from being printed, so it gets captured in the subprocess stdout when nmigen grabs the output from yosys.

I wonder if we can fix that upstream, or if that's something Verific forces on us.

> It might be a bit more foolproof to modify that to write the data to a tempfile instead and pass the file between nmigen/yosys.

The current behavior is very much intentional; eventually I would like to ship Yosys (in the default configuration at least) as a library (likely compiled as WASM, but other options would work too), and I find it rather distasteful when libraries communicate with their host application through files. You are right that it would be a bit more foolproof though.

nmigen/back/verilog.py
```diff
@@ -16,7 +16,7 @@ class YosysError(Exception):
 def _yosys_version():
     yosys_path = require_tool("yosys")
     version = subprocess.check_output([yosys_path, "-V"], encoding="utf-8")
-    m = re.match(r"^Yosys ([\d.]+)(?:\+(\d+))?", version)
+    m = re.search(r"^Yosys ([\d.]+)(?:\+(\d+))?", version, flags=re.M)
```
whitequark, line 19: A comment here would be handy:
```suggestion
    # If Yosys is built with Verific, then Verific license information is printed first.
    m = re.search(r"^Yosys ([\d.]+)(?:\+(\d+))?", version, flags=re.M)
```

whitequark:
Thanks!

hofstee:
> I wonder if we can fix that upstream, or if that's something Verific forces on us.

I think it's Verific. Here's a relevant gdb backtrace:

```
(gdb) bt
#0  __GI___libc_write (fd=1, buf=0x17ea360, nbytes=80) at ../sysdeps/unix/sysv/linux/write.c:26
#1  0x00007ffff750a3bd in _IO_new_file_write (f=0x7ffff764b760 <_IO_2_1_stdout_>, data=0x17ea360, n=80) at fileops.c:1183
#2  0x00007ffff750975f in new_do_write (fp=0x7ffff764b760 <_IO_2_1_stdout_>, fp@entry=0x50, 
    data=0x17ea360 "-- (c) Copyright 1999 - 2020 Verific Design Automation Inc. All rights reserved\n", 
    to_do=to_do@entry=80) at libioP.h:839
#3  0x00007ffff750b509 in _IO_new_do_write (to_do=80, data=<optimized out>, fp=0x50) at fileops.c:430
#4  _IO_new_do_write (fp=fp@entry=0x7ffff764b760 <_IO_2_1_stdout_>, data=<optimized out>, to_do=80) at fileops.c:430
#5  0x00007ffff750b8f3 in _IO_new_file_overflow (f=0x7ffff764b760 <_IO_2_1_stdout_>, ch=10) at fileops.c:791
#6  0x00000000011c4ce5 in Verific::Message::MsgVaList(Verific::msg_type_t, char const*, unsigned long, char const*, __va_list_tag*) ()
#7  0x00000000011c51d8 in Verific::Message::Msg(Verific::msg_type_t, char const*, unsigned long, char const*, ...) ()
#8  0x00000000011c524c in Verific::Message::PrintLine(char const*, char const*, char const*, char const*, char const*) ()
#9  0x00000000011d4aba in Verific::TimeBomb::TimeBomb() ()
#10 0x0000000000cbebd9 in _GLOBAL__sub_I__ZN7Verific9veri_file14_all_librariesE ()
#11 0x00000000012604c5 in __libc_csu_init ()
#12 0x00007ffff74b302a in __libc_start_main (main=0x78f736 <main>, argc=3, argv=0x7fffffffe518, 
    init=0x1260480 <__libc_csu_init>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe508)
    at ../csu/libc-start.c:264
#13 0x00000000007a1e7a in _start ()
```

I tried looking through the yosys code earlier and couldn't see anything that seemed like it would be the cause of it. It seems like yosys doesn't print anything before the header, which gets disabled if you pass `-q`.

whitequark:
> `Verific::TimeBomb::TimeBomb()`

That's, uh, one way to name your license check function.

whitequark:
See https://github.com/YosysHQ/yosys/issues/1980.

whitequark:
As a temporary workaround we could strip the initial lines that are empty or start with `--`, since such lines are not valid in RTLIL anyway.

whitequark:
@hofstee Looks like it's going to be a permanent workaround. Feel free to submit a PR that adds such a hack, it's pretty bad but using temporary files in a situation like this is IMO also quite bad so I prefer the bad option that won't have any impact on people not using Verific.

PR #365, opened 2020-04-23
remove non-verilog code at start of output

hofstee:
See comments at the end of #364.

### DIFF
--- a/nmigen/back/verilog.py
+++ b/nmigen/back/verilog.py
@@ -1,6 +1,7 @@
 import os
 import re
 import subprocess
+import itertools
 
 from .._toolchain import *
 from . import rtlil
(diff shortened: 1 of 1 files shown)


nmigen/back/verilog.py
```diff
@@ -65,6 +66,12 @@ def _convert_rtlil_text(rtlil_text, *, strip_internal_attrs=False, write_verilog
     if popen.returncode:
         raise YosysError(error.strip())
     else:
+        # Remove empty lines and things that are not legal Verilog
+        # syntax at start of file
```
whitequark, line 70: Instead of explaining what the code below does (which is fairly clear from the code itself), this comment should be explaining why this code is necessary, similar to what I asked for in the other PR.

nmigen/back/verilog.py
```diff
@@ -65,6 +66,12 @@ def _convert_rtlil_text(rtlil_text, *, strip_internal_attrs=False, write_verilog
...
+        verilog_text = "\n".join(itertools.dropwhile(
+            lambda x: not x or x.strip().startswith("--"),
```
whitequark, line 72: In general, avoid `not x` (and other implicit boolean conversions) in Python, since it obscures meaning by being less precise than necessary. A condition like `line == "" or line.startswith("--")` would be more clear here. (Is `strip` necessary? I thought Verific only uses `--` at the start of the line.)

whitequark:
Thanks, merged with minor changes.

PR #366, opened 2020-04-24
Check type of `ports` argument to `convert` functions.

### LINKED ISSUE #362: `convert(ports=)` is a hotspot of beginner issues
We have a disproportionally large amount of questions from beginners that relate to the `ports` argument of `convert()` functions. We should add stricter typechecking to it (accepting either `list` or `tuple`, but nothing else, seems like a good idea), plus a "did you mean?" type diagnostic for passing a single signal in place of an iterable of ports.

awygle:
The `ports` argument to the `convert` functions is a frequent hotspot of
beginner issues. Check to make sure it is either a list or a tuple, and
give an appropriately helpful error message if not.

Resolves #362 .

### DIFF
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -532,7 +532,7 @@ class Fragment:
         if ports is None:
             fragment._propagate_ports(ports=(), all_undef_as_ports=True)
         else:
-            # typecheck ports is either list or tuple (#362)
+            # typecheck ports is either list or tuple
             if not isinstance(ports, tuple) and not isinstance(ports, list):
                 msg = "`ports` must be either a list or a tuple, not {!r}"\
                         .format(ports)
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -532,7 +532,6 @@ class Fragment:
         if ports is None:
             fragment._propagate_ports(ports=(), all_undef_as_ports=True)
         else:
-            # typecheck ports is either list or tuple
             if not isinstance(ports, tuple) and not isinstance(ports, list):
                 msg = "`ports` must be either a list or a tuple, not {!r}"\
                         .format(ports)
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -535,9 +535,9 @@ class Fragment:
             # typecheck ports is either list or tuple (#362)
             if not isinstance(ports, tuple) and not isinstance(ports, list):
                 msg = "`ports` must be either a list or a tuple, not {!r}"\
-                        .format(type(ports).__name__)
+                        .format(ports)
                 if isinstance(ports, Value):
-                    msg += " (did you mean `ports=(<signal>,)`?)"
+                    msg += " (did you mean `ports=(<signal>,)`, rather than `ports=<signal>`?)"
                 raise TypeError(msg)
             mapped_ports = []
             # Lower late bound signals like ClockSignal() to ports.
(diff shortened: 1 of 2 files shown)


nmigen/hdl/ir.py
```diff
@@ -532,6 +532,13 @@ def prepare(self, ports=None, missing_domain=lambda name: ClockDomain(name)):
         if ports is None:
             fragment._propagate_ports(ports=(), all_undef_as_ports=True)
         else:
+            # typecheck ports is either list or tuple (#362)
+            if not isinstance(ports, tuple) and not isinstance(ports, list):
```
whitequark, line 536: This isn't a major issue here, but I prefer to have typechecking code at the very beginning of the function so that it "fails fast" (before any other errors can arise).

nmigen/hdl/ir.py
```diff
@@ -532,6 +532,13 @@ def prepare(self, ports=None, missing_domain=lambda name: ClockDomain(name)):
...
+                msg = "`ports` must be either a list or a tuple, not {!r}"\
+                        .format(type(ports).__name__)
```
whitequark, line 538: The nMigen convention is to use repr of the value in the error, not the name of the type. The reason is that it is often easier to recognize where `1024` came from than an abstract `int`. I would accept a PR that changes every diagnostic to report *both*, but in absence of that let's consistently use repr of the value.

nmigen/hdl/ir.py
```diff
@@ -532,6 +532,13 @@ def prepare(self, ports=None, missing_domain=lambda name: ClockDomain(name)):
...
+                if isinstance(ports, Value):
+                    msg += " (did you mean `ports=(<signal>,)`?)"
```
whitequark, line 540: This suggestion should be more specific, since a typo is often not obvious:
```suggestion
                    msg += " (did you mean `ports=(<signal>,)` rather than `ports=(<signal>)`?)"
```

nmigen/hdl/ir.py
```diff
@@ -532,6 +532,13 @@ def prepare(self, ports=None, missing_domain=lambda name: ClockDomain(name)):
         if ports is None:
             fragment._propagate_ports(ports=(), all_undef_as_ports=True)
         else:
+            # typecheck ports is either list or tuple (#362)
```
whitequark, line 535: In general, I avoid referencing bug numbers in code comments. Bugs can have a *lot* of comments in them and it's not always clear which of them apply. If a bug is important enough to mention in a comment, it's important enough to explain inline or at least in commit message (and as a bonus, doesn't require the reader to be online, GitHub available, etc). But in this case, the reference to the bug adds nothing as this kind of typechecking is pervasive in the codebase.

nmigen/hdl/ir.py
awygle: I'd like to keep this here, because it leverages the existing `if ports is None:` check. Otherwise we're checking for None redundantly (as we'd need that in the typecheck to avoid a false negative). I view this as being at the top of the "ports exists" branch. I don't feel _too_ strongly about this - I'll ultimately do what you say of course - but I wanted to explain the reasoning.

whitequark: Fine to keep as-is by me.

nmigen/hdl/ir.py
```diff
@@ -532,12 +532,12 @@ def prepare(self, ports=None, missing_domain=lambda name: ClockDomain(name)):
         if ports is None:
             fragment._propagate_ports(ports=(), all_undef_as_ports=True)
         else:
-            # typecheck ports is either list or tuple (#362)
+            # typecheck ports is either list or tuple
```
whitequark, line 535: Now this comment is just tautological with the code that follows it :)

awygle: :P

Issue #181, opened 2019-08-22
Reconsidering AsyncFIFO and resets


nmigen-issue-migration:
<a href="https://github.com/whitequark"><img src="https://avatars1.githubusercontent.com/u/54771?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Thursday Aug 22, 2019 at 02:05 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/180_

----

Both ports of a SyncFIFO belong to the same control set, which means that the entire FIFO is always reset at once.

However, the ports of AsyncFIFO belong to different control sets, which means that, if either if the control sets includes a reset, only half of the FIFO will get reset; and as a result, stale data will be read.

In some cases, this problem may be alleviated by using a reset synchronizer: one port is chosen as the one that resets the entire FIFO, its reset signal is brought into the other clock domain, and used to reset either the entire clock domain, or only the port state. However, if the duration of the reset pulse at the leading port is shorter than the clock period of the following port, the following port will not be reset properly, and stale data will be read. In particular this happens if the clock to one of the ports is gated completely.

In some cases, this problem may be further alleviated by using asynchronous reset (of course, with synchronous release) for the entire clock domain of the following port. However, this carries the restriction that the entire domain has to be reset: since the assertion of the reset is not synchronized to the clock of the following port, it incurs metastability. Further, many FPGA elements will not be inferred from logic that is asynchronously reset, notably BRAMs and DSP blocks, severely penalizing this design.

<hr>

It seems that there is no easy solution to the issue of resetting AsyncFIFO reliably in all circumstances, making it a completely safe design element that never spuriously produces stale data. I would suggest the following:

  1. AsyncFIFO should not have a reset at all, that is, it should not have a signal whose stated purpose is to ensure that one or both pointers becomes zero.
  2. The read side *only* of AsyncFIFO should have a *synchronous clear* signal that sets the read pointer to the same value as the write pointer, therefore flushing the buffer of any stale data that might be stuck there. Because AsyncFIFO reads are synchronous, this clear signal works regardless of how the read side is clocked. Because this clear signal operates entirely in the read domain, it works correctly regardless of how the write domain is clocked--the write clock could even be stopped completely.

This solves everything described above:
  1. The two ports of AsyncFIFO are never reset, therefore neither of them is reset while the other is not. Therefore stale data is never read as a result of incorrectly sequenced reset.
[...]

PR #367, opened 2020-04-24
Implement r_rst output signal for AsyncFIFO

awygle:
This is the last bit of #181 which still needs to be done before that issue can be closed.

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -280,7 +280,7 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
     """.strip(),
     r_data_valid="Valid if ``r_rdy`` is asserted.",
     r_attributes="""
-    r_rst : out
+    r_rst : Signal, out
         Asserted when the FIFO is reset by the write-domain reset.
     """.strip(),
     w_attributes="")
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -281,7 +281,8 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
     r_data_valid="Valid if ``r_rdy`` is asserted.",
     r_attributes="""
     r_rst : Signal, out
-        Asserted when the FIFO is reset by the write-domain reset.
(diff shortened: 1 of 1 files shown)


nmigen/lib/fifo.py
```diff
@@ -279,7 +279,10 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         Always set.
     """.strip(),
     r_data_valid="Valid if ``r_rdy`` is asserted.",
-    r_attributes="",
+    r_attributes="""
+    r_rst : out
```
whitequark, line 283:
```suggestion
    r_rst : Signal, out
```

nmigen/lib/fifo.py
```diff
@@ -279,7 +279,10 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
...
+        Asserted when the FIFO is reset by the write-domain reset.
```
whitequark, line 284: Asserted *when*? That implies to me that it's something like a 1-clock pulse. In reality it is asserted *while*, right?

awygle: I wrote _while_ initially, but the write domain reset pulse can technically be shorter than a read domain clock cycle, so I felt like it was inaccurate. Probably needlessly pedantic though - you're right that _when_ implies a single pulse (which we could do, i guess).

whitequark: I think holding it while the reset is asserted is a lot more useful, since you can go to a pulse from that but not the other way around.

Regaring pedantry, I think pedantry around CDC is pretty much a necessity. I think it would be very useful if we wrote that it is asserted "while the write-domain reset is, but no less than one read-domain clock cycle" or something like that.

whitequark: (No less than two in practice, but do we want to commit to that?)

awygle: OK, I agree with that. "Asserted while the FIFO is being reset by the write-domain reset (for at least one read-domain clock cycle)"

whitequark:
Thank you! This is a very nice addition that will enable Glasgow to, finally, support async applets properly.

PR #368, opened 2020-04-24
Deprecate `Record.connect`

### LINKED ISSUE #342: Deprecate `Record.connect`
`Record.connect` is very rarely useful, but _looks_ broadly useful. It should be deprecated (and ideally replaced with something better).

awygle:
Deprecate the unfortunate `Record.connect` method, and the `Direction` functionality of `Record` as well. Resolves #342 .

### DIFF
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -1,6 +1,7 @@
 from enum import Enum
 from collections import OrderedDict
 from functools import reduce
+import warnings
 
 from .. import tracer
 from .._utils import union, deprecated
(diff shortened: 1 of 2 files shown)


whitequark:
I'm wondering if we should hold off merging this until #355 is done, since until #355 is done, we really can't encourage people to add their own methods to `Record`.

awygle:
That is true (hence the deprecation message saying to connect fields manually). I am OK with holding off.

whitequark:
Since (as discussed on the last meeting) we're likely going to deprecate Record entirely and instead provide packed structs and interfaces, I'm closing this.

Issue #217, opened 2019-09-20
r_data is undefined after a simultaneous write to an empty AsyncFIFO


nmigen-issue-migration:
<a href="https://github.com/whitequark"><img src="https://avatars1.githubusercontent.com/u/54771?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Friday Sep 20, 2019 at 20:02 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/217_

----

The behavior of a dual port block RAM with read and write ports in different clock domains is not defined during a simultaneous read and write to the same address. (In fact it is not possible to make it defined.) However, this is exactly what is happening when an element is written into an empty `AsyncFIFO[Buffered]`, since the produce and consume pointers are equal. If the phase of read and write clocks is just wrong, the output register of the BRAM could end in a metastable state.

It is not clear how this can be fixed. A workaround is to use `AsyncFIFOBuffered` instead, which resynchronizes the BRAM output.

whitequark:
To fix this problem, we can add a small amount of logic that prevents this kind of condition from arising. We can convince ourselves that it's sound using formal verification.

PR #369, opened 2020-04-24
[WIP] AsyncFIFO: disable read from dual-port Memory when FIFO empty

awygle:
Avoids CDC hazard described in issue #217.


awygle:
I'm not comfortable doing this until #387 is implemented, so I'm just going to close this for now and put any time I'd have spent on this into that instead.

PR #382, opened 2020-05-07
vendor: add bit file generation for machXO2

trabucayre:
As others lattice and specially ECP5, machXO2 FPGA, a .bit file may be used to direct SRAM programming instead of FLASH programming.
Advantages are:
- extends FLASH lifetime
- reduces programming time.


whitequark:
Thank!

Issue #393, opened 2020-05-29
Signal decoder inference for Enums doesn't work for Layout members


shawnanastasio:
*This is a duplicate of https://github.com/m-labs/nmigen/issues/333 that I've moved here since this seems to be the correct upstream repository.*

When creating a signal with `shape` set to an Enum-derived type, the decoder function is automatically inferred to a built-in function that turns the Enum value and name into a nice human-readable string.

When defining a layout with an field's shape set to an Enum-derived type though, the decoder inference doesn't happen. As far as I can see, this is because the `Layout` constructor will cast all fields' shapes with `Shape.cast`, resulting in the original Enum type being lost.

https://github.com/nmigen/nmigen/blob/26a15b31f7fc278783cf9286d5af1877edc8290f/nmigen/hdl/rec.py#L48-L50

This results in an actual `Shape` object being passed to the `Signal` constructor instead of an Enum, and thus the built in Enum decoder can't be used.

I can think of a few ways to fix this, but wanted to hear the maintainers' thoughts to see if they have a preferred way of going about it.

My first idea is to not cast shapes in the `Layout` constructor, but instead check whether the user-provided type is a valid shape, probably by calling Shape.cast, discarding the value, and checking for an exception.

The other idea is to add a field to `Shape` which stores the original type that a given object was casted from. This would also require some extra logic in the `Signal` constructor to not only check the shape's type when checking for default decoders, but also checking the original type of casted `Shape` objects.

Curious to hear any thoughts on this.

**Update**: To test out the first idea, I simply removed the `shape =` from line 50 of the above code and it seems to behave as expected. If this is deemed an acceptable solution I can create a PR.

whitequark:
> As far as I can see, this is because the `Layout` constructor will cast all fields' shapes with `Shape.cast`, resulting in the original Enum type being lost.

Yep.

> My first idea is to not cast shapes in the `Layout` constructor, but instead check whether the user-provided type is a valid shape, probably by calling Shape.cast, discarding the value, and checking for an exception.

That seems like a reasonable solution which is otherwise inline with the current implementation.

shawnanastasio:
Though it seems to fix the original issue, removing the shape cast assignment causes a few test cases to fail:
```
======================================================================
FAIL: test_slice_tuple (nmigen.test.test_hdl_rec.RecordTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #394, opened 2020-05-31
hdl.rec: don't save casted shapes in Layout constructor

shawnanastasio:
Don't save the result of Shape.cast in the Layout constructor.
This allows the original user-provided shape to be preserved,
while still checking its validity.

This fixes Enum decoder inference for Signals (#393).

### DIFF
--- a/nmigen/test/test_hdl_rec.py
+++ b/nmigen/test/test_hdl_rec.py
@@ -12,6 +12,11 @@ class UnsignedEnum(Enum):
 
 
 class LayoutTestCase(FHDLTestCase):
+    def assertFieldEqual(self, field, expected):
+        (shape, dir) = field
+        shape = Shape.cast(shape)
+        self.assertEqual((shape, dir), expected)
+
     def test_fields(self):
         layout = Layout.cast([
             ("cyc",  1),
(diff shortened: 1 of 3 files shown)


shawnanastasio:
Looks like I missed some tests, let me fix those up

nmigen/test/test_hdl_rec.py
```diff
@@ -12,6 +12,11 @@ class UnsignedEnum(Enum):
 
 
 class LayoutTestCase(FHDLTestCase):
+    def compare_field(self, field, expected):
```
whitequark, line 15:
```suggestion
    def assertFieldEqual(self, field, expected):
```

nmigen/test/test_lib_io.py
```diff
@@ -4,53 +4,59 @@
 from ..back.pysim import *
 from ..lib.io import *
 
+def compare_layout(test_case, layout, expected):
```
whitequark, line 7: This could probably be a method of `PinLayoutTestCase`, and the rest could inherit from it.

shawnanastasio:
Updated to address review comments

whitequark:
Could you add a test that checks for the bug you're fixing here, i.e. the Enum decoder functionality?

shawnanastasio:
Added a test for the enum decoder in `RecordTestCase`

whitequark:
Thank you for your contribution!

PR #401, opened 2020-06-05
Preserve allow_reset_less when remapping ResetSignal. Closes #400

### LINKED ISSUE #400: ResetSignal's allow_reset_less fails on renamed reset_less domains
If a module uses a ResetSignal(allow_reset_less=True) on a clock domain which is later renamed into a reset-less domain, the ResetSignal still refers to the now non-existent reset:

```python
from nmigen import (Elaboratable, Module, Signal,
                    ResetSignal, ClockDomain, DomainRenamer)
from nmigen.back import verilog


class Mod(Elaboratable):
    def __init__(self):
        self.x = Signal()

        m = Module()
        s = ResetSignal(domain="x", allow_reset_less=True)
        m.d.comb += self.x.eq(s)
        return m


class Top(Elaboratable):
        m = Module()
        # Change reset_less=False to fix
        cd_y = ClockDomain("y", reset_less=True)
        m.domains += cd_y
        m.submodules += DomainRenamer({"x": "y"})(Mod())
        return m


if __name__ == "__main__":
    verilog.convert(Top())
```

```
<very long traceback snip>
nmigen.hdl.cd.DomainError: Signal (rst y) refers to reset of reset-less domain 'y'
```

adamgreig:
See #400

### DIFF
--- a/nmigen/hdl/xfrm.py
+++ b/nmigen/hdl/xfrm.py
@@ -453,7 +453,8 @@ class DomainRenamer(FragmentTransformer, ValueTransformer, StatementTransformer)
 
     def on_ResetSignal(self, value):
         if value.domain in self.domain_map:
-            return ResetSignal(self.domain_map[value.domain])
+            return ResetSignal(self.domain_map[value.domain],
+                               allow_reset_less=value.allow_reset_less)
         return value
 
     def map_domains(self, fragment, new_fragment):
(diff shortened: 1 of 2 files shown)


whitequark:
Could you add a test please?

adamgreig:
I've added a test which fails without the change and passes with, though it doesn't explicitly check that the newly created ResetSignal still has allow_reset_less set to True as I wasn't sure how to get at that signal.

whitequark:
This test is totally fine, thanks!

PR #410, opened 2020-06-22
Replace pnmainc with diamondc

jeanthom:
According to Lattice [support site](http://www.latticesemi.com/en/Support/AnswerDatabase/5/5/2/5522) diamondc sets the env variables required by Lattice tools. pnmainc also tends to fail without emitting logs even when `NMIGEN_verbose=1`.

### DIFF
--- a/nmigen/vendor/lattice_ecp5.py
+++ b/nmigen/vendor/lattice_ecp5.py
@@ -43,7 +43,7 @@ class LatticeECP5Platform(TemplatedPlatform):
     -----------------
 
     Required tools:
-        * ``pnmainc``
+        * ``diamondc``
         * ``ddtcmd``
 
     The environment is populated by running the script specified in the environment variable
@@ -177,7 +177,7 @@ class LatticeECP5Platform(TemplatedPlatform):
 
     _diamond_required_tools = [
         "yosys",
-        "pnmainc",
+        "diamondc",
         "ddtcmd"
     ]
     _diamond_file_templates = {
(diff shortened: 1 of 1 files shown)


cr1901:
Was this tested on Windows? I have a feeling this is likely to break on Windows based on my own experiences with Diamond, and that `pnmainc` is in fact the correct binary to use. I've reproduced my notes below:

```
# Lattice documentation provides 2 ways each for both Windows and Linux
# on how to set up environments so that the command-line tools can be invoked
# either from a TCL prompt or from an OS shell:
# Page 143-144 of Lattice Diamond User Guide (3.10) (UG)
# Page 2358-2359 of Lattice Diamond 3.10 Help (Help)
#
# The UG and Help give conflicting information wrt what environment variables
# need to be set/are set for you, and what scripts/executables are available
# from TCL or an OS shell after the environment set up. None of the 4
# documented methods are completely correct.
#
# What I have found works is the following:
# 1. Use pnmainc for everything. It exists even on Linux, which suggests using
# diamondc in the UG (and pnmainc in the Help). Setting up the environment
# variables for each OS (UG) will allow you to run commands from an OS shell.
# But you get that for free from pnmainc, in addition to TCL scripting.
#
# 2. On Windows, add Diamond binary directory to the path, and run pnmainc.
# pnmainc will set the remaining environment variables for you as suggested
# in the UG and Help, and will find its DLLs automatically by virtue of being
# on the path.
#
# 3. On Linux, contrary to what Help says, pnmainc will not set the remaining
# required environment variables (not to mention won't find it's shared
# libraries), so you need to source the provided diamond_env script in the
# Diamond binary directory. Set $bindir env variable, run diamond_env, and then
# invoke pnmainc.
```

jeanthom:
Thanks for sharing your notes. It was effectively only tested on Linux, I thought Lattice would have included a Windows version of diamondc for consistency reasons.

I can't spend time debugging this on Windows so I'm closing the PR.

Issue #409, opened 2020-06-22
_yosys fails to parse Yosys version from fomu-toolchain 


alanvgreen:
I have been yosys and nextpnr-ice40 from v1.5.6 of the fomu-toolchain, downloaded from https://github.com/im-tomu/fomu-toolchain/releases/tag/v1.5.6

The version number reported by the fomu-toolchain build of yosys is not compatible with the version number parser in _yosys.py. Here is the traceback from calling platform.build():

```
Traceback (most recent call last):
  File "./video.py", line 203, in <module>
    buildAndRunTest(DEMOS[args.demo], RESOLUTIONS[args.resolution], args.seed, args.optimize, args.program)
  File "./video.py", line 176, in buildAndRunTest
    platform.build(demo(resolution),
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/build/plat.py", line 90, in build
    plan = self.prepare(elaboratable, name, **kwargs)
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/build/plat.py", line 163, in prepare
    return self.toolchain_prepare(fragment, name, **kwargs)
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/build/plat.py", line 437, in toolchain_prepare
    render(content_tpl, origin=content_tpl))
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/build/plat.py", line 419, in render
    return compiled.render({
  File "/home/osboxes/ib/lib/python3.8/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "/home/osboxes/ib/lib/python3.8/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/home/osboxes/ib/lib/python3.8/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 2, in top-level template code
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/build/plat.py", line 314, in emit_debug_verilog
    return verilog._convert_rtlil_text(rtlil_text,
  File "/home/osboxes/ib/lib/python3.8/site-packages/nmigen/back/verilog.py", line 10, in _convert_rtlil_text
    yosys = find_yosys(lambda ver: ver >= (0, 9))
[...]

PR #411, opened 2020-06-23
_yosys.py: handle unparseable versions

alanvgreen:
Do not use yosys binaries with unparseable version numbers. This ensures
that nmigen always knows what version of yosys it is generating RTLIL
for.

The effect of this change is that if the version number of the system
yosys is unparsable, nmigen will attempt to fallback to the builtin
Yosys.

Issue #409.

### DIFF
--- a/nmigen/_yosys.py
+++ b/nmigen/_yosys.py
@@ -53,6 +53,8 @@ class YosysBinary:
 
         Returns
         -------
+        ``None`` if version number could not be determined, or a 3-tuple if it could.
+
         major : int
             Major version.
         minor : int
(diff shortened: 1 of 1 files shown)


nmigen/_yosys.py
```diff
@@ -208,8 +213,10 @@ def find_yosys(requirement):
                              "an unrecognized clause {!r}"
                              .format(clause))
     for proxy in proxies:
-        if proxy.available() and requirement(proxy.version()):
-            return proxy
+        if proxy.available():
+            version = proxy.version()
+            if version and requirement(version):
```
whitequark, line 218:
```suggestion
            if version is not None and requirement(version):
```

nmigen/_yosys.py
```diff
@@ -53,6 +53,8 @@ def version(cls):
 
         Returns
         -------
+        ``None`` if version number could not be determined, or a 3-tuple if it could.
```
whitequark, line 56:
```suggestion
        ``None`` if version number could not be determined, or a 3-tuple ``(major, minor, distance)`` if it could.
```

alanvgreen:
This is my first github code review - let me know if you'd prefer me to squash these 3 changes into one.

whitequark:
I've squashed it through GitHub's UI. Thanks for the patch!

Issue #412, opened 2020-06-24
LICENSE.txt and copyright


Fatsie:
Currently the LICENSE.txt file contains only copyright for M-Labs:
```
Copyright (C) 2011-2019 M-Labs Limited
```
I would like that it is made clear that the copyright of each commit is held by it's author or something like that.
I think this issue is valid also for the other nmigen-* repos.

whitequark:
Yeah, we should fix that.

mithro:
It would also be good if GitHub correctly identified the license in the repo.

whitequark:
> I would like that it is made clear that the copyright of each commit is held by it's author or something like that.
> I think this issue is valid also for the other nmigen-* repos.

Is there some boilerplate I can use?

> It would also be good if GitHub correctly identified the license in the repo.

Looks like the GitHub `licensee` gem cannot identify the 2-clause BSD license we're using. Any advice?

alanvgreen:
When I remove the final paragraph of LICENSE.txt, `licensee` recognizes the remaining text as the 2-clause BSD license.

IANAL, but...

It seems the final paragraph is not part of the license. It could be moved to a section of the README.md or some other file.

whitequark:
Hm, I see. Should we take a page from Rust's book? See their [LICENSE-MIT](https://github.com/rust-lang/rust/blob/master/LICENSE-MIT) and [COPYRIGHT](https://github.com/rust-lang/rust/blob/master/COPYRIGHT) files.

alanvgreen:
I think that would work for github, and also give room to explain/document anything you wanted about copyrights and licensing.

I was reading the final paragraph again: "All submissions are strongly encouraged to use the two-clause BSD license reproduced above." If it were me, I would insist on all contributions to this repo being 2-clause BSD licensed, because I don't understand the implications of mixing 2-clause BSD and other licenses. It sounds complicated and boring. However, this is just my opinion - IANAL, am definitely operating outside my area of expertise, and it's not my IP.

Happy to put together a patch if you give me a general direction.

whitequark:
> I would insist on all contributions to this repo being 2-clause BSD licensed, because I don't understand the implications of mixing 2-clause BSD and other licenses

Yes, we should require that. That clause was cargo-culted from Migen and really shouldn't have ever been there. There is no code that isn't under 2-clause BSD anywhere in nMigen, anyway.

Please do prepare a PR!

mithro:
[...]

PR #420, opened 2020-07-02
Update license and copyright info

alanvgreen:
Remove non-license explanatory text from LICENSE.txt.

Create CONTRIBUTING file with instructions and notes for contributors.

This change relates to issue #412

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,3 +1,4 @@
+Copyright (C) 2019-2020 whitequark
 Copyright (C) 2011-2019 M-Labs Limited
 
 Redistribution and use in source and binary forms, with or without modification,
@@ -19,10 +20,3 @@ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
 ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-
-Other authors retain ownership of their contributions. If a submission can
-reasonably be considered independently copyrightable, it's yours and we
-encourage you to claim it with appropriate copyright notices. This submission
-then falls under the "otherwise noted" category. All submissions are strongly
-encouraged to use the two-clause BSD license reproduced above.
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -1,0 +1,23 @@
+CONTRIBUTIONS and CODE REVIEWS
+
+We love contributions from the community. Please send pull requests and raise
+issues! All code submissions require review. We use GitHub pull requests for
+this purpose.
+
+
+COPYRIGHTS and LICENSE
+
+nMigen is licensed under the 2-clause BSD license, which is contained in the
+LICENSE.txt file.
+
+All authors retain copyright ownership of their contributions.
+
+Authors should note that all contributions are licensed under the 2-clause BSD
+license.
+
+
+CONTRIBUTORS
+
+The complete list of contributors is contained within the git repository.
+Tools such as the git log command can be used to see all contributions and
+contributors.


CONTRIBUTING
```diff
@@ -0,0 +1,27 @@
+CONTRIBUTIONS and CODE REVIEWS
+
+We love contributions from the community. Please send pull requests and raise
+issues! All code submissions require review. We use GitHub pull requests for
+this purpose.
+
+
+COPYRIGHTS and LICENSE
+
+Except as otherwise noted in individual files, nMigen is licensed under the
```
whitequark, line 10: I don't believe we currently have any such notices, and I would *much* prefer to not have them in the future, so let's leave that part out. nMigen isn't a massive amalgamation that has to contend with things like large differently licensed vendored dependencies; it's a fairly small library.

CONTRIBUTING
```diff
@@ -0,0 +1,27 @@
...
+2-clause BSD license, which is contained in the LICENSE.txt file.
+
+All authors retain copyright ownership of their contributions.
+
+Authors should note, that unless explicitly excepted, all contributions are
```
whitequark, line 15: Same here.

CONTRIBUTING
```diff
@@ -0,0 +1,27 @@
...
+licensed under the 2-clause BSD license.
+
+
+CONTRIBUTORS
+
+Major contributors to nMigen are:
```
whitequark, line 21: Let's leave this part out; the git history alone is more than sufficient.

alanvgreen: Done. 

Was just thinking: that if you wanted something like this in future, rather than "major contributors" it might be "Contributors with more than 100 PRs".

whitequark: I'm not sure if PR count really matters. In general I don't like having two classes of contributors for any reason.

CONTRIBUTING
```diff
@@ -0,0 +1,23 @@
+CONTRIBUTIONS and CODE REVIEWS
```
whitequark, line 1: This file should be caled `CONTRIBUTING.txt` for consistency.

LICENSE.txt
```diff
@@ -19,10 +19,3 @@ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
 ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-
-Other authors retain ownership of their contributions. If a submission can
-reasonably be considered independently copyrightable, it's yours and we
-encourage you to claim it with appropriate copyright notices. This submission
-then falls under the "otherwise noted" category. All submissions are strongly
-encouraged to use the two-clause BSD license reproduced above.
```
whitequark, line 28: Since this has been moved to `CONTRIBUTING.txt`, should the copyright lines above also be removed? After all we say that the ownership of contributions is indicated by the git history.

CONTRIBUTING
alanvgreen: Done.

LICENSE.txt
alanvgreen: Do you mean the M-Labs copyright notice? I think that needs to stay, according to point 1. of license. 

You could _add_ to the that list of copyright holders. I was a little surprised your name isn't there already, but I don't understand all the history.

Digression, doesn't have to be resolved in this change: 

(IANAL, but) my understanding is that the exact names in the list on the license will have little effect on the ways in which nMigen may be used - each and every contributor owns the copyright for their contributions, and each and every contributor has licensed their contribution under this license text. 

Where the list of copyright holders in the binary does have more of an effect is with giving credit when the source or binaries are distributed without git history. For instance, if someone were to use nMigen in their binary, the list of copyright holders might be the only remaining record of who wrote nMigen. Similar kind of situation if someone redistributed nMigen in a zip file.

I don't know how you feel about these scenarios, but I'd be little outraged on your behalf if there was a distribution of nMigen without whitequark's name.

whitequark: Ah right. I think `Copyright (C) 2019-2020 whitequark` should be added then.

alanvgreen: Done!

Issue #421, opened 2020-07-04
Adding Shapes gives counterintuitive results


awygle:
`Shape`s implement `__add__`, but adding them is currently not useful:

```python
>>> from nmigen import *
>>> x = Signal(16)
>>> y = Signal(32)
>>> s = x.shape() + y.shape()
>>> s
(16, False, 32, False)
```

The expected result here (I argue) would be `(48, False)`, but (probably because `Shape` is a named tuple under the hood) we get the not-very-useful result shown above.

Obviously in this case you can do `(x+y).shape()` but that's not always easy or possible (I hit a particular case involving an `Enum`)

whitequark:
So there are two unrelated issues here. First, that currently you can do `shape + shape` and the result is never desirable. We should fix that by not using `namedtuple` while still allowing to destructure shapes using `width, signed = shape`. (Is that possible? I think we can override `__iter__` to do it...)

Second, the idea of overloading the `+` operation on shapes to do something not really related to addition. It doesn't seem like a good idea to me, but we can discuss it in a different issue.

awygle:
I agree it's a bad idea. We should stop implementing extraneous operators for `Shape`, and that's all.

awygle:
Should we preserve comparing `Shape`s to tuples of `(int, bool)`? This currently works (and is used in the test suite) but feels like a bit of a footgun.

whitequark:
> This currently works (and is used in the test suite)

Where?

whitequark:
Oh sorry, you mean comparing *for equality*, not for ordering. Yeah let's preserve that, we can always deprecate it later if it turns out to cause issues.

PR #422, opened 2020-07-06
Don't inherit Shape from NamedTuple

awygle:
Fixes issue #421 .

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -32,7 +32,23 @@ class DUID:
         DUID.__next_uid += 1
 
 
-class Shape(typing.NamedTuple):
+class ShapeIter:
+    def __init__(self, width, signed):
+        self.width = width
+        self.signed = signed
+        self._i = 0
+
+    def __next__(self):
+        if self._i == 0:
+            self._i += 1
+            return self.width
+        elif self._i == 1:
+            self._i += 1
+            return self.signed
+        else:
+            raise StopIteration
+
+class Shape:
     """Bit width and signedness of a value.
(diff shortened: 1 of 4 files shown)


nmigen/hdl/ast.py
```diff
@@ -55,8 +71,15 @@ class Shape(typing.NamedTuple):
...
-    signed: bool = False
+    def __init__(self, width=1, signed=False):
+        if not isinstance(width, int) or width < 0:
+            raise TypeError("Width must be a non-negative integer, not {!r}"
+                            .format(width))
+        self.width = width
+        self.signed = signed
+    
+    def __iter__(self):
+        return ShapeIter(self.width, self.signed)
```
whitequark, line 82: Why not `return iter((self.width, self.signed))`?

awygle:
Because I'm bad at python and didn't know iter() existed, and also I've
been writing Rust a lot and this is how you have to do it there.

nmigen/test/test_hdl_ast.py
```diff
@@ -39,6 +39,11 @@ def test_make_wrong(self):
                 msg="Width must be a non-negative integer, not -1"):
             Shape(-1)
 
+    def test_compare_wrong(self):
+        with self.assertRaises(TypeError,
+                msg="Shapes may be compared with other Shapes and (int, bool) tuples, not (2, 3)"):
+            Shape(1, True) == (2, 3)
```
whitequark, line 45: Needs a test for the second `raise TypeError`.

whitequark:
Thanks!

Issue #425, opened 2020-07-08
Support for PLL primitives


GuzTech:
The goal of this issue is to document the kinds of PLL primitives of the various FPGA manufacturers, so that we can support them in nmigen.

---

## Lattice
Lattice PLLs are different for the different FPGA families.

### iCE40 (LP/LM/HX/Ultra/UltraLite/UltraPlus)
The iCE40 family has five types of PLLs. They are all capable of shifting the output clock by 0 and 90 degrees, and allow fine delay adjustments of up to 2.5 ns (typical) in 150 ps increments (typical).

#### SB_PLL40_CORE
Can be used if the source clock of the PLL originates on the FPGA or is driven by an input pad that is not the bottom IO bank (bank 2).

#### SB_PLL40_PAD
Can be used if the source clock of the PLL is driven by an input pad that is located in the bottom or top IO bank (banks 2 and 0 respectively). When using this PLL, the source clock cannot be used anymore.

#### SB_PLL40_2_PAD
Can be used if the source clock of the PLL is driven by an input pad that is located in the bottom or top IO bank (banks 2 and 0 respectively). This PLL outputs the requested clock as well as the the source clock.

#### SB_PLL40_2F_CORE
Can generate two different output frequencies. Can be used if the source clock of the PLL originates on the FPGA.

#### SB_PLL40_2F_PAD
Can generate two different output frequencies. Can be used if the source clock of the PLL is driven by an input pad that is located in the bottom or top IO bank (banks 2 and 0 respectively).

### ECP5
The ECP5 family has bunch of clocking elements, but only one type of PLL.

#### EHXPLLL
It has four outputs, but if the user wants hook the PLL feedback, then one of the outputs cannot be used in the fabric. It supports dynamic clock selection and control, dynamic phase adjustment, etc.

---

## Xilinx
Xilinx has two clock synthesis IPs as far as I know: PLL and Mixed-Mode Clock Manager (MMCM). The different FPGA families have different versions of these IPs.

### Spartan-6
The Spartan-6 family have Clock Management Tiles (CMTs), where each contain one PLL and two Digital Clock Managers (DCMs). The latter can be used to implement Delay Locked Loops, digital frequency synthesizers, digital phase shifters, or a digital spread spectrum.

There are two PLL primitives, that each have six clock outputs and a dedicated feedback output.

#### PLL_BASE
[...]

PR #426, opened 2020-07-08
[WIP] Initial implementation of ECP5 PLL instantiation

GuzTech:
This is the initial implementation where I took the liberty to be inspired by the [Litex implementation](https://github.com/enjoy-digital/litex/blob/8c572d2b3e3ccd2c98fd45fd69d425e7b3fa3191/litex/soc/cores/clock.py).

Signed-off-by: Oguz Meteer <info@guztech.nl>

### DIFF
--- a/nmigen/lib/clock.py
+++ b/nmigen/lib/clock.py
@@ -1,0 +1,133 @@
+from .. import *
+from ..build import Platform
+
+__all__ = ["ECP5PLL"]
+
+
+class ECP5PLL(Elaboratable):
+    """ECP5 PLL
+
+    Instantiates the EHXPLLL primitive, and provides up to three clock outputs. The EHXPLLL primitive itself
+    provides up to four clock outputs, but the last output (CLKOS3) is fed back into the feedback input.
+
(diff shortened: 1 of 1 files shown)


GuzTech:
I have no doubt that the implementation can be improved, but I'm not entirely sure how. My Python skills aren't the best, and there are parts that I'm a bit iffy about.

Some examples:
* In `register_clkin`, I don't know exactly how to check that the `clkin` parameter is a ClockSignal.
* Also in `register_clkin`, I had an intermediate signal `self.clkin` that I commented out. That actually caused me to not define a new Module in `elaborate`, but in `__init__`, since I needed to assign the `clkin` parameter to `self.clkin` combinatorially. I found that an intermediate signal is unnecessary, so I now instantiate a new Module and return in it in `elaborate`.
* In the [Litex implementation](https://github.com/enjoy-digital/litex/blob/8c572d2b3e3ccd2c98fd45fd69d425e7b3fa3191/litex/soc/cores/clock.py), they always feed the last clock output back into the feedback input. Maybe we should make this optional?

whitequark:
It's way too early to consider merging anything upstream. Let's discuss what API we want first in #425.

GuzTech:
Of course. I just posted this just to have some code that we could look at.

PR #430, opened 2020-07-13
write next state to vcd instead of curr state, since state is written just prior to commit

### LINKED ISSUE #429: values in vcd are zero despite simulation's nonzero values
Still haven't narrowed this one down to a reproducer that's all that useful:
http://lists.libre-riscv.org/pipermail/libre-riscv-dev/2020-July/008506.html

used revisions:
nmigen : 30e2f91176edcd1c8766c2cb11f413b9c77936b9
ieee754fpu : 610b4a381e70f45f5684cc281398ce77fb5441fa
nmutil : 3853df675a1e1db24950945f66b076266a7da409
soc : caceb716e9417ed8731ef08b7b260a4c077186b2

ieee754fpu, nmutil, and soc can be cloned using:

```bash
for i in ieee754fpu nmutil soc; do
    git clone --recursive git://git.libre-riscv.org/$i.git
done
```

activate a python3 venv:
```bash
<path/to/venv>/bin/activate
```

in each of nmigen, ieee754fpu, and nmutil, run:
```bash
python setup.py develop
```

in soc run:
```bash
make install
python src/soc/fu/div/test/test_pipe_caller.py # will fail the tests
```

the generated vcd file is `div_simulator.vcd` in soc's top level folder.

programmerjake:
Fixes #429

I tested it and it seems to work!
@whitequark You can add a regression test later if you want.

### DIFF
--- a/nmigen/sim/pysim.py
+++ b/nmigen/sim/pysim.py
@@ -348,7 +348,7 @@ class Simulator:
             for waveform_writer in self._waveform_writers:
                 for signal_state in self._state.pending:
                     waveform_writer.update(self._state.timeline.now,
-                        signal_state.signal, signal_state.curr)
+                        signal_state.signal, signal_state.next)
 
             # 2. commit: apply every queued signal change, waking up any waiting processes
             converged = self._state.commit()


whitequark:
So actually a better fix would be to accumulate `pending` throughout every delta cycle and then write things only once per step. Unless you need this fixed urgently I'd rather do that.

programmerjake:
Would you be willing to merge this fix while you finish off the better fix, since this one is pretty simple? I'm not 100% sure, but I think #429 is blocking several people in Libre-SOC to some extent.

programmerjake:
Thanks!

PR #433, opened 2020-07-13
docs: add install instructions for arch

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -68,6 +68,16 @@ Installing prerequisites
 
       If Yosys 0.9 (or newer) is not available, `build Yosys from source`_.
 
+   .. platform-choice:: arch
+      :altname: linux
+      :title: Arch Linux
+
+      Install Python, pip, GTKWave and Yosys by running:
+
+      .. code-block:: console
+
+         $ sudo pacman -S python python-pip gtkwave yosys
+
    .. platform-choice:: linux
       :title: Other Linux
 
@@ -129,6 +139,16 @@ Latest release
       .. code-block:: console
 
          $ pip3 install --user --upgrade nmigen
+
+   .. platform-choice:: arch
+      :altname: linux
+      :title: Arch Linux
+
+      |release:install|
+
+      .. code-block:: console
+
+         $ sudo pacman -S python-nmigen
 
 
 .. _install-snapshot:


FFY00:
Yosys will always be at the latest version since we follow the upstream closely. Arch is a rolling distribution so there are no installation with repo access to an older yosys.

As a packager I'd like to advise to install nmigen via the system package manager, this is the only proper way to install it system-wide. Does it make sense to do it here?

whitequark:
> Yosys will always be at the latest version since we follow the upstream closely. Arch is a rolling distribution so there are no installation with repo access to an older yosys.

Great!

> As a packager I'd like to advise to install nmigen via the system package manager, this is the only proper way to install it system-wide. Does it make sense to do it here?

Depends. Would Arch provide the latest release or the latest master?

FFY00:
Latest release, arch packages stable software.

whitequark:
Sure. Then you can add the Arch-specific instructions to the `Latest release` section, also with `:altname: linux`.

whitequark:
Note that, in the future (as in, when I figure out the least footgun-y path towards it) the primary mode of using nMigen would be through per-project virtualenvs. Once that happens, these installation instructions would be primarily targeted at developers who a are just trying it out and don't want to bother with relatively heavyweight project management yet.

docs/install.rst
```diff
@@ -68,6 +68,16 @@ Installing prerequisites
...
 
+   .. platform-choice:: arch
+      :altname: linux
+      :title: Arch Linux
+
+      Install GTKWave and Yosys by running:
+
+      .. code-block:: console
+
+         $ sudo pacman -S gtkwave yosys
```
whitequark, line 79: I think this should include Python (and pip) in case the reader wants to install an unreleased version of nMigen.

FFY00: Ah, right.

docs/install.rst
```diff
@@ -130,6 +140,14 @@ Latest release
 
          $ pip3 install --user --upgrade nmigen
 
+   .. platform-choice:: arch
+      :altname: linux
+      :title: Arch Linux
+
+      .. code-block:: console
```
whitequark, line 147: Missing `|release:install|`

whitequark:
Thanks!

PR #435, opened 2020-07-14
dsl: issue a warning if defining a case after the default

### LINKED ISSUE #404: Unused cases in a switch do not trigger warnings
Consider those two elaboratables:

```python
class Impl1(Elaboratable):
    def __init__(self):
        self.sel = Signal()
        self.A = Signal()
        self.B = Signal()

        m = Module()

        with m.Switch(self.sel):
            with m.Case():
                m.d.comb += [
                    self.A.eq(1),
                    self.B.eq(1),
                ]
            with m.Case(0):
                m.d.comb += [
                    self.A.eq(1),
                    self.B.eq(0),
                ]
            with m.Case(1):
                m.d.comb += [
                    self.A.eq(0),
                    self.B.eq(0)
                ]

        return m

class Impl2(Elaboratable):
    def __init__(self):
        self.sel = Signal()
        self.A = Signal()
        self.B = Signal()

        m = Module()

        with m.Switch(self.sel):
            with m.Case(0):
                m.d.comb += [
                    self.A.eq(1),
                    self.B.eq(0),
                ]
            with m.Case(1):
                m.d.comb += [
                    self.A.eq(0),
                    self.B.eq(0)
                ]
            with m.Case():
                m.d.comb += [
                    self.A.eq(1),
                    self.B.eq(1),
                ]

        return m
```

Because the default/wildcard case was declared first in Impl1, it prevents Case(0) and Case(1) from being matched. This is a bit misleading for newbies who would expect cases to be position agnostic (and therefore Impl1 having equivalent behaviour as Impl2).

As discussed on IRC, a sensible way of dealing with that issue would be to trigger warnings whenever some cases are unreachable. We still have to figure out how we can silence those warnings when they are undesired (think code generation).

FFY00:
Resolves #404

Signed-off-by: Filipe Laíns <lains@archlinux.org>

### DIFF
--- a/nmigen/hdl/dsl.py
+++ b/nmigen/hdl/dsl.py
@@ -306,6 +306,10 @@ class Module(_ModuleBuilderRoot, Elaboratable):
         src_loc = tracer.get_src_loc(src_loc_at=1)
         switch_data = self._get_ctrl("Switch")
         new_patterns = ()
+        for case in switch_data['cases']:
+            if not case:
+                warnings.warn("Defining a case after the default one, it will be ignored")
+                continue
         for pattern in patterns:
             if not isinstance(pattern, (int, str, Enum)):
                 raise SyntaxError("Case pattern must be an integer, a string, or an enumeration, "


whitequark:
Thank you for your contribution. I believe that #1059 will address the underlying issue.

PR #447, opened 2020-07-23
Add SSTL135D_I/SSTL15D_I/SSTL18D_I to the list of diff io types in ECP5

### LINKED ISSUE #446: nmigen generates _n signals when using diff pairs causing issues with nextpnr
Example code:
```python
from nmigen import *
from nmigen_boards.ecpix5 import ECPIX585Platform

class Top(Elaboratable):
        m = Module()

        ddr3 = platform.request("ddr3", 0)
        m.d.sync += ddr3.clk.o.eq(1)

        return m

if __name__ == "__main__":
    ECPIX585Platform().build(Top(), do_program=True)
```

Console output:
```
Warning: No clocks found in design
Warning: No clocks found in design
ERROR: cannot place differential IO at location PIOB
```

jeanthom:
This PR fixes #446 by adding the missing IO type.


miek:
Looks like `SSTL15D_I` and `SSTL18D_I` should be there too - worth adding those at the same time?

PR #449, opened 2020-07-24
Implement ValueCastable

awygle:
Implement the ValueCastable type as described in #355 .

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -18,7 +18,7 @@ __all__ = [
     "Value", "Const", "C", "AnyConst", "AnySeq", "Operator", "Mux", "Part", "Slice", "Cat", "Repl",
     "Array", "ArrayProxy",
     "Signal", "ClockSignal", "ResetSignal",
-    "UserValue", "ValueCastable", "memoized",
+    "UserValue", "ValueCastable",
     "Sample", "Past", "Stable", "Rose", "Fell", "Initial",
     "Statement", "Switch",
     "Property", "Assign", "Assert", "Assume", "Cover",
@@ -1284,31 +1284,11 @@ class UserValue(Value):
         return self._lazy_lower()._rhs_signals()
 
 
-def memoized(func):
-    """Decorator to memoize class methods.
-
-    Ensures the decorated method is called only once, with subsequent method calls returning the
-    object returned by the first first method call.
-
-    This decorator is required to decorate the ``as_value`` method of ``ValueCastable`` subclasses.
-    This is to ensure that nMigen's view of representation of all values stays internally
-    consistent.
-    """
-    @functools.wraps(func)
-    def wrapper_memoized(self, *args, **kwargs):
-        lowered_to = getattr(self, "__lowered_to", None)
-        if not hasattr(self, "__lowered_to"):
-            self.__lowered_to = func(self, *args, **kwargs)
-        return self.__lowered_to
-    wrapper_memoized.__memoized__ = True
-    return wrapper_memoized
-
-
 class ValueCastable:
     """Base class for classes which can be cast to Values.
 
-    A ``ValueCastable`` can be cast to ``Value``, meaning its recise representation does not have 
-    to be immediately known. This is useful in certain metaprogramming scenarios. Instead of 
+    A ``ValueCastable`` can be cast to ``Value``, meaning its recise representation does not have
+    to be immediately known. This is useful in certain metaprogramming scenarios. Instead of
     providing fixed semantics upfront, it is kept abstract for as long as possible, only being
     cast to a concrete nMigen value when required.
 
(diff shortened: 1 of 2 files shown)


awygle:
This PR does _not_ deprecate UserValue. If you'd rather that be included, let me know and I'll add it, but it seemed easier to review them separately.

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,53 @@ def _rhs_signals(self):
...
+    """Base class for classes which can be cast to Values.
+
+    A ``ValueCastable`` can be cast to ``Value``, meaning its recise representation does not have
+    to be immediately known. This is useful in certain metaprogramming scenarios. Instead of
+    providing fixed semantics upfront, it is kept abstract for as long as possible, only being
+    cast to a concrete nMigen value when required.
+
+    Note that it is necessary to ensure that nMigen's view of representation of all values stays 
+    internally consistent. The class deriving from ``ValueCastable`` must decorate the ``as_value``
+    method with the ``memoized`` decorator, which ensures that all calls to ``as_value``return the
```
whitequark, line 1297:
```suggestion
    method with the ``lowermethod`` decorator, which ensures that all calls to ``as_value``return the
```

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,53 @@ def _rhs_signals(self):
...
+    the first call to ``as_value``.
+    """
+    def __new__(cls, *args, src_loc_at=0, **kwargs):
+        self = super().__new__(cls)
+        if not hasattr(self, "as_value") or not hasattr(self.as_value, "_ValueCastable__memoized"):
+            raise TypeError("Classes deriving from `ValueCastable` must decorate the `as_value` "
+                            "method with the `ValueCastable.lowermethod` decorator")
+        return self
+
+    @abstractmethod
```
whitequark, line 1309: I don't think this works on a non-abstract class.

whitequark (changes requested):

awygle:
That should resolve the `as_value` subtleties around being an abstract method without using `ABCMeta` as the metaclass

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,51 @@ def _rhs_signals(self):
...
+    cast to a concrete nMigen value when required.
+
+    Note that it is necessary to ensure that nMigen's view of representation of all values stays 
+    internally consistent. The class deriving from ``ValueCastable`` must decorate the ``as_value``
+    method with the ``lowermethod`` decorator, which ensures that all calls to ``as_value``return the
+    same ``Value`` representation. If the class deriving from ``ValueCastable`` is mutable, it is
+    up to the user to ensure that it is not mutated in a way that changes its representation after
+    the first call to ``as_value``.
+    """
+    def __new__(cls, *args, src_loc_at=0, **kwargs):
```
whitequark, line 1302: `src_loc_at` doesn't seem to be used anywhere, is it a leftover from a past version of this code?

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,51 @@ def _rhs_signals(self):
...
+        self = super().__new__(cls)
```
whitequark, line 1303: Should this `__new__` method forward `*args, **kwargs`? (I'm not entirely sure, I would have to look up this part of the Python object model, but this is what I'd expect.)

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,51 @@ def _rhs_signals(self):
...
+        if not hasattr(self, "as_value"):
+            raise TypeError("Classes deriving from `ValueCastable` must override the `as_value` method")
```
whitequark, line 1305: I think it would be nice if the error message said which exact class it was. E.g. `Class 'Foo' deriving from ValueCastable must [etc]`

nmigen/hdl/ast.py
```diff
@@ -1280,6 +1284,51 @@ def _rhs_signals(self):
         return self._lazy_lower()._rhs_signals()
 
 
+class ValueCastable:
+    """Base class for classes which can be cast to Values.
+
+    A ``ValueCastable`` can be cast to ``Value``, meaning its recise representation does not have
```
Ravenslofty, line 1290: s/recise/precise/

tests/test_hdl_ast.py
```diff
@@ -1025,6 +1025,51 @@ def test_lower_to_user_value(self):
...
+class MockValueCastableNoOverride(ValueCastable):
+    def __init__(self):
+        pass
+
+
+class ValueCastableTestCase(FHDLTestCase):
+    def test_not_decorated(self):
+        with self.assertRaisesRegex(TypeError,
+                r"^Classes deriving from `ValueCastable` must decorate the `as_value` "
+                "method with the `ValueCastable.lowermethod` decorator$"):
```
whitequark, line 1054: Python string qualifies should be present for every continuation. I.e. if you concatenate two chunks of a raw string it looks like `r"foo" r"bar"`.

whitequark (changes requested):

nmigen/hdl/ast.py
awygle: Bad copy-paste.

nmigen/hdl/ast.py
awygle: This section is slightly magic to me (overriding `__new__`, and especially the assignment of `self`). I agree with you that that makes more sense.

whitequark: `self` is an ordinary variable in Python with no special meaning in any context whatsoever. (What *does* have special meaning is the very first argument of a class method, but it has nothing to do with its *name*.)

whitequark: As for `__new__` I will double-check whether this is correct, so don't worry about that.

awygle: Except it's an error (`TypeError: object() takes no parameters`). So I guess not.

whitequark: For the onlookers: [this](https://hg.python.org/cpython/file/78d36d54391c/Objects/typeobject.c#l3034) is the reason for the strange behavior of `__new__` here.

whitequark (approved):

PR #450, opened 2020-07-24
nmigen.lib: Add RoundRobin

### LINKED ISSUE #448: [RFC] Adding RoundRobin to nmigen.lib
* oMigen had RoundRobin, nMigen hasn't
 * We already have part of the implementation: https://github.com/nmigen/nmigen-soc/blob/master/nmigen_soc/wishbone/bus.py#L379-L402

jeanthom:
Fixes #448


nmigen/compat/genlib/roundrobin.py
```diff
@@ -0,0 +1,22 @@
...
+            "use `nmigen.lib.roundrobin.RoundRobin`; note that RoundRobin does not "
+            "require a policy anymore, and that the `ce` attribute has been renamed"
+            "to `en`")
+class CompatRoundRobin(NativeRoundRobin):
+    def __init__(self, n, switching_policy=SP_WITHDRAW):
+        super().__init__(n)
+
+        if switching_policy == SP_CE:
+        	self.ce = Signal()
+        	self.comb += self.en.eq(self.ce)
```
whitequark, line 19: I think this doesn't actually work, `NativeRoundRobin` doesn't have `self.comb`. I think you can just copy the oMigen code wholesale into `nmigen.compat`.

jeanthom:
Currently implementing @jfng's suggestions (code styling, better doc, etc.).

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,50 @@
...
+    width : int
+        Number of requests.
+
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request.
+    en : Signal, in
```
whitequark, line 24: Why not let the end user of the module apply `EnableInserter` instead?

jeanthom: Good call.

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,47 @@
...
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request.
+    """
+    def __init__(self, *, width):
+        if not isinstance(width, int) or width < 0:
+            raise ValueError("Width must be a positive integer, not {!r}"
```
whitequark, line 27:
```suggestion
            raise ValueError("Width must be a non-negative integer, not {!r}"
```

Zero width should be explicitly allowed, this is useful for generic code.

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,47 @@
...
+    ----------
+    width : int
+        Number of requests.
+
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request.
```
whitequark, line 23: What happens if no request is active?

nmigen/compat/genlib/roundrobin.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+__all__ = ["RoundRobin", "SP_WITHDRAW", "SP_CE"]
+
+(SP_WITHDRAW, SP_CE) = range(2)
+
+@deprecated("instead of `migen.genlib.roundrobin.RoundRobin`, "
+            "use `nmigen.lib.roundrobin.RoundRobin`; note that RoundRobin does not "
+            "require a policy anymore, and that the `ce` attribute has been renamed"
+            "to `en`")
```
whitequark, line 12: This comment is outdated. Also, I would explicitly emit two different warning from `__init__` depending on what `switch_policy` is set to.

nmigen/lib/roundrobin.py
jeanthom: grant isn't updated and keeps its value. Should I add that to the doc?

whitequark: I think it would be better to take a page from `lib.coding` and make an explicit `valid` output. (Actually, `lib.coding` uses a weird `n` convention for that--I think `valid` would be far better.)

jeanthom: #dbd3926 adds a valid signal.

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,54 @@
...
+    valid : Signal(), out
+        Indicate if grant is valid.
```
whitequark, line 27: This comment should explain the conditions under which the grant is valid. (Right now it doesn't say anything the name doesn't already.)

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,54 @@
...
+            for i in range(self.width):
+                with m.Case(i):
+                    for pred in reversed(range(i)):
+                        with m.If(self.requests[pred]):
+                            m.d.sync += self.grant.eq(pred)
+                    for succ in reversed(range(i + 1, self.width)):
+                        with m.If(self.requests[succ]):
+                            m.d.sync += self.grant.eq(succ)
+
+        m.d.sync += self.valid.eq(reduce(or_, self.requests))
```
whitequark, line 52:
```suggestion
        m.d.sync += self.valid.eq(self.requests.any())
```

nmigen/lib/roundrobin.py
jeanthom: "Indicate if `grant` is a valid request that was emitted at the previous clock cycle."

That would be okay?

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,52 @@
...
+        Number of requests.
+
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request.
+    valid : Signal(), out
+        Indicate if grant is valid.
```
whitequark, line 25:
```suggestion
        Asserted if `grant` corresponds to an active request. Deasserted otherwise, i.e. if no requests are active.
```

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,52 @@
...
+    ----------
+    width : int
+        Number of requests.
+
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request.
```
whitequark, line 23:
```suggestion
        Number of the granted request. Indeterminate if `valid` is deasserted.
```

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+__all__ = ["RoundRobin"]
+
+
+class RoundRobin(Elaboratable):
+    """Round-robin scheduler.
+
+    For a given set of requests, the round-robin scheduler will
+    grant one request.
```
whitequark, line 11: I feel like this should be elaborated. Something like (phrasing needs to be improved):

> For a given set of requests, the round-robin scheduler will grant one request. Once it grants a request, if any other requests are active, it grants the next active request with a greater number, restarting from zero once it reaches the highest one. Use :class:`EnableInserter` to control when the scheduler advances.

nmigen/lib/roundrobin.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    width : int
+        Number of requests.
+
+    Attributes
+    ----------
+    requests : Signal(width), in
+        Set of requests.
+    grant : Signal(range(width)), out
+        Number of the granted request. Indeterminate if `valid` is deasserted.
```
whitequark, line 23: I gave it more thought and realized that we can't change the current behavior without impacting fairness. I.e. consider a case where either two requests are active simultaneously, or none at all. If `grant` would change in any way when there are no active requests, the scheduler won't act as round-robin anymore--e.g. if it got reset to zero, it would always select the first request in this scenario.

So we can safely say that:
```suggestion
        Number of the granted request. Does not change if there are no active requests.
```

nmigen/lib/roundrobin.py
jeanthom: I tweaked a little bit what you wrote. What do you think?

whitequark:
Great, thanks for the work! There's only one remaining concern I have: should `RoundRobin` live in its own module? Do we plan to have other schedulers? Or perhaps should it be categorized differently?

jeanthom:
That's the way it was in omigen because there was only one scheduler. I seriously doubt that we're going to have other schedulers in nMigen, but just in case we could name the module `scheduler`? Or keep it this way until someone has another scheduler to offer?

whitequark:
> I seriously doubt that we're going to have other schedulers in nMigen, but just in case we could name the module `scheduler`? Or keep it this way until someone has another scheduler to offer?

I think we can name the module `scheduler` and at the same time accept that there likely won't ever be a second one. The module name is still nice because `from nmigen.lib.scheduler import RoundRobin` is more clear on its purpose than `from nmigen.lib.roundrobin import RoundRobin`.

nmigen/compat/genlib/roundrobin.py
```diff
@@ -12,15 +12,15 @@ def __init__(self, n, switch_policy=SP_WITHDRAW):
         self.grant = Signal(max=max(2, n))
         self.switch_policy = switch_policy
         if self.switch_policy == SP_CE:
-            warnings.warn("instead of `migen.genlib.roundrobin.RoundRobin`, "
-                          "use `nmigen.lib.roundrobin.RoundRobin`; note that RoundRobin does not "
+            warnings.warn("instead of `migen.genlib.scheduler.RoundRobin`, "
```
whitequark, line 15:
```suggestion
            warnings.warn("instead of `migen.genlib.roundrobin.RoundRobin`, "
```
Find/replace error

nmigen/compat/genlib/roundrobin.py
```diff
@@ -12,15 +12,15 @@ def __init__(self, n, switch_policy=SP_WITHDRAW):
...
+                          "use `nmigen.lib.scheduler.RoundRobin`; note that RoundRobin does not "
                           "require a policy anymore but to get the same behavior as SP_CE you"
                           "should use an EnableInserter",
                           DeprecationWarning, stacklevel=1)
             self.ce = Signal()
         else:
-            warnings.warn("instead of `migen.genlib.roundrobin.RoundRobin`, "
-                          "use `nmigen.lib.roundrobin.RoundRobin`; note that RoundRobin does not "
+            warnings.warn("instead of `migen.genlib.scheduler.RoundRobin`, "
```
whitequark, line 22:
```suggestion
            warnings.warn("instead of `migen.genlib.roundrobin.RoundRobin`, "
```
Find/replace error

nmigen/compat/genlib/roundrobin.py
jeanthom: oops

nmigen/lib/scheduler.py
```diff
@@ -0,0 +1,60 @@
...
+    requests are active, it grants the next active request with
+    a greater number, restarting from zero once it reaches the
+    highest one.
+
+    Use :class:`EnableInserter` to control when the scheduler
+    is updated.
+
+    Parameters
+    ----------
+    width : int
```
whitequark, line 21: Ah, one last bikeshed. Talking about the "width" of a scheduler doesn't make much sense to me. Perhaps "size" or "count"?

PR #452, opened 2020-07-27
Convert tests to regex versions

GravesJake:


whitequark:
Thanks. Can you remove the implementations of `assertRaises` and `assertWarns` too?

GravesJake:
The ones that don’t contain a `msg` argument?

GravesJake:
Okay, there are a few calls left to the non-reflex versions that don’t contain a `msg ` argument. I’m guessing those should be changed over to refer versions as well then?

whitequark:
I mean these: https://github.com/nmigen/nmigen/blob/c75fa45fd80063ad5681f389fb038d6d98f4f0a0/nmigen/test/utils.py#L31-L46

whitequark:
> I’m guessing those should be changed over to regex versions as well then?

Nope, they should work fine as-is, with no additional changes.

whitequark:
Thank you!

PR #461, opened 2020-07-31
SSH Client Support via Paramiko

cr1901:
Add `execute_remote_ssh` method to `BuildPlan`, and `RemoteSshBuildProducts` class to `nmigen.build.run`. Also add `paramiko` to `extras_require`.

### DIFF
--- a/nmigen/build/run.py
+++ b/nmigen/build/run.py
@@ -12,6 +12,7 @@ import hashlib
 __all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts"]
 
 
+
 class BuildPlan:
     def __init__(self, script):
         """A build plan.
@@ -98,6 +99,62 @@ class BuildPlan:
 
         finally:
             os.chdir(cwd)
+
+    def execute_remote_ssh(self, root, hostname, *, connect_args = {}):
+        """
+        Execute build plan using the remote SSH strategy. Files from the build
+        plan are transferred via SFTP to the directory ``root`` on the remote
+        server ``hostname``. The ``paramiko`` SSH client will then run ``{script}.sh``.
+
+        ``hostname`` corresponds to the first (required) input argument of ``paramiko``'s
+        ``SSHClient.connect``, and ``connect_args`` is a dictionary that holds
+        all other input arguments to ``SSHClient.connect``
+        (`documentation <http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect>`_).
+
+        This method will raise ``ImportError`` if `paramiko <https://www.paramiko.org>`_ is not installed.
+
+        Returns :class:`RemoteSshBuildProducts`.
+        """
+        from paramiko import SSHClient
+
+        with SSHClient() as client:
+            client.load_system_host_keys()
+            client.connect(hostname, **connect_args)
+            with client.open_sftp() as sftp:
+                try:
+                    sftp.mkdir(root)
+                except IOError as e:
+                    pass # mkdir fails if directory exists. This is fine.
+
+                sftp.chdir(root)
+                for filename, content in self.files.items():
+                    filename = os.path.normpath(filename)
+                    # Just to make sure we don't accidentally overwrite anything outside of build root.
+                    assert not filename.startswith("..")
+
+                    dirname = os.path.dirname(filename)
+                    if dirname:
+                        try:
+                            sftp.mkdir(dirname, exist_ok=True)
+                        except IOError as e:
+                            pass
+
+                    mode = "wt" if isinstance(content, str) else "wb"
+                    with sftp.file(filename, mode) as f:
+                        f.write(content)
+
+            cmd = "cd {} && bash -l {}.sh".format(root, self.script)
+            stdin, stdout, stderr = client.exec_command(cmd)
+
+            buf = stdout.read(1024)
+            while buf:
+                print(buf.decode("utf-8"), end="")
+                buf = stdout.read(1024)
+
+            buf_err = stderr.read(1024)
+            while buf_err:
+                print(buf_err.decode("utf-8"), end="")
+                buf_err = stderr.read(1024)
 
     def execute(self):
         """
--- a/nmigen/build/run.py
+++ b/nmigen/build/run.py
@@ -149,8 +149,10 @@ class BuildPlan:
                 sftp.chdir(root)
                 for filename, content in self.files.items():
                     filename = pathlib.PurePosixPath(filename)
-                    # Just to make sure we don't accidentally overwrite anything outside of build root.
-                    assert not filename.parts[0] == ".."
+                    # Forbid parent directory components completely to avoid the possibility
+                    # of writing outside the build root. We can't use normpath to remove ".."
+                    # because we may be running on Windows, but the remote is a *nix machine.
+                    assert ".." not in filename.parts
 
                     mkdirs(filename)
 
--- a/nmigen/build/run.py
+++ b/nmigen/build/run.py
@@ -76,9 +76,10 @@ class BuildPlan:
             os.chdir(root)
 
             for filename, content in self.files.items():
-                filename = os.path.normpath(filename)
-                # Just to make sure we don't accidentally overwrite anything outside of build root.
-                assert not filename.startswith("..")
+                filename = pathlib.Path(filename)
+                # Forbid parent directory components completely to avoid the possibility
+                # of writing outside the build root.
+                assert ".." not in filename.parts
                 dirname = os.path.dirname(filename)
                 if dirname:
                     os.makedirs(dirname, exist_ok=True)
@@ -149,9 +150,6 @@ class BuildPlan:
                 sftp.chdir(root)
                 for filename, content in self.files.items():
                     filename = pathlib.PurePosixPath(filename)
-                    # Forbid parent directory components completely to avoid the possibility
-                    # of writing outside the build root. We can't use normpath to remove ".."
-                    # because we may be running on Windows, but the remote is a *nix machine.
                     assert ".." not in filename.parts
 
                     mkdirs(filename)
--- a/nmigen/build/run.py
+++ b/nmigen/build/run.py
@@ -7,9 +7,10 @@ import subprocess
 import tempfile
 import zipfile
 import hashlib
+import pathlib
 
 
-__all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts", "RemoteSshBuildProducts"]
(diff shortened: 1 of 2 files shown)


nmigen/build/run.py
```diff
@@ -9,7 +9,8 @@
 import hashlib
 
 
-__all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts"]
+__all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts", "RemoteSshBuildProducts"]
```
whitequark, line 12:
```suggestion
__all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts", "RemoteSSHBuildProducts"]
```

nmigen/build/run.py
```diff
@@ -99,6 +100,64 @@ def execute_local(self, root="build", *, run_script=True):
...
+                        try:
+                            sftp.mkdir(dirname, exist_ok=True)
+                        except IOError as e:
+                            pass
+
+                    mode = "wt" if isinstance(content, str) else "wb"
+                    with sftp.file(filename, mode) as f:
+                        f.write(content)
+
+            cmd = "cd {} && bash -l {}.sh".format(root, self.script)
```
whitequark, line 145: You shouldn't call bash directly. Not all build scripts even use bash.

nmigen/build/run.py
```diff
@@ -99,6 +100,64 @@ def execute_local(self, root="build", *, run_script=True):
...
+            stdin, stdout, stderr = client.exec_command(cmd)
+
+            # Show the output from the server while products are built.
```
whitequark, line 148: This seems like it could lead to deadlocks if something first writes a lot to stderr but not to stdout, but I'm not sure.

nmigen/build/run.py
```diff
@@ -162,3 +221,25 @@ def get(self, filename, mode="b"):
...
+        self.__root = root
+        self.__hostname = hostname
+        self.__connect_args = connect_args
+
+    def get(self, filename, mode="b"):
+        super().get(filename, mode)
+
+        from paramiko import SSHClient
+
+        with SSHClient() as client:
```
whitequark, line 237: Can you reuse the SSH client created earlier?

cr1901: Sure. But if I do that, do you object to a `run_script=False` input argument to `execute_remote_ssh`, so that a user is capable of skipping building remotely while _allowing_ to still program by grabbing remote build artifacts?

whitequark: Sure, that seems completely reasonable to have for parity with local builds.

nmigen/build/run.py
cr1901: You are correct, my mistake.

`-l` argument is a bash-ism (TIL!). `dash` also has `-l`, but other variants don't. So `sh -l` won't work in general to invoke a login shell. Since all we really want is to source `.profile`<sup>1</sup>, what about:

`if [ -f ~/.profile ]; then . ~/.profile; fi && cd {} && sh {}.sh`

1. Even if `sh` is a symlink to `bash`, `.bash_profile` will [not be read](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html) if invoked via the name `sh`.

nmigen/build/run.py
cr1901: `stdout` and `stderr` are subclassed from `paramiko`'s `BufferedFile`- it's called `ChannelFile`.

I can portably (Windows and *nix) `select` on them by accessing their [`channel`](https://github.com/paramiko/paramiko/blob/master/paramiko/channel.py#L1350). Based on the docstring that says `ChannelFile` wraps a `Channel`, I think accessing the `channel` field directly should be supported?

nmigen/build/run.py
whitequark: Based on the doc you linked, it seems that the most portable option would be to run `cd {root} && sh {self.script}`, and allow setting arbitrary environment variables, which may include `ENV`, `NMIGEN_ENV_<toolchain>`, or just `<tool_name>` directly depending on the use case.

nmigen/build/run.py
cr1901: One caveat to reusing the SSH client is that I lose the advantage of a context manager to automatically close the client. Per http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.close:

>Warning
Failure to [close] may, in some situations, cause your Python interpreter to hang at shutdown (often due to race conditions). It’s good practice to close your client objects anytime you’re done using them, instead of relying on garbage collection.

I need something that is guaranteed to call `close()`. AFAIU, adding ` __del__` to `RemoteSSHBuildProducts` won't help because it's not guaranteed `__del__` will run.

nmigen/build/run.py
whitequark: Ah. I think in this case it would be more appropriate to do something like:
```
transport = client.get_transport()
channel = transport.open_session()
channel.set_combine_stderr(True)
channel.exec_command("...")
```

Then you can poll just stdout.

nmigen/build/run.py
whitequark: Yes, that's true. I guess it's imperfect either way, so let's go with two separate client instances.

setup.py
```diff
@@ -46,6 +46,7 @@ def doc_version():
     extras_require={
         # this version requirement needs to be synchronized with the one in nmigen.back.verilog!
         "builtin-yosys": ["nmigen-yosys>=0.9.*"],
+        "remote-build": ["paramiko>=2.7.*"],
```
whitequark, line 49: This should be a semver constraint:
```suggestion
        "remote-build": ["paramiko~=2.7"],
```

nmigen/build/run.py
```diff
@@ -99,6 +100,64 @@ def execute_local(self, root="build", *, run_script=True):
         finally:
             os.chdir(cwd)
 
+    def execute_remote_ssh(self, root, hostname, *, connect_args = {}):
```
whitequark, line 103: I think a signature like the following:
```suggestion
    def execute_remote_ssh(self, *, connect_to={}, root):
```
would work better: it doesn't require placing the directory before the hostname, and it lets you pass only a single dict to `RemoteSSHBuildProducts`.

nmigen/build/run.py
```diff
@@ -99,6 +100,67 @@ def execute_local(self, root="build", *, run_script=True):
...
+
+            with client.open_sftp() as sftp:
+                try:
+                    sftp.mkdir(root)
+                except IOError as e:
+                    pass # mkdir fails if directory exists. This is fine.
+
+                sftp.chdir(root)
+                for filename, content in self.files.items():
+                    filename = os.path.normpath(filename)
```
whitequark, line 133: This function shouldn't use `os.path`, since it might be running on Windows, I think you should be using `pathlib.PurePosixPath`.

nmigen/build/run.py
```diff
@@ -162,3 +224,24 @@ def get(self, filename, mode="b"):
...
+        from paramiko import SSHClient
+
+        with SSHClient() as client:
+            client.load_system_host_keys()
+            client.connect(**self.__connect_to)
+
+            with client.open_sftp() as sftp:
+                sftp.chdir(self.__root)
+
+                with sftp.file(filename, "r" + mode) as f:
```
whitequark, line 246: The `mode` is [ignored](http://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.open) here, so this likely breaks any downstream code that expects text.

nmigen/build/run.py
```diff
@@ -99,6 +100,67 @@ def execute_local(self, root="build", *, run_script=True):
...
+
+                    dirname = os.path.dirname(filename)
+                    if dirname:
+                        try:
+                            sftp.mkdir(dirname, exist_ok=True)
+                        except IOError as e:
+                            pass
+
+                    mode = "wt" if isinstance(content, str) else "wb"
+                    with sftp.file(filename, mode) as f:
```
whitequark, line 145: Same problem as with the other `sftp.file` call.

nmigen/build/run.py
cr1901: What do you mean by "downstream" code? Should I be converting to *nix line endings when writing the file (i.e. iterate over lines)?

whitequark: Well, if something calls `extract` and passes `"t"` as a mode it expects a `str` back, not `bytes`, and will break since this code returns `bytes`.

nmigen/build/run.py
```diff
@@ -99,6 +101,83 @@ def execute_local(self, root="build", *, run_script=True):
...
+                        if e.errno:
+                            raise e
+
+                def mkdirs(path):
+                    # Iteratively create parent directories of a file by iterating over all
+                    # parents except for the root ("."). Slicing the parents results in
+                    # TypeError, so skip over the root ("."); this also handles files
+                    # already in the root directory.
+                    for parent in reversed(path.parents):
+                        if parent == pathlib.PurePosixPath("."):
```
whitequark, line 142: Since `path.parents` is a lexical operation, this code will have issues if someone passes a path like `foo/../bar` (or, worse, `foo/../../bar`; instead of attempting to fix every possible case, I suggest simplifying it to:
```
try:
    sftp.chdir(root)
except IOError as e:
    if e.errno == errno.ENOENT:
        sftp.mkdir(root)
    else:
        raise e
```

whitequark: Oh sorry, I misread--that's actually exactly what you're doing here.

nmigen/build/run.py
```diff
@@ -149,8 +149,10 @@ def mkdirs(path):
                 sftp.chdir(root)
                 for filename, content in self.files.items():
                     filename = pathlib.PurePosixPath(filename)
-                    # Just to make sure we don't accidentally overwrite anything outside of build root.
-                    assert not filename.parts[0] == ".."
+                    # Forbid parent directory components completely to avoid the possibility
+                    # of writing outside the build root. We can't use normpath to remove ".."
+                    # because we may be running on Windows, but the remote is a *nix machine.
+                    assert ".." not in filename.parts
```
whitequark, line 155: Instead of explaining why we don't use normpath here, let's just drop it in local builds as well. This way we don't have to explain anything, and the next person who copies the code for local builds doesn't introduce a similar bug, either.

whitequark (approved):

nmigen/build/run.py
cr1901: Fixed. `os.path` functions accepting `Pathlib` objects works down to and include Python 3.6, which is our cutoff. So might as well use it.

whitequark:
Thanks!

PR #463, opened 2020-08-03
Add initial support for Symbiflow toolchain for Xilinx 7-series

mglb:
The PR adds initial support for Symbiflow toolchain for XC7 devices. To use Symbiflow instead of Vivado, a platform should inherit `Xilinx7SeriesSymbiflowPlatform` instead of `Xilinx7SeriesPlatform`.
The example used below is a copy of `arty_a7.py` from nmigen-boards, where in addition to changing the base class, the device (part) name has also been changed (original part is not present in prjxray/symbiflow database).

Test:
```
wget 'https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh' -O miniconda.sh
bash miniconda.sh -b -p ./miniconda
source ./miniconda/etc/profile.d/conda.sh
hash -r

conda config --set always_yes yes --set changeps1 no
conda config --add channels conda-forge
conda config --add channels symbiflow
conda update -q conda
conda activate
conda install -c symbiflow yosys yosys-plugins vtr prjxray-tools
conda install lxml simplejson intervaltree python-constraint git pip
pip install git+https://github.com/symbiflow/fasm

git clone https://github.com/SymbiFlow/prjxray.git
git clone https://github.com/SymbiFlow/prjxray-db.git

cd ./prjxray
python setup.py install
cd ..

wget 'https://storage.googleapis.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/presubmit/install/385/20200702-165441/symbiflow-arch-defs-install-3f58546e.tar.xz'
tar -xf symbiflow-arch-defs-install-3f58546e.tar.xz

export PATH=$(realpath ./install/bin/):$PATH
export XRAY_DATABASE_DIR=$(realpath ./prjxray-db)
export XRAY_FASM2FRAMES=$(realpath ./prjxray/utils/fasm2frames.py)
export XRAY_TOOLS_DIR=$(realpath ./miniconda/bin)

# Get PR code
git clone https://github.com/antmicro/nmigen.git -b mglb/new-initial-symbiflow-support
cd nmigen
python setup.py develop
cd ..

# The example uses nmigen-boards
pip install git+https://github.com/nmigen/nmigen-boards.git

# Download and run example target for Arty A7
wget 'https://raw.githubusercontent.com/antmicro/nmigen/mglb/new-initial-symbiflow-support-example/examples/symbiflow/arty_a7.py'
python arty_a7.py
```


Ravenslofty:
Considering `vendor.lattice_{ice40,ecp5}` can get away with using the same platform for different toolchains, may I ask why you chose to create a whole new platform here?

mglb:
I've implemented your suggestions. Now you just have to pass `toolchain="Symbiflow"` to platform's initializer (and use slightly different device name and speed in Arty example).

whitequark:
Thanks, this is much better.

> and use slightly different device name and speed in Arty example

Would it be possible to use the same device name for the vendor and the FOSS toolchains? For other platforms that support both, this is handled transparently: either by adjusting the FOSS toolchain to accept the same device names, or by mapping them within nMigen. E.g. you can see this done in the ECP5 platform file.

whitequark (approved):
LGTM

nmigen/vendor/xilinx_7series.py
```diff
@@ -165,17 +312,27 @@ def create_missing_domain(self, name):
...
+                # breaks Vivado 2017.4.
+                m.submodules += Instance("BUFGCTRL",
+                    i_I0=clk_i,   i_S0=C(1, 1), i_CE0=ready,   i_IGNORE0=C(0, 1),
+                    i_I1=C(1, 1), i_S1=C(0, 1), i_CE1=C(0, 1), i_IGNORE1=C(1, 1),
+                    o_O=ClockSignal("sync")
+                )
+            elif self.toolchain == "Symbiflow":
+                cd_sync = ClockDomain("sync", reset_less=self.default_rst is None)
+                m.domains += cd_sync
+                m.submodules += Instance("BUFG", i_I=clk_i, o_O=cd_sync.clk)
```
whitequark, line 331: Does Symbiflow actually *need* this `BUFG`, or can it insert one automatically?

mglb: Simple example (blinky) works without this bufg. However, we had some problems with Litex+VexriscV+Linux without it.

whitequark:
Thanks!

whitequark:
Have you tested the PR? It currently completely breaks 7-series because `if toolchain == "Symbiflow":` should be `if self.toolchain == "Symbiflow":`.

whitequark:
Fixed in abaa9091f45d1c5801ec7162322bbc04706e3b43.

PR #466, opened 2020-08-04
vendor.lattice_machxo_2_3l: add SRAM svf generation

rroohhh:
Might be nice to rename the FLASH svf to `{{name}}_flash.svf`, but this would be a breaking change I guess.


whitequark:
> but this would be a breaking change I guess.

Let's produce all three files at first, then fix the nmigen-boards repo to use `_sram.svf`, then remove the unsuffixed svf file before 0.4.

rroohhh:
Sounds good

whitequark:
> Sounds good, although I don't see anything that actually uses the svf file in nmigen-boards, did I miss anything?

I might be misremembering.

PR #468, opened 2020-08-08
Generic Vivado FPGA programming script.

Fatsie:
Will likely also work for xilinx_ultrascale.py but I can't test that. Tested it here on Arty S7-50.

### DIFF
--- a/nmigen/vendor/xilinx_7series.py
+++ b/nmigen/vendor/xilinx_7series.py
@@ -1,3 +1,5 @@
+import subprocess
+import textwrap
 from abc import abstractproperty
 
 from ..hdl import *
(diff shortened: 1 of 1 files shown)


whitequark:
I'm not sure if this is the right approach, for two reasons.

First, it won't work with non-Vivado toolchains at all (see #463).

Second, while the *build process* is inherently tied to Vivado (if Vivado is chosen as the toolchain), there's nothing that inherently ties the *programming process* to Vivado, whether with Digilent developer boards, or with others. So it's not obvious to me that this is a sensible default.

Fatsie:
> First, it won't work with non-Vivado toolchains at all (see #463).

Sure, but I added it to a platform that only has Vivado as toolchain. I agree that this would need to be revisited if more than one toolchain would be supported for Xilinx 7 Series.

> Second, while the _build process_ is inherently tied to Vivado (if Vivado is chosen as the toolchain), there's nothing that inherently ties the _programming process_ to Vivado, whether with Digilent developer boards, or with others. So it's not obvious to me that this is a sensible default.

Are there boards where build process is Vivado but can't be programmed with Vivado ?
Personally I prefer taking Vivado by default if possible as people will need to have that installed anyway and not necesserily xc3sprog or openFPGALoader.

whitequark:
> I agree that this would need to be revisited if more than one toolchain would be supported for Xilinx 7 Series.

Well, I'd rather not accept a PR that I know would need fundamental rework (that will likely break backward compatibility) almost immediately.

> Are there boards where build process is Vivado but can't be programmed with Vivado ?

I believe any board where the FPGA is programmed by an FX2 would qualify, since effectively all of them use custom one-off programmers (integrated on board, but still).

Fatsie:
From the other side I am not convinced that the programming is totally disconnected from the building. When changing toolchain one may also want to adapt the (default) programmer.

Fatsie:
Anyway, closing as this PR is not considered fit for merging.

whitequark:
> When changing toolchain one may also want to adapt the (default) programmer.

Yes, I'm open to consider that option, though there may be unexpected consequences.

PR #469, opened 2020-08-08
back+cli: Add FIRRTL, json and dot backends. Add ability to use them with cli.main

pbsds:
All of these backends enable you to inspect your design more easily, by using netlistsvg, diagrammer and graphviz.

I did try to expand `ExamplesTestCase` for some coverage on these backends, but it seems like `write_firrtl` in yosys is unable to create the memory defined in `examples/mem.py`. Therefore I left the tests out of this PR for now.

### DIFF
--- a/nmigen/back/dot.py
+++ b/nmigen/back/dot.py
@@ -1,0 +1,34 @@
+from .._toolchain.yosys import *
+from . import rtlil
+from .verilog import _convert_rtlil_text
+import tempfile, os
+
+
+__all__ = ["YosysError", "convert", "convert_fragment"]
+
+
+def _convert_rtlil_text_dot(rtlil_text, *, strip_internal_attrs=False, show_opts=("-stretch", "-colors 42")):
+    # TODO: add option in yosys to output graphviz to stdout
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        prefix = os.path.join(tmpdirname, "tmp")
+        viewer = "cat" if os.name != "nt" else "type"
+
+        write_script = []
+        # TODO: can these procs be improved?
+        write_script.append("proc_mux")
+        write_script.append("proc_clean")
+        write_script.append("show -format dot -prefix {} -viewer {} {}"
+            .format(prefix, viewer, " ".join(show_opts)))
+
+        return _convert_rtlil_text(rtlil_text,
+            strip_internal_attrs=strip_internal_attrs, write_script=write_script)
+
+
+def convert_fragment(*args, strip_internal_attrs=False, **kwargs):
+    rtlil_text, name_map = rtlil.convert_fragment(*args, **kwargs)
+    return _convert_rtlil_text_dot(rtlil_text, strip_internal_attrs=strip_internal_attrs), name_map
+
+
+def convert(*args, strip_internal_attrs=False, **kwargs):
+    rtlil_text = rtlil.convert(*args, **kwargs)
+    return _convert_rtlil_text_dot(rtlil_text, strip_internal_attrs=strip_internal_attrs)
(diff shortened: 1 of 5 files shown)


whitequark:
Are you adding these backends with some specific application in mind? They significantly expand our public API (with all the resulting commitment to support), and I don't think this is appropriate at this time; I'd like to help you find a different approach to achieve your goals.

pbsds:
Fair criticism. I made this PR to avoid littering yosys scripts everywhere. My end goal is a live updating graph of the module.

Would a PR which makes `cli.main` extendable through plugins be of interest instead?

whitequark:
> I made this PR to avoid littering yosys scripts everywhere. My end goal is a live updating graph of the module.

Sure, I'd like to help you achieve that in a way that would work well with nMigen's approach to API stability. Can you tell me more about the way you're using nMigen and Yosys in your project?

> Would a PR which makes `cli.main` extendable through plugins be of interest instead?

The current design of `cli.main` is, frankly, largely a mistake (I wrote it so I can say that...) so I'm not that fond of any extensions that further tie us towards it.

pbsds:
I do like the idea of supplying a `main` function to use with nmigen. I've been drafting a new `nMain` as i've been exploring the different guides for nmigen, which i intend to add to nmigen at some point.

<details>
<summary>my current idea draft</summary>

```python

class MyModule(Elaboratable):
    def __init__(self, param: int):
        self.param = param
        self.out = Signal()
    ...

# the largest issue is the lack of a consistent cli interface to expect from a given file
if __name__ == "__main__":
    top = MyModule(param=4) # warning: MustUse

    if sys.argv[1] = "build":
        if sys.argv[2] = "icebreaker":
            plat = ICEBreakerPlatform()
            plat.add_resources(...)

        elif sys.argv[2] = "fomu"
            plat = FomuHackerPlatform()
            plat.add_resources(...)

        # `do_program` is usually hard-coded.
        # The folder "build" should also be configurable
        plat.build(top, do_program=True)

    elif sys.argv[1] == "test": # "simulate" already taken in main
        # logic for multiple simulations to select from is manually made
        sim = Simulator(top)

        def my_proc():
            yield ...
        sim.add_sync_process(my_proc)

        with sim.write_vcd("uart.vcd", "uart.gtkw"): # hardcoded filenames, never printed to user
            sim.run()

    else:
        main(top, ports=[ top.out ]) # '-h' doesn't mention the two modes defined above




# === can become: ===




class MyModule(Elaboratable):
    def __init__(self, param):
        self.param = param
        self.out = Signal()
    ...

if __name__ == "__main__":
    platforms, simulations = [], []

    # supporting multiple platforms encourages designing the
    # top-level module in a platform agnostic way

    @platforms.append
    def icebreaker(): # function.__name__ is used by the cli interface in nMain
        plat = ICEBreakerPlatform()
        plat.add_resources(...)
        return plat

    @platforms.append
    def fomu():
        plat = FomuHackerPlatform()
        plat.add_resources(...)
        return plat

    @simulations.append
    def sim_the_foo(sim): # sim = Simulator(design)
        sim.add_clock(1e-6)

        @sim.add_sync_process
        def my_proc():
            yield ...
        # sim could then be automatically run, with configurable filenames

    nMain(
        name = "blinker", # current default is "top", perhaps use design().__class__.__name__ ?
        design = lambda: MyModule(param=4), # avoid MustUse warnings
        ports = lambda x: [ x.out ], # optional, platform.request is a thing
        platforms = platforms, # optional
        simulations = simulations, # optional
    )


```

</details>

When i feel it's ready i could make an issue/PR about it. If you have some ideas i can take into account.

> Can you tell me more about the way you're using nMigen and Yosys in your project?

I'm very much a beginner with hardware, and therefore I've been having fun with all the points of friction with learning vivado, yosys, chisel3 and nmigen. Therefore i'm currently looking into how to [structure a nmigen project](https://github.com/pbsds/nmigen-learning).
I've also been playing around with [reducing the visual noise of the nmigen dsl](https://github.com/pbsds/nmigen_dg), although the ends does not really justify the means in that one.

Issue #464, opened 2020-08-03
[RFC] Shape Conversion Operators Round 2


awygle:
This RFC is a continuation of #381 .

Currently, in order to change the shape of a signal, the following workarounds are needed: 
  * for truncation, use `v[:new_width]` 
  * for extension, use a no-op bitwise operation, like `v | C(0, new_width)`.

This code is confusing when it is encountered among a bunch of arithmetic operations, and it's more of a side effect than an explicit request to change shape. This makes it fragile, because neither of these operations is guaranteed to return a value with width `new_width`.

There are three proposals on the table currently for improving this situation.

**Proposal 1**: Add three new operators, each of which takes a `width`

 * `v.zero_extend(new_width)` that works only on unsigned values, requires `new_width >= len(v)`, and returns unsigned values,
 * `v.sign_extend(new_width)` that works only on signed values, requires `new_width >= len(v)`, and returns signed values,
 * `v.truncate(new_width)` that works on values of either signedness, requires `new_width <= len(v)`, returns a value of the same signedness as the input, and checks (in simulation and formal) that the truncated bits were all identical to 0 (for unsigned values) or the new MSB (for signed values).

**Proposal 2**: Add a single general-purpose shape conversion operator, `v.as(new_shape)`, for truncation, zero/sign extension, and type conversion. The new operator would extend v if `new_shape.width > len(v)`, truncate v if `new_shape.width < len(v)`, and return a value with shape `new_shape`. The semantics of this operator would be to preserve the mathematical value (modulo output size), as described in [this comment](https://github.com/nmigen/nmigen/issues/381#issuecomment-625669445).

**Proposal 3**: Add two new operators, each of which takes a `Shape`

 * `v.extend(new_shape)` that works on signed or unsigned values and requires `new_shape.width >= len(v)`
 * `v.truncate(new_shape)` that works on values of either signedness and requires `new_shape.width <= len(v)`

Both of these operators would also preserve the mathematical value modulo output size and return a value with shape `new_shape`.

awygle:
Switching hats, I'm personally strongly in favor of Proposal 1. My mental model is that `Value`s are just wires, and it's the operations which are signed or unsigned (think of the difference between `atomic_load`/`atomic_store` and `volatile`).

whitequark:
cc @programmerjake -- it was your comment which had the insight about preserving mathematical value that led to a second round of this RFC, so I'd certainly like to know your opinion here as well

[...]

PR #481, opened 2020-08-19
hdl.ast.Value: Add .to_signed() method

pbsds:
We currently have a `.as_unsigned()`, but if the MSB is set
then its numerical value will be changed by it.

Consider this case for the unsigned variables a and b of same width, where a < b:

```python
a - b               # will underflow
a.as_unsigned() - b # value of a is changed if a[-1] == 1
a.to_unsigned() - b # correct
```


whitequark:
Sorry it took so long to get back to you!

> ```python
> a - b               # will underflow
> a.as_unsigned() - b # value of a is changed if a[-1] == 1
> a.to_unsigned() - b # correct
> ```

I assume you mean the following:
```python
a.to_signed() - b # correct
```

We discussed your proposal on the last meeting. I agree that it addresses an issue with nMigen language as it exists. However, I have two concerns:
  * It would be very easy to confuse `as_signed` (whose function is already not entirely obvious) with `to_signed` (which differs in a subtle way).
  * The way `to_signed` works is a subset of how a shape conversion operator would work, i.e. Proposal 1 from #464 would make it unnecessary.

Because of these concerns I think the issue you are highlighting would be best addressed as a part of #464.

PR #486, opened 2020-08-24
Fix ArrayProxy shape for signed numbers

### LINKED ISSUE #476: ArrayProxy shape does not equal element shape for a homogeneous array of signed values
https://github.com/nmigen/nmigen/blob/e46118dac0df315694b0fc6b9367d285a8fc12dd/nmigen/hdl/ast.py#L1202
Why is the signedness of the element added to its width? Maybe the intention was to add the "sign bit", but my understanding is that this is purely a marker, and the width is unchanged.

pepijndevos:
fixes #476

This time for real hopefully. Any edge cases I forgot?

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -1197,11 +1197,20 @@ class ArrayProxy(Value):
         return (Value.cast(elem) for elem in self.elems)
 
     def shape(self):
-        width, signed = 0, False
-        for elem_width, elem_signed in (elem.shape() for elem in self._iter_as_values()):
-            width  = max(width, elem_width + elem_signed)
-            signed = max(signed, elem_signed)
-        return Shape(width, signed)
+        shapes = [elem.shape() for elem in self._iter_as_values()]
+        widths, signs = zip(*shapes)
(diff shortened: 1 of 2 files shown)


whitequark:
I had a hard time convincing myself about the new logic in this PR, so eventually I rewrote it in 00026c6e4a0d3f9c2eff054b1bd4d8d18342a2ed. I did use your tests though; thanks!

pepijndevos:
Clarity is in the eye of the beholder apparently haha
My brain seems to be wired in a more functional programming way than is usually considered Pythonic I guess.
I tend to favour complex list comprehensions over for loops that touch 4 variables.
Glad it's fixed either way :)

whitequark:
> Clarity is in the eye of the beholder apparently haha
> My brain seems to be wired in a more functional programming way than is usually considered Pythonic I guess.

It's mostly the use of arithmetic operations on bools (which most functional languages I know don't allow). The original code that I wrote did it, and that was pretty confusing (I think you actually got confused at first). I had to read it a few times to convince myself that it's correct, so I thought I might as well rewrite it.

PR #490, opened 2020-08-26
vendor.xilinx_{7series,ultrascale}: set BUFG* SIM_DEVICE as appropriate

### LINKED ISSUE #438: wrong type of buffer primitive used in series 7
when I try to elaborate a very simple blinking-led example I wrote, I get this from vivado (2019.2):
```
WARNING: [Netlist 29-345] The value of SIM_DEVICE on instance 'cd_sync/U$$1' of type 'BUFGCTRL' is 'ULTRASCALE'; it is being changed to match the current FPGA architecture, '7SERIES'. For functional simulation to match hardware behavior, the value of SIM_DEVICE should be changed in the source netlist. 
```
it looks to me like nMigen instantiates the wrong primitive for some buffer

whitequark:
Fixes #438 (again).


whitequark:
This PR breaks 2017.4. There is no way to add a compatibility shim, so 2017.4 is not supported anymore.

PR #499, opened 2020-09-15
hdl.mem: document ReadPort and WritePort.

### LINKED ISSUE #496: no doc string for `en` in `ReadPort`
typing `help(Memory(width=1,depth=1).read_port())` into the interpreter yields no information about the `en` signal of the read port.

jfng:
Fixes #496.

### DIFF
--- a/nmigen/hdl/mem.py
+++ b/nmigen/hdl/mem.py
@@ -81,9 +81,35 @@ class Memory:
                             .format(addr, e)) from None
 
     def read_port(self, *, src_loc_at=0, **kwargs):
+        """Get a read port.
+
+        See :class:`ReadPort` for details.
+
+        Arguments
+        ---------
+        domain : str
+        transparent : bool
+
+        Returns
+        -------
+        An instance of :class:`ReadPort` associated with this memory.
+        """
         return ReadPort(self, src_loc_at=1 + src_loc_at, **kwargs)
 
     def write_port(self, *, src_loc_at=0, **kwargs):
+        """Get a write port.
+
+        See :class:`WritePort` for details.
+
+        Arguments
+        ---------
+        domain : str
+        granularity : int
+
+        Returns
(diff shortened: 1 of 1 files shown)


nmigen/hdl/mem.py
```diff
@@ -92,6 +118,37 @@ def __getitem__(self, index):
...
+    memory : :class:`Memory`
+    domain : str
+    transparent : bool
+    addr : Signal(range(memory.depth)), in
+        Read address.
+    data : Signal(memory.width), out
+        Read data.
+    en : Signal or Const, in
+        Read enable. If asserted, ``data`` is updated with the word stored at ``addr``. If the
+        port is transparent, it is hardwired to 1 and cannot be driven by user logic.
```
whitequark, line 146: It's worth mentioning that this hardwiring behavior is not what we want to do, but in fact the consequence of a Yosys design issue.

nmigen/hdl/mem.py
```diff
@@ -92,6 +118,37 @@ def __getitem__(self, index):
...
 class ReadPort(Elaboratable):
+    """A memory read port.
+
+    Parameters
+    ----------
+    memory : :class:`Memory`
+        Memory associated with the port.
+    domain : str
+        Clock domain. Defaults to ``"sync"``. If set to ``"comb"``, the port is asynchronous.
+        Otherwise, reads have a latency of 1 clock cycle.
```
whitequark, line 129: I would rephrase this along the lines of "the read data becomes available on the next clock cycle", since otherwise there's a bit of ambiguity with pipelined reads where that'd be an *extra* clock cycle.

whitequark (changes requested):
Good work! There's a few clarifications I would add.

jfng:
Updated to address previous comments.

whitequark (approved):

PR #504, opened 2020-10-14
Add quicklogic platform

kowalewskijan:
This PR introduces a support for QuickLogic EOS-S3 platform.


nmigen/vendor/quicklogic.py
```diff
@@ -0,0 +1,214 @@
...
+            -f {{name}}.fasm
+            -d {{platform.device}}
+            -P {{platform.part}}
+            -b {{name}}.bit
+        """
+    ]
+
+    # Common logic
+
+    def __init__(self, *, toolchain="Quicklogic"):
```
whitequark, line 123: I don't think this parameter (and all of the branching below) is necessary, since there's only one Quicklogic toolchain, the FOSS one.

nmigen/vendor/quicklogic.py
```diff
@@ -0,0 +1,214 @@
...
+
+        if pin.xdr == 0:
+            if "i" in pin.dir:
+                i = pin_i
+            if "o" in pin.dir:
+                o = pin_o
+            if pin.dir in ("oe", "io"):
+                t = ~pin.oe
+        else:
+            assert False
```
whitequark, line 212: Does Quicklogic have registered and DDR inputs? If yes, they should be supported. If not, they should not be advertised above.

whitequark (commented):
Looks good other than the comments, thanks for this work!

nmigen/vendor/quicklogic.py
kowalewskijan: Makes sense. I removed `toolchain` variable from platform definition.

nmigen/vendor/quicklogic.py
kowalewskijan: Neither SDR nor DDR are supported. I set `valid_xdrs=0` and I [simplified](https://github.com/nmigen/nmigen/pull/504/files#diff-5b82fd62f94e74b0762f9965e2b93236426be4957ebbaa42325de8c0501ec880R172-R175) handling them a bit since there is only one option.

nmigen/vendor/quicklogic.py
```diff
@@ -0,0 +1,195 @@
...
+            -b {{name}}.bit
+        """
+    ]
+
+    # Common logic
+
+    def __init__(self, *):
+        super().__init__()
+
+    @property
```
whitequark, line 124: None of the following four functions are necessary--just make them normal properties. They only need to be functions with multiple supported toolchains.

nmigen/vendor/quicklogic.py
whitequark: Okay, but now this code doesn't make any sense. As far as I can tell, you can just remove `get_input` and `get_output` since they are not doing anything useful.

kowalewskijan: All right - removed them

whitequark: You should then also remove `_get_xdr_buffer`, since it is not used anymore. Please put some effort into addressing review comments, because it is nearing the point where it would be easier for me to rewrite your code than to review it.

whitequark (approved):
LGTM, thanks!

whitequark:
Note: despite the earlier comment, EOS S3 *does* support SDR registers (see [here](https://www.quicklogic.com/wp-content/uploads/2020/06/QL-S3-Technical-Reference-Manual.pdf#page=273)), but the FOSS toolchain does not currently support them (see [here](https://github.com/QuickLogic-Corp/quicklogic-fpga-toolchain#hardware-features-that-are-not-supported-in-this-release)). Once it does we should add SDR support.

PR #506, opened 2020-10-16
hdl.ir: Update error message for Instance arguments

Xiretza:
48d4ee4 added the option to specify attributes using Instance arguments, but the error message wasn't updated accordingly.


whitequark (approved):

whitequark:
You need to update the tests too.

whitequark:
Thanks!

PR #509, opened 2020-10-19
setup.py: Exclude "tests" package

Xiretza:
67b957d moved the tests from `nmigen/test/` to `tests/`, and removed the `exclude=` parameter from `find_packages()` in setup.py. However, even if the new location is not inside the module tree, it is still found by `find_packages()`, resulting in a stray "tests" module on the system.


whitequark:
Thanks!

PR #511, opened 2020-10-22
improve spec directory name of FHDLTestCase

anuejn:
This PR improves the internal version of FHDLTestCase by trying to get the last stack frame that is still inside the nmigen codebase when creating names for the spec file. This helps to make the FIFO test cases produce all distinct spec directories and makes the nmigen test suite thread-safe (can be run successfully with pytest-xdist)

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -28,7 +28,12 @@ class FHDLTestCase(unittest.TestCase):
         self.assertEqual(prepare_repr(repr(obj)), prepare_repr(repr_str))
 
     def assertFormal(self, spec, mode="bmc", depth=1):
-        caller, *_ = traceback.extract_stack(limit=2)
+        stack = traceback.extract_stack()
+        for frame in reversed(stack):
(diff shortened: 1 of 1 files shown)


whitequark:
This is kind of cursed, but... the whole point of making `FHDLTestCase` internal is that we can do whatever cursed stuff we want. So I'm fine with it.

whitequark:
> trying to get the last stack frame that is still inside the nmigen codebase when creating names for the spec file

Wait, how does this actually work?

whitequark:
Ah okay, I see how it works. I think you shouldn't match `nmigen` in the path because there's no guarantee or requirement for the git checkout name; in particular I think CI often check out the project as something like `~/work` (Appveyor does it iirc?). Maybe match against dirname of `utils.py`?

anuejn:
Ah you are right, I missed that. Sounds good; will do :)

whitequark:
Thanks!

PR #512, opened 2020-10-22
Fix {r,w}_level in AsyncFIFOBuffered

### LINKED ISSUE #485: {r,w}_level is broken on AsyncFIFO{Buffered,}
It seems like the r_level / w_level on AsyncFIFO / AsyncFIFOBufered are broken in multiple ways: 

* If the FIFO is full the level indicators become 0
* The semantics of w_level of AsyncFIFOBuffered are quite unintuitive since it cannot reach depth and is quite different from r_level / w_level of AsyncFIFO.

@awygle started to work on a fix for this and I picked up his work and added a simulation-based test case (https://github.com/nmigen/nmigen/compare/master...anuejn:fifo_bugfix). 
However, it is unclear to me, how one would implement the same behavior of w_level for AsyncFIFOBuffered as for AsyncFIFO without violating CDC rules. Maybe anyone has an idea here? Am I wrong here?

anuejn:
This Fixes #485 as per discussion in https://freenode.irclog.whitequark.org/glasgow/2020-10-22#1603367304-1603369561

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -506,10 +506,15 @@ class AsyncFIFOBuffered(Elaboratable, FIFOInterface):
             fifo.w_data.eq(self.w_data),
             self.w_rdy.eq(fifo.w_rdy),
             fifo.w_en.eq(self.w_en),
-            self.w_level.eq(fifo.w_level),
         ]
 
-        m.d[self._r_domain] += self.r_level.eq(fifo.r_level + self.r_rdy - self.r_en)
+        r_consume_buffered = Signal()
+        m.d.comb += r_consume_buffered.eq(self.r_rdy - self.r_en)
+        m.d[self._r_domain] += self.r_level.eq(fifo.r_level + r_consume_buffered)
+
+        w_consume_buffered = Signal()
+        m.submodules.consume_buffered_cdc = AsyncFFSynchronizer(r_consume_buffered, w_consume_buffered, o_domain=self._w_domain)
+        m.d.comb += self.w_level.eq(fifo.w_level + w_consume_buffered)
 
         with m.If(self.r_en | ~self.r_rdy):
             m.d[self._r_domain] += [
--- a/tests/test_lib_fifo.py
+++ b/tests/test_lib_fifo.py
@@ -282,6 +282,7 @@ class FIFOFormalCase(FHDLTestCase):
         self.check_async_fifo(AsyncFIFOBuffered(width=8, depth=4))
 
 
+# we need this testcase because we cant do model equivalence checking on the async fifos (at the moment)
 class AsyncFIFOSimCase(FHDLTestCase):
     def test_async_fifo_r_level_latency(self):
         fifo = AsyncFIFO(width=32, depth=10, r_domain="sync", w_domain="sync")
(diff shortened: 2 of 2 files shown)


nmigen/lib/fifo.py
```diff
@@ -369,7 +369,7 @@ def elaborate(self, platform):
         produce_dec = m.submodules.produce_dec = \
             GrayDecoder(self._ctr_bits)
         m.d.comb += produce_dec.i.eq(produce_r_gry),
-        m.d[self._r_domain] += produce_r_bin.eq(produce_dec.o)
+        m.d.comb += produce_r_bin.eq(produce_dec.o)
```
whitequark, line 372: Wait, why does this PR change level-unrelated AsyncFIFO logic at all?

anuejn: Hmm... you are right, that seems wrong. 
I don't remember to write that, so we should probably ask @awygle or undo that change

anuejn:
This should be rebased onto master now (which includes the other PR)

nmigen/lib/fifo.py
```diff
@@ -380,8 +380,8 @@ def elaborate(self, platform):
             r_empty.eq(consume_r_gry == produce_r_gry),
         ]
 
-        m.d[self._w_domain] += self.w_level.eq((produce_w_bin - consume_w_bin)[:self._ctr_bits-1])
-        m.d.comb += self.r_level.eq((produce_r_bin - consume_r_bin)[:self._ctr_bits-1])
+        m.d[self._w_domain] += self.w_level.eq((produce_w_bin - consume_w_bin))
+        m.d.comb += self.r_level.eq((produce_r_bin - consume_r_bin))
```
whitequark, line 384: Why did you change this? I thought this PR only affects AsyncFIFOBuffered.

nmigen/lib/fifo.py
```diff
@@ -506,10 +506,15 @@ def elaborate(self, platform):
             fifo.w_data.eq(self.w_data),
             self.w_rdy.eq(fifo.w_rdy),
             fifo.w_en.eq(self.w_en),
-            self.w_level.eq(fifo.w_level),
         ]
 
-        m.d[self._r_domain] += self.r_level.eq(fifo.r_level + self.r_rdy - self.r_en)
+        buffer_used = Signal()
```
whitequark, line 511: I think this code would be easier to understand if the signals subject to CDC used the same naming conventions as the ones in `AsyncFIFO`; something like `r_consume_buffered` and `w_consume_buffered`.

nmigen/lib/fifo.py
```diff
@@ -506,10 +506,15 @@ def elaborate(self, platform):
...
+        m.d.comb += buffer_used.eq(self.r_rdy - self.r_en)
+        m.d[self._r_domain] += self.r_level.eq(fifo.r_level + buffer_used)
+
+        buffer_used_in_wr_domain = Signal()
+        m.submodules.rst_cdc = AsyncFFSynchronizer(buffer_used, buffer_used_in_wr_domain, o_domain=self._w_domain)
```
whitequark, line 516: The `rst_cdc` name is a copy/paste leftover. I'd use `consume_buffered_cdc`.

whitequark (changes requested):
Looks good overall, just a few minor issues.

nmigen/lib/fifo.py
anuejn: If I remember correctly this is needed so that the counter doesnt overflow (as tested by the testsuite)

anuejn: should I split that change into its own pr then?

whitequark: I'm still not entirely certain why it's done; is this basically fixing a second bug unrelated to the `w_level` issue?

anuejn: without that change https://github.com/nmigen/nmigen/pull/512/files#diff-5a301e8309610ecfa06031693c54b1c02c7d05a74fa97224ecffd0470d950183R339 would fail as the reported level would be 0

whitequark: Yes, please split it out.

whitequark (approved):

whitequark:
Thanks!

PR #513, opened 2020-10-23
Fix latency issues with AsyncFIFO

anuejn:
This is the split out commit with testcase as requested for #512 in IRC


anuejn:
Whoops there is an unrelated commit that slipped in :(

whitequark:
Thanks @awygle for fixing this and @anuejn for splitting the PR!

Issue #510, opened 2020-10-20
[RFC] Allow DiffPairs and Pins to be inverted individually


anuejn:
Currently DiffPairs and Pins cann be either inverted completely or not at all.
This is a problem in corner cases where the hardware inverts some parts of a logical Bundle of ios (for example to ease routing in case of differential pairs; see #386). 

I herby propose to extend the Platform DSL in the following way:
* Add a `invert=` kwarg to the constructor of `DiffPairs`
* Allow Tuples of the width of the pin list to be passed to the `invert=` kwarg of both `DiffPairs` and `Pins`

whitequark:
Right, this is the obvious solution. The main problem I have is that when you have a lot of diffpairs, it can be hard to correlate the booleans in the `invert=` tuple with the actual diffpairs.

whitequark:
As an alternative I was wondering if something like `DiffPairs("A1 !A2", "B1 !B2")` would work. The `!` sigils would then have to match in both the P and the N pin lists.

adamgreig:
I like putting it in the pin list string(s), it's compact but I think also quite clear. Maybe "~A1" would be a better match to the nmigen negation syntax?

whitequark:
Sure, `~` is fine by me.

anuejn:
I also really like that idea :)

whitequark:
Sounds great. I'll bring this up on the next meeting but I think you can start implementing it as it seems unlikely anyone's going to be opposed.

anuejn:
sounds good :)

anuejn:
I think we should support similar syntax for the use with connectors (which can be routed through e.g. inverted level shifters or have crossed differential pairs). So I think it would be nice to have syntax elements in connectors that can say "invert this signal if it is a diff pair" or "invert this signal always".

whitequark:
I would like to avoid introducing that functionality (and the associated syntactic burden) until someone actually needs it.

anuejn:
/me needs the latter (inverted differential pairs in connectors)

whitequark:
Okay. I still think it should be done separately because while this one is fairly straightforward (it's just a way to write a tuple that's less error prone than usual), what you request with connectors requires completely rethinking the way we do connectors.

anuejn:
Okay I see that.
So I will implement this RFC as-is and go with out of tree hacks for now.
If I really need this feature to become upstream I will write another RFC :)

whitequark:
[...]

PR #514, opened 2020-10-23
Individual Pin Inversion

anuejn:
This PR implements individual pin inversion as discussed in RFC #510.
I was unsure about changing the internal representation of invert from a single bool to a tuple of bool. Is that a breaking change and should methods in vendor still support being passed a single bool?

### DIFF
--- a/nmigen/build/dsl.py
+++ b/nmigen/build/dsl.py
@@ -11,6 +11,8 @@ class Pins:
             raise TypeError("Names must be a whitespace-separated string, not {!r}"
                             .format(names))
         names = names.split()
+        self.individual_inverts = tuple(name.startswith("~") for name in names)
+        names = [name[1:] if invert else name for name, invert in zip(names, self.individual_inverts)]
 
         if conn is not None:
             conn_name, conn_number = conn
@@ -38,6 +40,9 @@ class Pins:
     def __iter__(self):
         return iter(self.names)
 
+    def get_inverts(self):
+        return tuple(invert ^ self.invert for invert in self.individual_inverts)
+
     def map_names(self, mapping, resource):
         mapped_names = []
         for name in self.names:
@@ -49,9 +54,12 @@ class Pins:
             mapped_names.append(name)
         return mapped_names
 
+    @property
+    def names_string(self):
+        return " ".join(("~" if invert else "") + name for name, invert in zip(self.names, self.individual_inverts))
+
     def __repr__(self):
-        return "(pins{} {} {})".format("-n" if self.invert else "",
-            self.dir, " ".join(self.names))
+        return "(pins{} {} {})".format("-n" if self.invert else "", self.dir, self.names_string)
 
 
 def PinsN(*args, **kwargs):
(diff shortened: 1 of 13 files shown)


nmigen/build/dsl.py
```diff
@@ -70,9 +78,19 @@ def __init__(self, p, n, *, dir="io", conn=None, assert_width=None):
                             "and {!r} do not"
                             .format(self.p, self.n))
 
+        if self.p.individual_inverts != self.n.individual_inverts:
+            raise SyntaxError("The individual invertaitons of the positive and negative pins do not match.")
```
rroohhh, line 82:
```suggestion
            raise SyntaxError("The individual inversions of the positive and negative pins do not match.")
```

whitequark:
I would strongly prefer it if there was only a single `invert` property, containing a tuple of booleans or perhaps more nicely a single `int` that can be XOR'd in the vendor code.

anuejn:
> I would strongly prefer it if there was only a single `invert` property

In the DSL?
That would introduce a bunch of breaking changes (for example the repr of the Pins and DiffPairs objects), Moreover, the invert property of those DSL object would not be implementable as is (or am I wrong?).
However, I am fine with that as well.

whitequark:
> In the DSL?

Internally. Let me review your changes again more carefully first, though.

anuejn:
Okay :+1:

whitequark (changes requested):
As promised, I carefully reviewed the PR. Please do the following:

  * Replace the `invert` attribute in the DSL with a different attribute called `flip`*. The `flip` attribute would contain an integer that may be XOR'd with the buffer input or output to apply inversion as appropriate. Both the `~` prefix and the `invert=` keyword argument would affect this property.
     * It is almost never useful to iterate through the flips one by one (I think that's only used in iCE40), and even if it was, `zip()` of two lists isn't the nicest way to do it. If that ever becomes an operation we want to simplify, we should just add a custom iterator.
     * We can preserve the `invert` attribute by deprecating it at the same time as replacing it with a `@property` that would return `False` for `0`, `True` for `~0 & ((1 << width) - 1)`, and raise an exception for any other flip value. Feel free to do it if you think it is valuable; I strongly doubt anyone depends on `.invert`.
  * Emit a warning if `invert=` and `~`-prefix are used simultaneously (unless maybe there are some cases where it's reasonable...).
  * Revert the addition of `vendor.util`. The vendor support files are kept separate and with a large amount of redundancy very intentionally. I am aware of the downsides of that. Nevertheless, I am convinced that any common functionality should either be duplicated inline (as it currently happens), or turned into a public API elsewhere (which we aren't ready to do for something like `get_ireg`).

* I'm not married to the name `flip`, but it's the one I dislike the least.

whitequark:
@anuejn This PR is pretty old at this point. Are you still interested in pursuing it?

anuejn:
Oh sorry, I completely forgot about this PR. I still really want this functionality, and I am willing to continue. It would be sad if the work that already went into this was lost. 

That said, I am on holiday for the following two weeks but would continue afterward if it is okay for you. Otherwise also feel free to close this PR for now, and I'll reopen another when I get back to it.

whitequark:
No problem if you want to continue pursuing it. I think the existing RFC, despite not being written quite to the same high standard that we now require, seems to be good enough to pick up working on this feature. (Among my reasons are that it's not a core language feature and the platform system is likely to get a major reshape before 1.0.)

anuejn:
closing this and maybe re-implementing it on top of rfc-55 once it lands

PR #516, opened 2020-10-27
Handle installation when not running from a full git checkout

newhouseb:
I'm working on setting up a Nix shell environment, which builds nmigen from a checkout that does not contain a full .git directory. In the this scenario, setuptools_scm is installed correctly, but `parse`/`parse_git` returns `None` because there's no active git checkout.

As a fix, I check to see if `git` is falsey and, if so, return a blank version string. 

I've verified that this installs properly from nix via `fetchFromGithub`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,9 @@ def doc_version():
         return ""
 
     git = parse_git(".")
-    if git.exact:
+    if not git:
+        return ""
+    elif git.exact:
         return git.format_with("{tag}")
     else:
         return "latest"


whitequark (approved):

whitequark:
Is this for Nixpkgs?

newhouseb:
I might build up to submitting to Nixpkgs, but am just setting up a shell for the moment (I'm a nix noob). 

For future travelers, here's an example package definition (note this is against my fork, rather than the merged diff):

```
    nmigen = python38.pkgs.buildPythonPackage rec {
      pname = "nmigen";
      version = "0.2";

      src = pkgs.fetchFromGitHub {
        name = "nmigen";
        owner = "newhouseb";
        repo = "nmigen";
        rev = "6e585a6ad850cbc02ab730acbb01f17f08eb7152";
        sha256 = "121fa20cd5q9ijbihpcj04j2pil78h2i1gdciq2ysj2vjkg46zk8";
        fetchSubmodules = true;
      };

      doCheck = false;

      propagatedBuildInputs = [ python38.pkgs.numpy python38.pkgs.jinja2 python38.pkgs.pyvcd python38.pkgs.importlib-resources ];
      nativeBuildInputs = [ python38.pkgs.setuptools_scm ];
      preBuild = ''
           export SETUPTOOLS_SCM_PRETEND_VERSION="0.2"
           '';

      meta = {
        homepage = "https://github.com/nmigen/nmigen";
        description = "A refreshed Python toolbox for building complex digital hardware";
      };
    };
```

whitequark:
> I might build up to submitting to Nixpkgs, but am just setting up a shell for the moment (I'm a nix noob).

Nixpkgs should already have an nMigen package.

whitequark:
> `export SETUPTOOLS_SCM_PRETEND_VERSION="0.2"`

Ah I wasn't aware of that. It would have been nicer to return that instead of `""` in your patch, though it's not especially important either way (it only really matters for PyPI I think).

newhouseb:
Ah, I missed that Nixpkgs already has a [package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/nmigen/default.nix), looking now it's basically what I posted, but better (although pinned to a version from April, before the offending code was landed).

My macro use-case is pinning a deterministic set of of newish {lambdasoc, nmigensoc, nmigen, minerva, etc} with my code, since the aforementioned packages are all (understandably) moving targets.

PR #517, opened 2020-10-27
lib.fifo.AsyncFFSynchronizer: check input and output signal width

rroohhh:
wasted more time than I would like to admit due to this :)


whitequark (changes requested):
A few nits:
 * By convention, all error checking happens at the function entry.
 * The error messages would be easier to read and more consistent with our other error messages if they used `1`, not `one`.

rroohhh:
`async_edge` is also not checked at the function entry, should that be changed aswell?

whitequark:
Yup.

whitequark (approved):

whitequark:
Thanks!

PR #524, opened 2020-11-03
lib.fifo: fix level on fifo full

anuejn:
This is the split out stuff from #512 that affects AsyncFIFO and not AsyncFIFOBuffered


whitequark (approved):

whitequark:
Thanks!

PR #529, opened 2020-11-06
Migrate Record to ValueCastable, from UserValue

### LINKED ISSUE #528: Redesign Record to be based on ValueCastable
`UserValue` will be removed in nMigen 0.4 per RFC #355. Although `Record` (which currently derives from `UserValue`) is also slated for deprecation and eventual removal, its replacement is not implemented (or designed) yet, and a significant amount of external code depends on it. Because of this `Record` will survive significantly longer than `UserValue` (likely several releases), and has to be unaffected by its removal.

The (undesirable and hazardous!) behavior that `Record` currently has can be simulated by overriding `__getattr__` to make `Record` behave as if it inherits from `Value` without actually doing so.

awygle:
This PR migrates Record from being based on UserValue to being based on ValueCastable, clearing the way to deprecate UserValue.

Would close #528 .

### DIFF
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from collections import OrderedDict
-from functools import reduce
+from functools import reduce, wraps
 
 from .. import tracer
 from .._utils import union, deprecated
@@ -149,7 +149,10 @@ class Record(ValueCastable):
             # it seems we lose the implicit `self` argument at some point, so curry manually
             res = getattr(Value, name)
             if callable(res):
-                return lambda *args, **kwargs: res(self, *args, **kwargs)
+                @wraps(res)
+                def _wrapper(*args, **kwargs):
+                    return res(self, *args, **kwargs)
+                return _wrapper
             return res
         except AttributeError:
             return self[name]
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -145,7 +145,6 @@ class Record(ValueCastable):
     def __getattr__(self, name):
         # must check `getattr` before `self` - we need to hit Value methods before fields
         try:
-            # it seems we lose the implicit `self` argument at some point, so curry manually
             value_attr = getattr(Value, name)
             if callable(value_attr):
                 @wraps(value_attr)
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -147,13 +147,13 @@ class Record(ValueCastable):
         # must check `getattr` before `self` - we need to hit Value methods before fields
         try:
             # it seems we lose the implicit `self` argument at some point, so curry manually
-            res = getattr(Value, name)
-            if callable(res):
-                @wraps(res)
+            value_attr = getattr(Value, name)
+            if callable(value_attr):
+                @wraps(value_attr)
                 def _wrapper(*args, **kwargs):
-                    return res(self, *args, **kwargs)
+                    return value_attr(self, *args, **kwargs)
                 return _wrapper
-            return res
+            return value_attr
         except AttributeError:
             return self[name]
 
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -85,7 +85,6 @@ class Layout:
         return "Layout([{}])".format(", ".join(field_reprs))
 
 
-# Unlike most Values, Record *can* be subclassed.
 class Record(ValueCastable):
     @staticmethod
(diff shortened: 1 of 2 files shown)


nmigen/hdl/rec.py
```diff
@@ -86,7 +86,7 @@ def __repr__(self):
 
 
 # Unlike most Values, Record *can* be subclassed.
```
whitequark, line 88: Record is no longer a Value.

nmigen/hdl/rec.py
```diff
@@ -166,11 +175,23 @@ def __getitem__(self, item):
...
+                raise AttributeError("{} does not have a field '{}'. Did you mean one of: {}?"
+                                     .format(reference, item, ", ".join(self.fields))) from None
 
-    def lower(self):
+    @ValueCastable.lowermethod
+    def as_value(self):
         return Cat(self.fields.values())
 
+    def __len__(self):
+        return len(Cat(self.fields.values()))
```
whitequark, line 193: Doing `len(self.as_value())` here would result in less garbage because of memoization (nMigen calls len() a lot so it might actually matter).

nmigen/hdl/rec.py
```diff
@@ -146,7 +144,18 @@ def concat(a, b):
                                                      src_loc_at=1 + src_loc_at)
 
     def __getattr__(self, name):
-        return self[name]
+        # must check `getattr` before `self` - we need to hit Value methods before fields
+        try:
+            # it seems we lose the implicit `self` argument at some point, so curry manually
```
whitequark, line 149: Python adds a `self` argument on its own in exactly one (AFAIK) case: when an attribute access on an *object* resolves to an attribute of its *class* and this attribute is *callable*. Since in this case you are accessing a class attribute, not an object attribute, you simply fetch the function from the class unmodified:
```pyrepl
>>> class c:
...     def f(x,y): return x,y
...
>>> c.f
<function c.f at 0x7fb25aeafe18>
>>> c.f(1,2)
(1, 2)
```
There's nothing special about the code above--the class acts as a fancy dictionary.

nmigen/hdl/rec.py
```diff
@@ -146,7 +144,18 @@ def concat(a, b):
                                                      src_loc_at=1 + src_loc_at)
 
     def __getattr__(self, name):
-        return self[name]
+        # must check `getattr` before `self` - we need to hit Value methods before fields
```
whitequark, line 147: Good catch--I had to double-check the old behavior.

nmigen/hdl/rec.py
```diff
@@ -146,7 +144,18 @@ def concat(a, b):
...
+        try:
+            # it seems we lose the implicit `self` argument at some point, so curry manually
+            res = getattr(Value, name)
```
whitequark, line 150: Maybe something like
```suggestion
            value_attr = getattr(Value, name)
```
?

whitequark (changes requested):

nmigen/hdl/rec.py
awygle: To make sure I understand what you're saying - I'm doing the right thing here, but the comment displaying my confusion is unnecessary?

whitequark: More or less.

nmigen/hdl/rec.py
```diff
@@ -146,7 +143,18 @@ def concat(a, b):
                                                      src_loc_at=1 + src_loc_at)
 
     def __getattr__(self, name):
-        return self[name]
+        # must check `getattr` before `self` - we need to hit Value methods before fields
+        try:
+            # it seems we lose the implicit `self` argument at some point, so curry manually
+            value_attr = getattr(Value, name)
+            if callable(res):
```
whitequark, line 150: The sad part of this GitHub feature is that I think it only works one line at a time...
```suggestion
            if callable(value_attr):
```

awygle: Yep, was just seeing that. I'll fix it in a proper commit...

whitequark (approved):

whitequark:
Oh, a good addition to this PR would be to rip out every place where Record is treated somehow differently in the middle end and back end. (I don't think we have any in the front end.)

awygle:
I don't actually see any of those? A `grep -R Record nmigen/` shows `Record` being _used_ in a few places but I only see things of the form `isinstance(x, Record)` or similar in the RPC server and the compat layer, both of which I'd consider to be special cases.

whitequark:
> I don't actually see any of those?

Oh right! We already ripped those out after we migrated it to `UserValue`.

PR #530, opened 2020-11-06
intel: Add support for Cyclone V internal oscillator

kbeckmann:
When using the default clock "cyclonev_oscillator" on Cyclone V devices, the internal oscillator will be used.

### DIFF
--- a/nmigen/vendor/intel.py
+++ b/nmigen/vendor/intel.py
@@ -146,6 +146,31 @@ class IntelPlatform(TemplatedPlatform):
         super().add_clock_constraint(clock, frequency)
         clock.attrs["keep"] = "true"
 
+    @property
+    def default_clk_constraint(self):
+        # Internal high-speed oscillator on Cyclone V devices.
+        # It is specified to not be faster than 100MHz, but the actual
+        # frequency seems to vary a lot between devices. Measurements
+        # of 78 to 84 MHz have been observed.
+        if self.default_clk == "cyclonev_oscillator":
+            assert self.device.startswith("5C")
+            return Clock(100e6)
+        # Otherwise, use the defined Clock resource.
+        return super().default_clk_constraint
+
+    def create_missing_domain(self, name):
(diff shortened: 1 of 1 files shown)


nmigen/vendor/intel.py
```diff
@@ -146,6 +146,29 @@ def add_clock_constraint(self, clock, frequency):
         super().add_clock_constraint(clock, frequency)
         clock.attrs["keep"] = "true"
 
+    @property
+    def default_clk_constraint(self):
+        # Internal high-speed oscillator on Cyclone V devices.
+        # It is specified to not be faster than 100MHz, but the actual
+        # frequency seems to vary a lot between devices. Measurements
+        # of 78 to 84 MHz have been observed.
+        if self.default_clk == "cyclonev_oscillator" and self.device.startswith("5C"):
```
whitequark, line 155: The check for `self.device.startswith("5C")` should be an assertion, since otherwise the error message you get if you misconfigure the oscillator is less clear.

nmigen/vendor/intel.py
```diff
@@ -146,6 +146,29 @@ def add_clock_constraint(self, clock, frequency):
...
+            return Clock(100e6)
+        # Otherwise, use the defined Clock resource.
+        return super().default_clk_constraint
+
+    def create_missing_domain(self, name):
+        if name == "sync" and self.default_clk == "cyclonev_oscillator" and self.device.startswith("5C"):
```
whitequark, line 161: Here too.

whitequark (changes requested):

nmigen/vendor/intel.py
kbeckmann: Thanks, fixed it.

whitequark (approved):

whitequark:
Thanks!

kbeckmann:
Thanks for the quick feedback, happy to contribute!

PR #537, opened 2020-11-07
Improve error message.

BracketMaster:
I run into this error every so often and thought
this could be a more helpful error message.

### DIFF
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -47,7 +47,10 @@ class Fragment:
                 code = obj.elaborate.__code__
                 obj = obj.elaborate(platform)
             else:
-                raise AttributeError("Object {!r} cannot be elaborated".format(obj))
+                raise AttributeError(
+                    "Object {!r} cannot be elaborated. " + \
+                    "Perhaps you forgot ``return m`` where m is of type ``Module()``."
+                    .format(obj))
             if obj is None and code is not None:
                 warnings.warn_explicit(
                     message=".elaborate() returned None; missing return statement?",


BracketMaster:
Assuming anyone else runs into this...

BracketMaster:
Or maybe it could say ``Expected Module not {!r}``?

whitequark:
There's a warning (just below your change) that says the exact same thing, but also points to the `elaborate` function in question which I think should be more useful. Does it not fire for you?

BracketMaster:
Oh. You're right. My fault for not noticing that.

whitequark:
I think I should do *something* with that warning, if people don't notice it, it's as good as not having it in first place...

BracketMaster:
Maybe making it yellow?

whitequark:
That's... hard to do in a good way. Think of the people with white terminals where it'll be unreadable.

BracketMaster:
don't forget, you can change block backgrounds per character.

whitequark:
Right. That will then break on Windows if I use direct escape sequences. And if I want to do it portably it'll be a new dependency and people would complain about *that*. Colors aren't a good solution.

BracketMaster:
This is tricky...
I'm honestly surprised that Windows users use nMigen in 
Windows-Python native, instead of just using Bash-for-Windows.

Is red not a good color? GCC and clang always seem to
emit red regardless of the color scheme. And honestly...
if you have a red terminal background, you're asking for 
trouble...

whitequark:
> I'm honestly surprised that Windows users use nMigen in Windows-Python native, instead of just using Bash-for-Windows.

One thing I've learned is that no matter how liberal my assumptions about the downstream environment are, they are always insufficient to cover the workflow of a surprising amount of people. I think every nMigen tutorial that exists describes a completely different installation method, and that's just on Linux!

whitequark:
Also, Windows Store has an official Python package, so if you just type `python` it prompts you to install that. That package actually works pretty well, so I'm happy to have people follow that path.

BracketMaster:
You can also just ask nMigen users to read errors 
more carefully. Something I should do more often
in general, ha

BracketMaster:
I think I've also seen error messages prefixed with
a sign like ``==> error foo on line 1234``.

mithro:
Here is a horrible suggestion (but /might/ work);
 (a) Collect all warnings as they are emitted.
 (b) On exit, output a summary of the warnings that were emitted (maybe things like the number of each type of warning).
 (c) If a traceback / error occurs, make sure to remit the warnings just before the trace back output.

I would generally say that anything which isn't the in the last few lines of output is going to be very easily ignored?

Maybe you could also having a `-Werror` type thing that is **enabled by default** and people can turn off if needed?

whitequark:
> (c) If a traceback / error occurs, make sure to remit the warnings just before the trace back output.

Just before (i.e. above traceback) or just after (below traceback)?

BracketMaster:
Just to confirm, I was ignoring what was above traceback
as I assumed it _was_ part of the traceback.

UserWarning _was_ above traceback.

So collecting and going below might be better.

PR #541, opened 2020-11-09
Fix Record Operators

### LINKED ISSUE #533: Breaking change in Record
abbebf8efe931938f0de95d041f1d91c693efddc introduced a breaking change in Record:
Some (all?) operators are not overloaded on it anymore because they are looked up on the class rather than on the instance so` __getattr__()` doesnt forward them.

pre abbebf8efe931938f0de95d041f1d91c693efddc:
```python
>>> Record([("a", 1)])
(rec <unnamed> a)
```

post abbebf8efe931938f0de95d041f1d91c693efddc:
```python
>>> Record([("a", 1)]) + 1
TypeError: unsupported operand type(s) for +: 'Record' and 'int'
```

awygle:
Proxy Operator calls on Record through to Value implementation, using explicit enumeration to avoid depending on future changes to Value.

Fixes #533 .

### DIFF
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -142,18 +142,15 @@ class Record(ValueCastable):
                     self.fields[field_name] = Signal(field_shape, name=concat(name, field_name),
                                                      src_loc_at=1 + src_loc_at)
 
+    def _valueproxy(name):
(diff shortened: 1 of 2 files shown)


whitequark (approved):

whitequark:
Thanks!

PR #543, opened 2020-11-13
vendor.quicklogic: enable SoC clock configuration

kowalewskijan:
Related PR: https://github.com/nmigen/nmigen-boards/pull/117
Signed-off-by: Jan Kowalewski <jkowalewski@antmicro.com>

### DIFF
--- a/nmigen/vendor/quicklogic.py
+++ b/nmigen/vendor/quicklogic.py
@@ -129,10 +129,18 @@ class QuicklogicPlatform(TemplatedPlatform):
         python3 -m quicklogic_fasm.bitstream_to_openocd
             {{name}}.bit
             {{name}}.openocd
+            --osc-freq {{platform.osc_freq}}
+            --fpga-clk-divider {{platform.osc_div}}
         """,
     ]
 
     # Common logic
+
+    @property
+    def default_clk_constraint(self):
+        if self.default_clk == "sys_clk0":
+            return Clock(self.osc_freq / self.osc_div)
+        return super().default_clk_constraint
 
     def add_clock_constraint(self, clock, frequency):
         super().add_clock_constraint(clock, frequency)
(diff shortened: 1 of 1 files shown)


whitequark (commented):
The code looks basically reasonable. However, I don't understand the meaning of `--osc-freq` and `--fpga-clk-divider` being provided at the same time. Normally, I would expect an on-chip oscillator, such as a ring oscillator, to have a fixed frequency, and the system clock frequency could be configured by specifying a divisor explicitly, or providing a frequency and letting the tools come up with the closest matching divisor.

Could you explain the relevant part of the clock tree to me, or refer me to a section of the EOS S3 TRM that does?

kowalewskijan:
Here is the TRM link related to clocks: https://www.quicklogic.com/wp-content/uploads/2020/06/QL-S3-Technical-Reference-Manual.pdf#page=88
Table on page 90 ([link](https://www.quicklogic.com/wp-content/uploads/2020/06/QL-S3-Technical-Reference-Manual.pdf#page=90)) provides information about maximum frequencies for FPGA domain.

Currently I can see that the `bitstream_to_openocd` tool [assumes](https://github.com/QuickLogic-Corp/quicklogic-fasm/blob/master/quicklogic_fasm/bitstream_to_openocd.py#L61) that the divider is always on, so it is not possible for now to achieve frequencies beyond 40MHz (which is max 80MHz divided by minimum divisor: 2). I think it would be nice to have an option to configure higher frequency in future like maximum 72MHz for FPGA domain, to do so we'll need `--osc-freq` . We can set default `osc_freq` and `osc_div` in platform definition thus setting them would be optional in board definition.

> or providing a frequency and letting the tools come up with the closest matching divisor.

I agree this is something that is missing in `bitstream_to_openocd` tool. For now it is a very manual approach if user wants to configure other frequency for FPGA domain than the default one.

whitequark (approved):

whitequark:
Thanks. I've reviewed the EOS S3 clocking and the approach you picked is perfectly fine.

PR #544, opened 2020-11-14
pysim: Run mask on Mux selection operand.

mwkmwkmwk:
Otherwise it behaves funny when it's eg. the result of operator ~.


whitequark (commented):
The change is correct (good catch!)

Could you add a test?

mwkmwkmwk:
Added a test; It seems we should have all operators go through similar treatment though, to avoid similar problems.

whitequark:
> It seems we should have all operators go through similar treatment though, to avoid similar problems.

The original idea was to use `self(x)` rather than `mask(x)` where it's safe to do so as an optimization. But it keeps backfiring..

whitequark (approved):

PR #545, opened 2020-11-17
Implement shape() for Record

awygle:
Record lost shape() in the migration to ValueCastable. This PR restores it.


whitequark (approved):

whitequark:
Thank you!

Issue #546, opened 2020-11-19
platform.add_clock_constraint does not work for instances with lattice diamond for machxo2


anuejn:
Similiar to #373 platform.add_clock_constraint does not work for instances in lattice diamond.
However, the same fix does not seem to be applicable. Do you have any ideas what is going wrong?

repro:
```
from nmigen import *
from nmigen.build import *
from nmigen_boards.tinyfpga_ax2 import *

class Test(Elaboratable):
    def elaborate(self, plat):
        m = Module()
        clk = Signal()
        m.submodules.inst = Instance("JTAGF", o_JTCK=clk)

        m.domains += ClockDomain("sync")
        m.d.comb += ClockSignal().eq(clk)

        plat.add_clock_constraint(clk, 2e6)

        plat.add_resources([Resource("gpio", 0, Pins("1", conn=("gpio", 0), dir="o"), Attrs(IOSTANDARD="LVCMOS33"))])
        gpio0 = plat.request("gpio", 0)
        m.d.sync += gpio0.eq(~gpio0)

        return m

TinyFPGAAX2Platform().build(Test())
```

whitequark:
What is the error you're getting?

anuejn:
in `build/top_impl/top_impl_cck.rpt`:
```
create_clock -name clk -period 500.0 [get_nets {clk$1}]
	@E::"/home/anuejn/tmp/repro/build/top.sdc":1:0:1:0|Source for clock clk not found in netlist.
```

anuejn:
[build.zip](https://github.com/nmigen/nmigen/files/5568856/build.zip)

whitequark:
It probably doesn't like the `$`. That doesn't happen for ports because ports are named first. Try `Instance("JTAGF", o_JTCK=ClockSignal())`.

anuejn:
How can I constrain it then?

whitequark:
```
cd_sync = ClockDomain()
platform.add_clock_constraint(cd_sync.clk, freq)
```

anuejn:
Thanks that works. Somehow it doesnt work for my non minimized use case :(. I will try to find another (less minimized) example

anuejn:
This also doesnt work, probably due to hierarchy:
```python
from nmigen import *
from nmigen.build import *
from nmigen_boards.tinyfpga_ax2 import *

class Test(Elaboratable):
    def elaborate(self, plat):
        m = Module()

        m.submodules.jtag = JTAG()

        return m

[...]

PR #547, opened 2020-11-19
vendor.lattice_machxo_2_3l: fix sdc generation

anuejn:
This PR fixes both issues reported in #546 .
Probably there is a similiar issue for ecp5 and ice40 since they both use synplify.

### DIFF
--- a/amaranth/build/plat.py
+++ b/amaranth/build/plat.py
@@ -426,6 +426,12 @@ class TemplatedPlatform(Platform):
         def tcl_quote(string):
             return '"' + re.sub(r"([$[\\])", r"\\\1", string) + '"'
 
+        def sdc_quote(string):
+            # synplify used in lattice diamond requires different quoting
+            # than standard tcl quoting: wrapped in curly braces but escaped
+            # as if wrapped in quotation marks.
+            return "{" + re.sub(r"([${}\\])", r"\\\1", string) + "}"
+
         def verbose(arg):
             if get_override_flag("verbose"):
                 return arg
@@ -448,6 +454,7 @@ class TemplatedPlatform(Platform):
                 compiled.environment.filters["ascii_escape"] = ascii_escape
                 compiled.environment.filters["tcl_escape"] = tcl_escape
                 compiled.environment.filters["tcl_quote"] = tcl_quote
+                compiled.environment.filters["sdc_quote"] = sdc_quote
             except jinja2.TemplateSyntaxError as e:
                 e.args = (f"{e.message} (at {origin}:{e.lineno})",)
                 raise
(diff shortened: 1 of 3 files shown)


whitequark:
Why not use `tcl_quote`?

anuejn:
I thought that this way it is more clear that we are working around a bug here.
But I can change that.

anuejn:
Also if we simply chain tcl_escape and tcl_quote, we end up transforming `\` to `\\\\`

whitequark:
That function is also for working around a similar issue in Vivado...

anuejn:
So is it safe to just use tcl_quote even if it doesnt escape `{` and `}`

whitequark:
Yes, `tcl_escape` and `tcl_quote` are mutually exclusive and they have the same function.

whitequark:
Please do the same for (at least) ECP5, though if you have iCECube, then iCE40 as well.

anuejn:
Should I do it as part of this pr?

whitequark:
Yup.

anuejn:
What we need here is actually really something different than tcl_quote. It seems like synplify really expects the net to be wrapped in curly braces but escaped as if they were wrapped in quotation marks. Just using tcl_quote doesnt work :(.

whitequark:
Ok well that's horrible (and explains why I couldn't get it right several times already...) In that case please add `sdc_quote` or something like that.

whitequark:
@anuejn Have you had a chance to give this another look?

anuejn:
sorry for the long delay. the code I pushed now seems to work for me.

nmigen/build/plat.py
```diff
@@ -395,6 +395,9 @@ def tcl_escape(string):
         def tcl_quote(string):
             return '"' + re.sub(r"([$[\\])", r"\\\1", string) + '"'
 
+        def sdc_quote(string):
+            return "{" + re.sub(r"([${}\\])", r"\\\1", string) + "}"
```
whitequark, line 399: Would be nice to have a comment here explaining that this separate method is necessary because SDC does not actually behave like Tcl.

whitequark (changes requested):
Can I ask you to split this PR into two separate commits? As far as I understand/remember (correct me if I'm wrong), the SDC quoting issue and the hierarchy separator issue are different, so it would be a lot easier to understand the VCS history later if they were split. In cases like this it's OK to do cross-cutting changes through the entire `vendor/` subdirectory.

nmigen/build/plat.py
whitequark: Wasn't the whole point of adding this function the fact that *all* Synplify frontends use this quoting?

anuejn: that would be ice40 in addition to the two I already changed, right?
sadly I do not have icecube installed so i cannot test easily. should I change it anyways?

whitequark: Do you have a testcase handy? I can check all of the currently supported toolchains since I have all of them.

whitequark:
I don't believe this PR fixes the issue it claims to fix. I still get this error:

```
@W: MT548 :"/home/whitequark/Projects/amaranth/build/top.sdc":2:0:2:0|Source for clock clk not found in netlist.
```

PR #571, opened 2021-01-02
AsyncFIFOBuffered {r,w}_level fixes

rroohhh:
This includes a number of fixes for `{r,w}_level` in AsyncFIFOBuffered.

1. The accounting for the output register used `AsyncFFSynchronizer` instead of `FFSynchronizer`. This is wrong as we want to synchronize both edges
2. Use the proper clock domains in the tests (not really a bug in the fifo logic obviously)
3. The FFSynchronizer should have a latency of 4 to match the latency of `consume_w_bin` / `w_level`. Two cycles for the `FFSynchronizer` from the read to the write domain of `consume_w_gry`, one for the `GrayDecoder` and one for the `w_level` calculation.
4. Fix the output register accounting. If the user asserts `r_en` even if `r_rdy` is low, we get `-1` in the intermediate calculation which overflows to `1` as we have a unsigned `Signal`

Let me know if I should split this into multiple PRs or elaborate more on the fixes. The last two include additional testcases that fail without the fix. I did not include a testcase for the `AsyncFFSynchronizer` to `FFSynchronizer` change, as this is obviously wrong and would be hard to properly test without proper multi clock support in formal.

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -509,7 +509,7 @@ class AsyncFIFOBuffered(Elaboratable, FIFOInterface):
         ]
 
         r_consume_buffered = Signal()
-        m.d.comb += r_consume_buffered.eq(self.r_rdy - self.r_en)
+        m.d.comb += r_consume_buffered.eq((self.r_rdy - self.r_en) & self.r_rdy)
         m.d[self._r_domain] += self.r_level.eq(fifo.r_level + r_consume_buffered)
 
         w_consume_buffered = Signal()
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -513,7 +513,7 @@ class AsyncFIFOBuffered(Elaboratable, FIFOInterface):
         m.d[self._r_domain] += self.r_level.eq(fifo.r_level + r_consume_buffered)
 
         w_consume_buffered = Signal()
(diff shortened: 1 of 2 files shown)


whitequark:
> Let me know if I should split this into multiple PRs or elaborate more on the fixes.

Everything looks good. However, does anyone remember how we ended using `AsyncFFSynchronizer` there in the first place? I vaguely remember that the ability to use it there motivated a refactoring elsewhere, so it was probably not a simple typo, and I'd like to understand why it's there before we change it. It is also possible that I misremember.

anuejn:
To me it happens quite often that I end up using AsyncFFSyncronizer by accident because of the naming. My thought process is something like "I need to synchronize a signal that is async to $domain into $domain. Must be AsyncFFSyncronizer". I think it is very likely that it slipped in there similarly?

whitequark:
@awygle mentioned on IRC that he thinks it's deliberate; let's wait for him to investigate.

awygle:
I misread this because of what @whitequark said about it motivating a refactoring. This is a different case than what I was thinking of (the reset `AsyncFFSynchronizer`, which is what that module was invented for). This seems correct to me, I am OK with merging it.

whitequark (approved):

whitequark:
Thanks everyone!

Issue #572, opened 2021-01-05
Add option to suppress code generation comments


hansfbaier:
When generating code, almost every line has code generation comments.
While this is useful for debugging, it makes the code very hard to read.
Therefore it would be great to have a code generation option
to suppress those.

hellow554:
hey @hansfbaier 
you seem to have been working on this issue (according to the three references commits here).
What is your status on this? Are you able to submit a PR?

hansfbaier:
@hellow554 : I have this: https://github.com/hansfbaier/nmigen/commit/304ad1ffa2b64fe60cb709d9dad2cb9edaf9d87c
But I am hesitant to create a pull request, because whitequark seems opposed to having this functionality in nmigen,
see https://github.com/nmigen/nmigen/pull/573

whitequark:
I think the functionality is fine, the implementation just needs to be improved. I apologize for my disproportionally intense reaction, @hansfbaier--I'll work something out to have this upstream.

hansfbaier:
@whitequark  Thank you!

PR #573, opened 2021-01-06
#572: add option to suppress generated source line comments

hansfbaier:
Hi I implemented the feature.
Since for verilog the functionality was already there,
I did a regex replace on the other two.

### DIFF
--- a/nmigen/cli.py
+++ b/nmigen/cli.py
@@ -1,4 +1,5 @@
 import argparse
+import re
 
 from .hdl.ir import Fragment
 from .back import rtlil, cxxrtl, verilog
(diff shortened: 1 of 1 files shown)


whitequark:
Using a regex for this purpose is completely unacceptable.

hansfbaier:
@whitequark  can you give me a clue to where the right place would be for IL? I looked into it, but the src comments are all over the  place in the code.

whitequark:
I don't think this is necessary to do for the IL at all.

hansfbaier:
I got it: in _ModuleBuilder.__enter__. I would like to actually read and understand the IL too betimes, and also there the source line comments really get in the way of readability

whitequark:
In that case you can use Yosys to remove the comments like it is done in `back.verilog`.

hansfbaier:
You mean, run yosys to filter in rtlil.py:convert or external to nmigenl?

PR #575, opened 2021-01-13
Lattice: Add support for MachXO2/XO3L internal oscillator

jreyesr:
This PR adds support for the internal high-speed oscillator on MachXO2/XO3L devices, called OSCH. Migen had support for the oscillator, but nMigen doesn't have it yet.

It essentially replicates the changes that https://github.com/nmigen/nmigen/pull/530 made on Intel Cyclone V devices and https://github.com/nmigen/nmigen/pull/338 on the iCE40, with the difference that MachXO2/XO3 devices have a configurable frequency and a list of allowed freqs, as opposed to the Cyclone V which has a fixed frequency with a maximum of 100MHz, and the iCE40 which uses a fixed 48MHz oscillator and a divider.

This PR is accompanied by (and would allow) PR https://github.com/nmigen/nmigen-boards/pull/136 in nmigen-boards, which changes the TinyFPGAAX2Platform to use the internal clock (at 2.08 MHz) by default, since the TinyFPGA AX2 has no external clock and thus no better default alternative for the clock signal.

### DIFF
--- a/amaranth/vendor/lattice_machxo_2_3l.py
+++ b/amaranth/vendor/lattice_machxo_2_3l.py
@@ -154,6 +154,27 @@ class LatticeMachXO2Or3LPlatform(TemplatedPlatform):
             -if {{name}}_impl/{{name}}_impl.bit -of {{name}}_sram.svf
         """,
     ]
+    # These numbers were extracted from
+    # "MachXO2 sysCLOCK PLL Design and Usage Guide"
+    _supported_osch_freqs = [
+        2.08, 2.15, 2.22, 2.29, 2.38, 2.46, 2.56, 2.66, 2.77, 2.89,
+        3.02, 3.17, 3.33, 3.50, 3.69, 3.91, 4.16, 4.29, 4.43, 4.59,
+        4.75, 4.93, 5.12, 5.32, 5.54, 5.78, 6.05, 6.33, 6.65, 7.00,
+        7.39, 7.82, 8.31, 8.58, 8.87, 9.17, 9.50, 9.85, 10.23, 10.64,
+        11.08, 11.57, 12.09, 12.67, 13.30, 14.00, 14.78, 15.65, 15.65, 16.63,
+        17.73, 19.00, 20.46, 22.17, 24.18, 26.60, 29.56, 33.25, 38.00, 44.33,
+        53.20, 66.50, 88.67, 133.00
+    ]
+
+    @property
+    def default_clk_constraint(self):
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
Our other platforms that support a configurable internal clock source, iCE40, ECP5, and QuickLogic, use a divider to set the frequency. This simplifies the logic, in particular it means that any input configuration exactly (within the limits of the hardware) specifies output frequency. In addition, ECP5 uses the same cell library as MachXO2/XO3L and is a broadly similar family of devices, so they should definitely use the same general approach.

Please change your implemenation to closely follow the logic in the ECP5 platform.

jreyesr:
There's a difference between iCE40, ECP5, and QuickLogic devices, on the one hand, and MachXO2, on the other. On MachXO2 devices, the oscillator is configured by explicitly setting the frequency as a Verilog parameter. See [pages 35 to 37 here](http://www.latticesemi.com/~/media/LatticeSemi/Documents/ApplicationNotes/MO/MachXO2sysCLOCKPLLDesignandUsageGuide.pdf?document_id=39080) for an example. 

The allowed frequencies _do_ seem to come from a 266 MHz oscillator with a configurable divider with ratios between 2 to 128 (this is not in the documentation, it simply fits the list provided and you can't set the oscillator frequency by passing a divider). Hovewer, not all ratios are valid. Ratios 2 to 32 are all valid, ratios between 32 to 64 are only valid if divisible by 2, and ratios between 64 and 128 are valid if divisible by 4.

Should I still go ahead and use a divider ratio as the configurable parameter, as opposed to the raw frequency that is now being used? If so, there will still be "dead spots" in the input values (i.e., it won't be enough to just "enter a divider between 2 and 128", as in the other families, since, say, 65 would request a frequency of 4.09 MHz, which is not in the list of valid frequencies). Is that acceptable?

whitequark:
> Should I still go ahead and use a divider ratio as the configurable parameter, as opposed to the raw frequency that is now being used?

No, with this additional information, I believe your approach is correct. I'll take another look soon.

modwizcode (dismissed):
LGTM, just needs to be rebased I think.

jreyesr:
> LGTM, just needs to be rebased I think.

Done. I just made a brand new commit, since my code was outdated by almost a year.

modwizcode (approved):
Thanks for rebasing.

whitequark (approved):
LGTM

whitequark:
Thank you, and apologies for the delay!

jreyesr:
> Thank you, and apologies for the delay!

Of course, no worries. Thank you!

PR #577, opened 2021-01-14
Remove outdated comment in ECP5 platform

adamgreig:
Starting with nextpnr c6401413a, nextpnr does pack *FS1P3DX into IOLOGIC cells.


whitequark (approved):

whitequark:
Thanks!

PR #591, opened 2021-02-09
lib.fifo: fix reset handling of asynchronous FIFOs.

jfng:
Asynchronous FIFOs do not properly handle a write domain reset (i.e. `w_level` is not reset to 0). This PR attempts to fix their behaviour.

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -355,7 +355,7 @@ class AsyncFIFO(Elaboratable, FIFOInterface):
         consume_enc = m.submodules.consume_enc = \
             GrayEncoder(self._ctr_bits)
         consume_cdc = m.submodules.consume_cdc = \
-            FFSynchronizer(consume_r_gry, consume_w_gry, o_domain=self._w_domain)
+            FFSynchronizer(consume_r_gry, consume_w_gry, o_domain=self._w_domain, reset_less=False)
         m.d.comb += consume_enc.i.eq(consume_r_nxt)
         m.d[self._r_domain] += consume_r_gry.eq(consume_enc.o)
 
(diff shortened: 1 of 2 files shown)


whitequark (commented):
Looks good and passes tests; I'm going to merge this after checking how it works with Glasgow Interface Explorer (where this issue is readily observed with the QSPI gateware).

amaranth/lib/fifo.py
```diff
@@ -477,24 +478,26 @@ def elaborate(self, platform):
...
         rst_cdc = m.submodules.rst_cdc = \
             AsyncFFSynchronizer(w_rst, r_rst, o_domain=self._r_domain)
 
-        # Decode Gray code counter synchronized from write domain to overwrite binary
-        # counter in read domain.
+        # Decode Gray code counter synchronized from read domain to overwrite binary counter in
+        # write domain.
+        with m.If(w_rst):
+            m.d[self._w_domain] += produce_w_gry.eq(consume_w_gry)
+            m.d[self._w_domain] += produce_w_bin.eq(_gray_decode(consume_w_gry))
```
jfng, line 497: This does not work: if a write domain reset happens one clock cycle after a read, then the next value of `consume_w_gry` is still inside an `FFSynchronizer`.

jfng (commented):
After discussing with @whitequark, it is no longer clear whether the current approach of propagating write domain resets to the read side is correct; doing so empties the FIFO, which can violate the invariant of a stream interface (`valid` must not be deasserted on it own).

Ideally, there would be a configurable "primary" and "secondary" side, where the primary side controls the reset of the secondary side.

In any case, further iterations of this change will require thorough testing and would benefit from (multi-clock) BMC or fuzzing.

PR #595, opened 2021-02-11
raise error on space in fsm name

hellow554:
This testcase will throw an error on gtkwave start:

```python
from nmigen.sim import Delay, Simulator

class A(Elaboratable):
    def elaborate(self, _):
        m = Module()

        with m.FSM(name="space here"):
            with m.State(0):
                pass
        return m

if __name__ == "__main__":
    def proc():
        yield Delay(1e-4)

    m = Module()
    m.submodules.a = A()
    sim = Simulator(m)
    sim.add_process(proc)
    with sim.write_vcd("broken.vcd"):
        sim.run()
```

```
GTKWave Analyzer v3.3.108 (w)1999-2020 BSI

Near byte 172, $VAR parse error encountered with 'top.a.space'
No symbols in VCD file..nothing to do!
```

After this: nmigen will throw this error:

```
  File "/home/marcel/projects/fpgastuff/a.py", line 21, in <module>
    sim = Simulator(m)
  File "/home/marcel/projects/fpgastuff/.VE/lib/python3.9/site-packages/nmigen/sim/core.py", line 66, in __init__
    self._fragment = Fragment.get(fragment, platform=None).prepare()
  File "/home/marcel/projects/fpgastuff/.VE/lib/python3.9/site-packages/nmigen/hdl/ir.py", line 39, in get
    obj = obj.elaborate(platform)
  File "/home/marcel/projects/fpgastuff/.VE/lib/python3.9/site-packages/nmigen/hdl/dsl.py", line 540, in elaborate
    fragment.add_subfragment(Fragment.get(self._named_submodules[name], platform), name)
  File "/home/marcel/projects/fpgastuff/.VE/lib/python3.9/site-packages/nmigen/hdl/ir.py", line 39, in get
    obj = obj.elaborate(platform)
  File "/home/marcel/projects/fpgastuff/a.py", line 10, in elaborate
    with m.FSM(name="space here"):
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/home/marcel/projects/fpgastuff/.VE/lib/python3.9/site-packages/nmigen/hdl/dsl.py", line 361, in FSM
    raise ValueError("FSM name '{}' contains space, which is not allowed".format(name))
ValueError: FSM name 'space here' contains space, which is not allowed
```

Other character, even non visible, e.g. `\x03` seems fine, only a space `\x20` is bad.

### DIFF
--- a/nmigen/sim/pysim.py
+++ b/nmigen/sim/pysim.py
@@ -94,6 +94,8 @@ class _VCDWriter:
                 var_init = signal.reset
 
             for (*var_scope, var_name) in names:
+                if ' ' in var_name:
+                    raise ValueError("Variable '{}' cannot contain a space.".format(var_name))
                 suffix = None
                 while True:
                     try:


whitequark:
Why not replace spaces with underscores instead? This is arguably a gtkwave bug.

hellow554:
> Why not replace spaces with underscores instead
I think that would be a weird feature. So you couldn't use `foo_a` and `foo a` together, because they would clash and the programmer wouldn't know why.

> This is arguably a gtkwave bug.
I haven't found a specification for vcd files, specifically for the id/name. Can you help out here?

awygle:
> I haven't found a specification for vcd files, specifically for the id/name. Can you help out here?

The Extended VCD File Format is defined in IEEE 1364-2001 and later revisions, however it doesn't shed much light on this issue.

I did find [this GTKWave thread](https://sourceforge.net/p/gtkwave/mailman/gtkwave-users/thread/CAHw-xapjz%3D-_6rjwv_c%2BGskCje9agrzd50gQOUA7vDLT%3DrJLhw%40mail.gmail.com/#msg30309039) about escaping spaces as hex or octal, but the context isn't quite clear. I'd suggest you try that and see what results you get.

awygle:
Is it possible to migrate this error to the VCD writer? I'm not sure I agree with whitequark that this is a GTKWave bug exactly, I think it's actually a deficiency in the VCD file format. Throwing from the VCD writer would prevent creating useless VCD files without restricting the names that can be used in nmigen proper.

programmerjake:
I think it would be better to escape spaces when writing vcd files, rather than failing

awygle:
> 
> 
> I think it would be better to escape spaces when writing vcd files, rather than failing

I agree but I don't think there's an unambiguous way to do so. In the "foo a" vs "foo_a" example above, what do you do?

hellow554:
I don't like it when the program does something that I didn't attended to do.

As @awygle and me already saying, you are creating two different names but both will have the same name in the end, because the program does something not very obvious.
Then the next person will spent an hour to figure out what the heck is going on and files the next bug. So raising an error is the lesser evil, especially when it's clear what to do (e.g. replace the space).

I'm more than willing to raise the error in the vcdwriter. Will have (hopefully) time tomorrow to do that.

hellow554:
The ValueError will now be raised in the VCDWriter instead, so it catches all variables that contain a space. 

I think this is the better solution :)

whitequark:
Thanks for the report and the PR! I've implemented a slightly extended version of the check in 66295fa388be1d92f038f390e5aea937b5e5c01b that covers the grammar recognized by GTKWave, and added a test.

Issue #405, opened 2020-06-17
Making an assignment to an input results in unclear assertion error


jeanthom:
```python
from nmigen import *
from nmigen_boards.versa_ecp5 import VersaECP5Platform

class Top(Elaboratable):
    def elaborate(self, platform):
        m = Module()
        input = platform.request("eth_clk125", 0) # This is an input
        m.d.comb += input.i.eq(1)
        return m

if __name__ == "__main__":
    VersaECP5Platform().build(Top())
```

```
Traceback (most recent call last):
  File "bidir.py", line 16, in <module>
    VersaECP5Platform().build(Top())
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/build/plat.py", line 90, in build
    plan = self.prepare(elaboratable, name, **kwargs)
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/build/plat.py", line 162, in prepare
    fragment._propagate_ports(ports=self.iter_ports(), all_undef_as_ports=False)
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 462, in _propagate_ports
    self._prepare_use_def_graph(parent, level, uses, defs, ios, self)
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 436, in _prepare_use_def_graph
    subfrag._prepare_use_def_graph(parent, level, uses, defs, ios, top)
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 428, in _prepare_use_def_graph
    add_defs(value._lhs_signals())
  File "/home/jeanthomas/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 396, in add_defs
    assert defs[sig] is self
AssertionError
```

Ideally the error message would point to the faulty assignment line.

whitequark:
Same underlying problem as #398.

FFY00:
I had a look at this, I don't quite understand the context of how this is happening. @whitequark perhaps you could give me some hints?

whitequark:
> perhaps you could give me some hints?

I wrote that code a while ago, and it's not very good. It'd probably be easier for me to fix it myself if I have to sit down and remember how it works.

Issue #398, opened 2020-06-03
Obscure AssertionError when assigning an io pin to the output of an instance


x44203:
Executing
```python
from nmigen import *
from nmigen.build import *
from nmigen_boards import versa_ecp5_5g as FPGA

class Test(Elaboratable):
    def elaborate(self, platform):
        broken = True
        if broken:
            platform.add_resources([Resource("test", 0, Pins("A6"))])
        else:
            platform.add_resources([Resource("test", 0, Pins("A6", dir="o"))])

        pin = platform.request("test")

        m = Module()
        m.submodules += Instance("OSCG", o_OSC=pin) # It is important that this is an output
        return m

if __name__ == "__main__":
    FPGA.VersaECP55GPlatform().build(Test())
```
fails with
```
python mcve-assert-defs.py                                                                                       1 ✘  21:04:46 
Traceback (most recent call last):
  File "mcve-assert-defs.py", line 20, in <module>
    FPGA.VersaECP55GPlatform().build(Test())
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/build/plat.py", line 90, in build
    plan = self.prepare(elaboratable, name, **kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/build/plat.py", line 162, in prepare
    fragment._propagate_ports(ports=self.iter_ports(), all_undef_as_ports=False)
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 462, in _propagate_ports
    self._prepare_use_def_graph(parent, level, uses, defs, ios, self)
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 436, in _prepare_use_def_graph
    subfrag._prepare_use_def_graph(parent, level, uses, defs, ios, top)
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 428, in _prepare_use_def_graph
    add_defs(value._lhs_signals())
  File "/home/user/.local/lib/python3.8/site-packages/nmigen/hdl/ir.py", line 396, in add_defs
    assert defs[sig] is self
AssertionError
```
setting broken to 'False' makes it compile just fine.
Similar issues are in #320 and #191

Issue #320, opened 2020-02-15
Unclear error message on direction mismatch between Instance in submodule and toplevel usage


rroohhh:
Executing the following snippet
```python
from nmigen import *
from nmigen.back import verilog

class WrappedInstance(Elaboratable):
    def __init__(self):
        self.a = Signal()
        self.b = Signal()

    def elaborate(self, _):
        m = Module()
        m.submodules += Instance("test", o_aout = self.a, o_bout = self.b)
        return m

m = Module()
m.submodules.instance = i = WrappedInstance()
m.d.comb += i.a.eq(i.b)

print(verilog.convert(m))
```

fails with this rather unclear error message:
```
Traceback (most recent call last):
  File "test.py", line 20, in <module>
    print(verilog.convert(m))
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/back/verilog.py", line 76, in convert
    rtlil_text = rtlil.convert(*args, **kwargs)
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/back/rtlil.py", line 1017, in convert
    fragment = ir.Fragment.get(elaboratable, platform).prepare(**kwargs)
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/hdl/ir.py", line 533, in prepare
    fragment._propagate_ports(ports=(), all_undef_as_ports=True)
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/hdl/ir.py", line 462, in _propagate_ports
    self._prepare_use_def_graph(parent, level, uses, defs, ios, self)
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/hdl/ir.py", line 436, in _prepare_use_def_graph
    subfrag._prepare_use_def_graph(parent, level, uses, defs, ios, top)
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/hdl/ir.py", line 428, in _prepare_use_def_graph
    add_defs(value._lhs_signals())
  File "/home/robin/.guix-profile/lib/python3.7/site-packages/nmigen/hdl/ir.py", line 396, in add_defs
    assert defs[sig] is self
AssertionError
```

The problem is, that `i.a` and `i.b` were used by the instance as outputs, but in the toplevel module `a` is assigned to. The error message look similar to #191 but I am not sure this is actually the same bug.

whitequark:
Different issue. I think @awygle also hit this one.

PR #596, opened 2021-02-12
Improve Instance driver-driver conflict error message

awygle:
This PR attempts to address issues #405 , #398 , and #320 by improving error messages for driver-driver conflicts involving Instances.

I took my best stab at a better message, but am totally open to changing it.

I am *almost* sure an Instance conflict is the only case in which this code path is hit (I called out "input pin" because while it's the same issue under the hood it looks different to the user), but please do review closely.

No rush on this - not sure if it makes sense to go into 0.3 or not.

### DIFF
--- a/amaranth/hdl/ir.py
+++ b/amaranth/hdl/ir.py
@@ -394,7 +394,12 @@ class Fragment:
                 if sig not in defs:
                     defs[sig] = self
                 else:
-                    assert defs[sig] is self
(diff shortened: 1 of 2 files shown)


modwizcode (changes requested):
Can you rebase? LGTM after rebase.

whitequark:
Obsoleted by #1104.

PR #598, opened 2021-02-24
Add run_script to build options

slan:
When iterating it's convenient to have the project files updated without running the full synth/impl loop. This PR propagates the top-level `run_script` argument to the execution phase.


whitequark:
If you want to customize the build process in this way I suggest passing `do_build=False` and then using the build plan yourself.

slan:
Makes a lot more sense indeed, thanks.

PR #563, opened 2020-12-16
Unify Xilinx platforms into a single class, support more devices

mwkmwkmwk:
This merges existing code, and also adds support for:

- Virtex, Virtex E (also known as Spartan 2, Spartan 2E)
- Virtex 2, Virtex 2 Pro
- Spartan 3, Spartan 3E (in addition to existing Spartan 3A, Spartan 3A
  DSP support)
- Virtex 4
- Virtex 5
- Virtex 6
- ISE synthesis for Series 7

Fixes #552.

PR #602, opened 2021-03-22
Add synth_design_options override for vivado

slan:
Apparently the only way to pass options for synthesis in Vivado non-project mode is to add the arguments to the synth_design tcl command directly.

This PR adds an override to the xilinx_7series.py.

One use case is to inspect the synthesized design in Vivado with the following overrides in `platform.build`:
```
        synth_design_options="-flatten_hierarchy none -keep_equivalent_registers -no_lc",
        script_after_synth="write_checkpoint -force top.dcp",
```

There might be a more elegant/generic way of doing this for other workflows/platforms (environment variables?).

### DIFF
--- a/nmigen/vendor/xilinx_7series.py
+++ b/nmigen/vendor/xilinx_7series.py
@@ -108,7 +108,7 @@ class Xilinx7SeriesPlatform(TemplatedPlatform):
                 read_xdc {{file|tcl_escape}}
             {% endfor %}
             {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
-            synth_design -top {{name}}
+            synth_design -top {{name}} {{get_override("synth_design_options")|default("# (synth_design_options placeholder)")}}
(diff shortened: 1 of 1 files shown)


nmigen/vendor/xilinx_7series.py
```diff
@@ -108,7 +108,7 @@ def _part(self):
                 read_xdc {{file|tcl_escape}}
             {% endfor %}
             {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
-            synth_design -top {{name}}
+            synth_design -top {{name}} {{get_override("synth_design_options")|default("# (synth_design_options placeholder)")}}
```
whitequark, line 111: Elsewhere we use `opts`, so this should be called `synth_design_opts`.

whitequark (commented):
Concept looks reasonable, but this will not be merged until #563 is, so you might want to hold up updating this PR until that happens. I'll try to do that soon.

slan:
Re-opening with proper branch name so I can merge locally. Sorry for the hassle

PR #607, opened 2021-04-09
lib.fifo: add custom domain to sync FIFOs

slan:
Allow SyncFIFO and SyncFIFOBuffered to use a user-specified clock domain


anuejn:
Why would one want that? Couldn't you use `DomainsRenamer` ?

slan:
I wasn't aware of this option... it felt natural and consistent with e.g. `AsyncFIFO` to set the clock domain explicitly. As a beginner, I developed the habit of having an explicit domain for most of my modules. I found it easier to reason about, but that might be a bad habit for larger-scale projects or libraries (I guess domain name collision).

I'll start to use `DomainRenamer`, feel free to close this PR.

anuejn:
Generally the nMigen convention is that modules that work only in one clock domain use the `sync` domain and are then used in combination with `DomainRenamer`. The `AsyncFIFO` spans multiple clock domains and is therefore different.

whitequark:
What @anuejn said is accurate.

PR #609, opened 2021-04-12
Remove -forceAll from ECP5 and MachXO2 Diamond build scripts

### LINKED ISSUE #604: In ECP5 and MachXO2 backends, is the `-forceAll` option necessary?
While creating `nmigen` designs for MachXO2 that _require_ Diamond, I noticed that the synthesis was significantly slower than compared to the manual shell script that @gatecat created for `prjtrellis`. I talked to them today, and we determined that removing the `-forceAll` option from each step significantly increases the synthesis speed because synthesis steps are not run repeatedly.  By extension, removing `-forceAll` significantly reduces the output log size from running `build_*.{bat, sh} > log.txt`. @gatecat doesn't believe `-forceAll` is necessary at this time, and looking at a test bitstream built without `-forceAll` in EPIC, I'm inclined to agree.

I have created an MVCE with all the files already generated to show the synthesis speeds/log file sizes. Run `python3 mk_build.py > log.txt` to regenerate the output log. `log-noforceall` was generated by removing `-forceAll` manually and running the build script directly. The size difference between the logs is significant; 800 vs 4700 lines (the latter doing completely redundant work?).

## Files
MVCE: [diamond-test.zip](https://github.com/nmigen/nmigen/files/6254298/diamond-test.zip)
Log of the above, with `-forceAll` manually removed from the generated `.bat` (should also apply to `.sh`): [log-noforceall.txt](https://github.com/nmigen/nmigen/files/6254299/log-noforceall.txt)

adamgreig:
Closes #604, cc @cr1901.

### DIFF
--- a/nmigen/vendor/lattice_ecp5.py
+++ b/nmigen/vendor/lattice_ecp5.py
@@ -218,11 +218,11 @@ class LatticeECP5Platform(TemplatedPlatform):
             prj_src add {{name}}.sdc
             {{get_override("script_project")|default("# (script_project placeholder)")}}
             prj_project save
-            prj_run Synthesis -impl impl -forceAll
-            prj_run Translate -impl impl -forceAll
-            prj_run Map -impl impl -forceAll
-            prj_run PAR -impl impl -forceAll
-            prj_run Export -impl impl -forceAll -task Bitgen
+            prj_run Synthesis -impl impl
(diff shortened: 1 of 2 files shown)


whitequark (approved):

whitequark:
Thanks @adamgreig and @cr1901!

PR #619, opened 2021-06-13
nmigen/lib/fifo.py: syncfifo domain option

feede7:
Hi, here there is a propousal tu use a SyncFifo in another domain different than `sync`

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -115,13 +115,16 @@ class SyncFIFO(Elaboratable, FIFOInterface):
     r_attributes="",
     w_attributes="")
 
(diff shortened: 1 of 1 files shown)


rroohhh:
I believe this is usually done using `DomainRenamer`, for example:

```python 
m = Module()
m.submodules += DomainRenamer("other_domain")(SyncFIFO(width = 123, depth = 123))
```

Is this not working for you?

feede7:
I really didn't know it. I'll probe it, but I was expecting pass a domain name instead :sweat_smile:

whitequark:
Yes, @rroohhh is correct: `DomainRenamer` is the intended way to change the domain of a fully synchronous module.

PR #622, opened 2021-07-16
_toolchain: substitute '+' with 'X' in tool_env_var().

jfng:
`tool_env_var("g++")` returns `"G++"`, which is not a valid environment variable name.


whitequark (approved):

whitequark:
Thank you!

PR #624, opened 2021-08-07
sim: Implemented signal notability for use in generating actually useful GTKW files

lethalbit:
This adds a `notable` param to the `Signal` object which is then flattened into a dictionary of groups and signal names in the `pysim` engine for use in writing a GTKW file that contains only what you're interested in, rather than all of the signals in the design, or what has been specified when calling `write_vcd`.

You can specify the group for the signal, or just mark it as notable and it will be included in the `global` group by default when written out to the GTKW file.

A small example would be as follows:

```py
m = Module()
nya = Signal(notable=True)
awoo = Signal()

m.d.sync += [
    nya.eq(~nya),
    awoo.eq(nya),
]

sim = Simulator(m)

sim.add_clock(1 / 12e6, domain = "sync")

with sim.write_vcd("nya.vcd", "nya.gtkw"):
    sim.run()
```

There is also a functioning test case in the `tests/test_sim.py` file, and as of this commit all the tests still pass.

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -925,6 +925,9 @@ class Signal(Value, DUID):
         names). If an ``Enum`` subclass is passed, it is concisely decoded using format string
         ``"{0.name:}/{0.value:}"``, or a number if the signal value is not a member of
         the enumeration.
+    notable : None or bool or str
+        Indicates that this function is notable for storage in the gtkw file for simulation
+        and if so what group to place it in.
 
     Attributes
     ----------
@@ -938,7 +941,7 @@ class Signal(Value, DUID):
     """
 
     def __init__(self, shape=None, *, name=None, reset=0, reset_less=False,
-                 attrs=None, decoder=None, src_loc_at=0):
+                 attrs=None, decoder=None, notable=None, src_loc_at=0):
         super().__init__(src_loc_at=src_loc_at)
 
         if name is not None and not isinstance(name, str):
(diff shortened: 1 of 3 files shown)


whitequark:
The ability to write GTKW files itself is auxiliary; it was added for parity with Migen. I don't think we should add any core language features purely to improve convenience of using GTKW files.

lethalbit:
Yeah that's entirely fair, figured it was worth a shot as manually creating GTKW's and organizing the signals is kinda a pain.

Wasn't that much work anyway so no harm no foul~

anuejn:
I think there could be a nice way to implement that in downstream code if there was a way to attach something like tags to signals and later on collect all signals with a certain tag. This mechanism could also be handy for something like an ila or very ad-hoc debug CSRs.

mithro:
I do like the idea of being able to add arbitrary tags (metadata?) to signals.

whitequark:
I'm not sure how I feel about signal metadata. It's certainly useful, but it's also liable to be the place where all the miscellaneous stuff gets put in, without structure or namespacing, leading to conflicts between different users of metadata.

anuejn:
Wouldn't it be possible to design the API in a way that avoids such problems?
One thing I could imagine would be that all metadata one can attach are objects of downstream classes. Then, one can query all metadata that is instance of a certain class.

whitequark:
> One thing I could imagine would be that all metadata one can attach are objects of downstream classes. Then, one can query all metadata that is instance of a certain class.

That's the approach I picked for nmigen-soc, yeah. It could be applicable here, too; I like the idea!

modwizcode:
That solution probably solves avoiding the mess. But it doesn't solve the potential for conflicting needs for similar metadata. Like a general "export this trace to gtkw" or "add to ILA group x" metadata class might be worth having as a built-in given just how common such features would be generally helpful (maybe not the ILA stuff if we're not going to include anything like that, but auto-add to gtkw/listing of interesting sim signals is probably worth it)

PR #627, opened 2021-08-17
cli: add new show backend which shows a diagram of the generated RTL

hansfbaier:
Hi,
I added a new backend to show diagrams of the RTL represented by the nmigen code.
Please let me know, what you think and what you would like to have changed.

Kind regards,
Hans

### DIFF
--- a/nmigen/back/show.py
+++ b/nmigen/back/show.py
@@ -1,0 +1,38 @@
+from .._toolchain.yosys import *
+from . import rtlil
+
+__all__ = ["convert", "convert_fragment"]
+
+
+def _show_rtlil_text(rtlil_text, *, src_loc_at=0):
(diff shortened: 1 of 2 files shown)


whitequark:
Honestly, I think the entire `cli` module should be deprecated and reworked. It was quickly put together to make the examples work, and it doesn't have much of a _design_.

hansfbaier:
Yes, it's a bit tedious to use.
It would be really cool, if one could just hand it the class name of the Elaboratable,
it would scan it via reflection and instantiate it on its own by default, so that no
extra code would be needed to do the basic tasks like generate/simulate/show.
Anyway, thanks for taking the time.

hansfbaier:
I think I found a more elegant solution.
A commandline tool which already can handle show, and also will handle generation and other common tasks.
https://github.com/hansfbaier/nmigen-tool

PR #628, opened 2021-08-20
sim: Add Observer

alanvgreen:
Adds an Observer interface to the simulator, allowing arbitrary code
to receive values from the simulation.

The implementation is based on the implementation of write_vcd. An
obvious next step would be to make _VCDWriter a subclass of Observer,
then
- rewrite Simulator.write_vcd() as a call to Simulator.oberve()
- remove PySimEngine._vcd_writers and associated code.

This commit is an RFC for https://github.com/nmigen/nmigen/issues/327.
It is not intended to be submitted as-is.

### DIFF
--- a/nmigen/cli.py
+++ b/nmigen/cli.py
@@ -2,10 +2,24 @@ import argparse
 
 from .hdl.ir import Fragment
 from .back import rtlil, cxxrtl, verilog
-from .sim import Simulator
+from .sim import Simulator, Observer
 
 
 __all__ = ["main"]
+
+
+class ObserveAll(Observer):
+    def open(self, timestamp):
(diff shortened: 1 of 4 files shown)


whitequark:
I think this interface isn't appropriate for larger designs (especially ones too large to simulate with the Python simulator and which require CXXRTL). Since a Python function is called for every single value change, it is inevitable that (in cases other than full design waveform capture) most of the calls will be for naught, and a new, more efficient interface would be added later. Even in the case of waveform capture, it's pretty common for full design VCD files to end up being too large, with only a subset being captured.

As such I'm closing this PR. Any implementation improving on it needs to be able to observe only a subset of value changes.

PR #632, opened 2021-09-21
_toolchain: Properly set compiler/linker executables on Gentoo

ajelinski:
The `test_toolchain_cxx.py` tests definitely use compiler and linker set
with `_so_cxx`-suffixed parameters. After this change tests use a proper
executable instead of `c++`.

### DIFF
--- a/nmigen/_toolchain/cxx.py
+++ b/nmigen/_toolchain/cxx.py
@@ -36,6 +36,16 @@ def build_cxx(*, cxx_sources, output_name, include_dirs, macros):
             linker_so=ld_cxxflags,
         )
 
+        # Sometimes CCompiler is modified to have additional executable entries for compiling and
+        # linking CXX shared objects (e.g. on Gentoo). These executables have to be set then.
+        try:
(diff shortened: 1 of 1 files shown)


whitequark:
The tests are failing.

ajelinski:
At first I was extremely confused by the error:
```
ValueError: unknown executable 'compiler_so_cxx' for class UnixCCompiler
```
as it worked and helped in my case.

After doing some research it seems the whole thing is quite Gentoo-specific. Python is built there after applying quite a few patches and adding `{compiler,linker}_so_cxx` to `UnixCCompiler` is in one of these patches: https://github.com/OpenMandrivaAssociation/python/blob/ccf96eae53697ec09e8d03ad46cd8be4ce044548/0005-Improve-distutils-C-support.patch#L243-L246 (that's not the latest version of the patch but these lines haven't changed).

I've modified the commit to try to set them but skip raising the exception if not available.

whitequark (approved):
LGTM, thank you!

PR #633, opened 2021-09-25
Change the character used for constraint hierarchy

H-S-S-11:
This patch was proposed by whitequark originally:
https://freenode.irclog.whitequark.org/nmigen/2021-04-10#29646879

I came across an issue when instantiating a PLL inside a submodule. The generated clock constraint by nMigen would  have `/` characters in the string, which caused errors in ISE. Changing the character to `__` fixed the problem.


whitequark (approved):

whitequark:
Thanks!

Issue #635, opened 2021-09-28
empty signal slice generates verilog which can not be parsed by Quartus


hansfbaier:
```python
from nmigen import *

class QuartusHiccup(Elaboratable):
    def __init__(self):
        self.out = Signal()

    def elaborate(self, platform):
        m = Module()
        data = Signal(1)
        m.d.sync += self.out.eq(data[1:])
        return m
```
This will generate verilog containing:
`  assign \$1  = + ""; `
which is valid verilog, which the quartus parser cannot parse:
```
Error (10170): Verilog HDL syntax error at top.v(19973) near text: """;  expecting "(".
```

hansfbaier:
Also the CSR register multiplexer from nmigen_soc seems to generate similar results,
but I have not been able to reduce that yet. I am not sure whether an empty signal slice has something to do with that.

whitequark:
This is actually an upstream issue; Yosys should not be generating empty strings in this context. I've fixed one instance of this bug in https://github.com/YosysHQ/yosys/pull/1203 (in 2019!), but it turns out there was another one, which I've fixed just now in https://github.com/YosysHQ/yosys/pull/3103.

Your testcase synthesizes fine on Quartus with Yosys 0.10, which is required as of the current `main` branch, so I'm going to close this issue. I'm not sure whether there are other inputs that will generate the RTLIL that causes the pathological Verilog output; please comment on this issue if you do have such a testcase.

hansfbaier:
Wow, that's awesome! Many thanks!

PR #636, opened 2021-09-28
An attempt at fixing #635

hansfbaier:
Hi,
this patch would emit something that works for the simple example in #635.
What do you think?


whitequark:
I don't think it's an appropriate fix; it addresses just one way the issue manifests itself and it adds Quartus-specific hacks to the RTLIL backend.

hansfbaier:
OK no problem, I know that my understanding of nmigen internals is very limited.

PR #638, opened 2021-10-02
Mention a bit the origin of migen and m-labs

doronbehar:
Addresses https://github.com/NixOS/nixpkgs/issues/140015

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # nMigen
 
+> An actively developed version of [`m-labs/nmigen`](https://github.com/m-labs/nmigen/).
+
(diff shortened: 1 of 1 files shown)


whitequark:
What you wrote is not true; the development of nMigen did not start at M-Labs, but rather it began its life as my personal project that was not coordinated with M-Labs in any way. I would appreciate it if you did not make assertions about events you have no knowledge of.

whitequark:
That being said, I don't see the utility of mentioning an _outdated_ repository in the README of an _up-to-date_ one. What you want to avoid is people thinking that `m-labs/nmigen` is the up-to-date source tree, right? So it would be nice to add a mention there.

PR #640, opened 2021-10-09
fix reversed bit order in PriorityEncoder

hansfbaier:
After the change it works as expected:
image
Slicing already has reversed bit numbering in comparison to verilog,
so the reversed function was not necessary.
Or, rather, because the last m.d.comb wins, that did the reversing.

### DIFF
--- a/nmigen/lib/coding.py
+++ b/nmigen/lib/coding.py
@@ -78,7 +78,7 @@ class PriorityEncoder(Elaboratable):
 
     def elaborate(self, platform):
         m = Module()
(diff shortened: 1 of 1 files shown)


hansfbaier:
with reversed(..) it looks like this:
image

whitequark:
This PR breaks the testsuite.

whitequark:
Yes. The invariant of the `PriorityEncoder` is:

    Assert(enc.n | enc.i.bit_select(enc.o, 1))

hansfbaier:
Ok,. Thanks

PR #651, opened 2021-12-13
sim: represent time internally as 1ps units

### LINKED ISSUE #535: Don't represent simulation time as floating point
There's really no excuse for implementing it that way in the first place, and we should just fix it. /cc @mwkmwkmwk

modwizcode:
Actually represent time internally without using floating point numbers. 

This is mostly important for after simulation API changes result in an API that allows user facing code to specify these parameters in exact units as well, however for now we don't wish to have a breaking change before this new API is defined. 

This will make that change far more trivial and has already been tested.

Fixes #535.


amaranth/sim/core.py
```diff
@@ -21,13 +21,13 @@ def __repr__(self):
 
 class Delay(Command):
     def __init__(self, interval=None):
-        self.interval = None if interval is None else float(interval)
+        self.interval = None if interval is None else int(interval * (10 ** 10))
```
whitequark, line 24: This is technically a breaking change, since `interval` is a public field.

whitequark (approved):
LGTM

whitequark:
Thank you!

PR #652, opened 2021-12-13
vendor.xilinx: add workaround for ISE

### LINKED ISSUE #641: Attributes are not properly applied to IBUFDS (and possibly other primitives) in Xilinx platform
https://pastebin.com/qvymcNWi

In the example an `IBUFDS` resource is created and requested, with particular Attributes:
```python
Resource("IBUFDS", 0,
                DiffPairs("48", "46", conn=("gpio", 1), dir="i"),
                Attrs(IOSTANDARD="LVDS_25", IBUF_LOW_PWR="FALSE")
            ),
```
When run the attributes appear in the generated ucf, and an IBUFDS primitive is instantiated in the generated verilog. However when ISE actually runs synthesis, the instance properties do not show these attributes applied:
image
In order for ISE to apply the attributes, they need to be parameters to the IBUFDS which is `Instance`d in xilinx.py.

If the generated verilog is edited where the IBUFDS is Instanced, then the attributes are properly applied:
```verilog
IBUFDS #(
  .IOSTANDARD("LVDS_25"),
  .IBUF_LOW_PWR("FALSE")
) IBUFDS_0_0 (
  .I(IBUFDS_0__p),
  .IB(IBUFDS_0__n),
  .O(IBUFDS_0__i)
);
```
In ISE:
image

Lunaphied:
ISE doesn't support using the constraints file to specify attributes on
IO buffers, this works around that by specifying them also as module
parameters which ISE does support properly.

Fixes #641.

Note: This is a horrible hack in my opinion and is currently untested on ISE (in-progress). Comments welcome.

### DIFF
--- a/amaranth/vendor/xilinx.py
+++ b/amaranth/vendor/xilinx.py
@@ -883,6 +883,8 @@ class XilinxPlatform(TemplatedPlatform):
         i, o, t = self._get_xdr_buffer(m, pin, attrs.get("IOSTANDARD"), i_invert=invert)
         for bit in range(pin.width):
             m.submodules["{}_{}".format(pin.name, bit)] = Instance("IBUF",
+                # Ugly hack to workaround ISE not supporting attributes in constraints
+                **{"p_" + k: v for k,v in attrs.items()},
                 i_I=port.io[bit],
                 o_O=i[bit]
             )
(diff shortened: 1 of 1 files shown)


whitequark:
This PR has been in a draft status in a while and it doesn't seem like it's going anywhere, so I'll close it, and reopen if there's interest in it again.

PR #658, opened 2021-12-13
back.{rtlil,verilog}: deprecate implicit ports.

### LINKED ISSUE #630: Undriven signal ends up as module port
If a signal has no driver, it ends up as a port:

```
from nmigen import *
from nmigen.back import verilog

class Test(Elaboratable):
    def __init__(self):
        self.val_out = Signal()

        m = Module()
        int_sig = Signal(8)
        m.d.comb += self.val_out.eq(int_sig)

        return m

if __name__ == "__main__":
    top = Test()
    with open("test.v", "w") as f:
        f.write(verilog.convert(top))
```

```
module top(int_sig);
  input [7:0] int_sig;
  wire val_out;
  assign val_out = int_sig[0];
endmodule
```

This is pretty confusing. Reporting an error here would be great.

modwizcode:
Fixes #630.

### DIFF
--- a/amaranth/back/rtlil.py
+++ b/amaranth/back/rtlil.py
@@ -1,6 +1,7 @@
 import io
 from collections import OrderedDict
 from contextlib import contextmanager
+import warnings
 
 from .._utils import bits_for, flatten
 from ..hdl import ast, ir, mem, xfrm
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM

PR #659, opened 2021-12-13
Fix 'ports' argument getting lost as it's no longer 'kwargs'

gatecat:
Fixes #658; because `ports` isn't part of `kwargs` any more it ends up going missing and, ironically, an implicitly determined set of ports is used instead.


gatecat:
Looking at the test failure, I think the tests should be modified to specify ports explicitly?

whitequark:
Thanks for the report!

PR #662, opened 2021-12-14
setup: add workaround for pypa/pip#7953.

modwizcode:
This is a horrible hack and I hate it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,10 @@
 from setuptools import setup, find_packages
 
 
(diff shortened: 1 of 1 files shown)


whitequark (approved):
I don't like that this is needed and I wish it wasn't, but after having to previously revert 7fca037f9 due to

  * pypa/pip#6375
  * pypa/pip#6434
  * pypa/pip#6438

I don't want to do it again.

PR #664, opened 2021-12-16
sim.pysim: use "bench" as a top level root for testbench signals.

### LINKED ISSUE #561: pysim: testbench-only signals are not placed at root level on the VCD file
Consider:
```python
from nmigen import Signal, Module
from nmigen.sim import Simulator

m = Module()
s = Signal()
t = Signal()
m.d.sync += s.eq(1)


def process():
    yield t.eq(1)


for engine in ["pysim", "cxxsim"]:
    sim = Simulator(m, engine=engine)
    sim.add_clock(1e-6)
    sim.add_sync_process(process)
    with sim.write_vcd(f"bug17.{engine}.vcd", traces=[s, t]):
        sim.run()
```

On PySim, both `s` and `t` are placed under the `top`module.
On CXXSim, `t` is at the root.

The generated GTKWave document (if any) correctly display the trace, but an already existing manually-generated document would need modification, which is inconvenient.

modwizcode:
Fixes #561.

### DIFF
--- a/amaranth/sim/pysim.py
+++ b/amaranth/sim/pysim.py
@@ -19,7 +19,7 @@ class _NameExtractor:
     def __init__(self):
         self.names = SignalDict()
 
-    def __call__(self, fragment, *, hierarchy=("top",)):
+    def __call__(self, fragment, *, hierarchy=("bench", "top",)):
         def add_signal_name(signal):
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM

whitequark (approved):

PR #677, opened 2022-01-16
Loosen dependency on Jinja2 version

twam:
For me also latest Jinja2 version is working fine and strict requirement of old version is conflicting with other libs using nmigen, e.g. https://github.com/lambdaconcept/lambdasoc/blob/master/setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "importlib_metadata; python_version<'3.8'",  # for __version__ and amaranth._toolchain.yosys
         "importlib_resources; python_version<'3.9'", # for amaranth._toolchain.yosys
         "pyvcd~=0.2.2", # for amaranth.pysim
-        "Jinja2~=2.11", # for amaranth.build
+        "Jinja2>=2.11,<4.0", # for amaranth.build
     ],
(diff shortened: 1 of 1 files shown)


pbsds:
Leaving this here to educate the tradeoff: https://python-poetry.org/docs/faq/#why-are-unbound-version-constraints-a-bad-idea
Bounded versions are however a maintenance burden

setup.py
```diff
@@ -43,7 +43,7 @@ def doc_version():
         "importlib_metadata; python_version<'3.8'",  # for __version__ and amaranth._toolchain.yosys
         "importlib_resources; python_version<'3.9'", # for amaranth._toolchain.yosys
         "pyvcd~=0.2.2", # for amaranth.pysim
-        "Jinja2~=2.11", # for amaranth.build
+        "Jinja2>=2.11", # for amaranth.build
```
whitequark, line 46: How about:
```suggestion
        "Jinja2>=2.11,<4.0", # for amaranth.build
```

whitequark (approved):
LGTM

whitequark:
Thank you!

PR #678, opened 2022-01-27
back.verilog: Fix strip_internal_attrs

alyssarosenzweig:
Fix the strip_internal_attrs parameter to verilog.convert by passing it
down the call stack as intended.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>


whitequark (approved):

whitequark:
Thanks!

Issue #12, opened 2018-12-18
Is there any reason to keep `xxx.submodules`?


nmigen-issue-migration:
<a href="https://github.com/mithro"><img src="https://avatars2.githubusercontent.com/u/21212?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mithro](https://github.com/mithro)**
_Tuesday Dec 18, 2018 at 00:37 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/10_

----

A huge number of issues in the old migen are caused by forgetting to assign something to `xxx.submodules`.

Is there any reason that we can't do something like this?
```python
class Module:
    def __setattr__(self, k, v):
        if isinstance(v, Module):
           self.add_module(k, v)
        object.__setattr__(self, k, v)
```

PR #681, opened 2022-01-31
Add CRC generator

adamgreig:
This PR adds a CRC generator to the stdlib. It supports the full set of Williams/Rocksoft model parameters, which in practice means it should be able to generate and validate pretty much every commonly used CRC directly, and significantly that includes all the ones in the [reveng catalogue](https://reveng.sourceforge.io/crc-catalogue/) and [crcmod predefined list](http://crcmod.sourceforge.net/crcmod.predefined.html).

It's configurable for arbitrary input word widths with a constant single-cycle latency, so can be used to generate a classic bit-at-a-time shift-register CRC, or update wider words in parallel. It doesn't support masking the input words, which might limit its use in very wide masked data paths, though I expect that could be added later in a backwards-compatible way if necessary. I haven't thought about how masking might be implemented.

The generated logic is a product/XOR equation for each bit of the new state which includes bits of the previous state and current input. As far as I'm aware this is essentially the best way to generate an arbitrary parallel CRC in hardware.

Currently it uses a `reset` and `valid` input signal, rather than e.g. relying on users to add `EnableInserter` and `ResetInserter` as appropriate. I don't know what the preferred way to deal with that is in stdlib items. Since `reset` will pretty much always be required it seems easier to have it as an input than requiring `ResetInserter`; I guess some applications may never need to change `valid` if they always have valid new data after resetting.

The test suite uses a variety of standard CRCs to provide coverage of the various input parameters, and tests each at input word widths of 1, 2, 4, 8, 16, 32, and 64 bits.

I originally wrote this code for my work, where we've used it in a variety of applications, but I have permission to release it under the BSD two-clause licence for inclusion in Amaranth.

### DIFF
--- a/amaranth/lib/crc/__init__.py
+++ b/amaranth/lib/crc/__init__.py
@@ -1,0 +1,426 @@
+"""
+Utilities for computing cyclic redundancy checks (CRCs) in software and in
+hardware.
+
+CRCs are specified using the :py:class:`Algorithm` class, which contains
+settings for CRC width, polynomial, initial value, input/output reflection, and
+output XOR. Many commonly used CRC algorithms are available in the
+:py:mod:`~amaranth.lib.crc.catalog` module, while most other CRC designs can be
+accommodated by manually constructing :py:class:`Algorithm`.
+
+Call the :py:class:`Algorithm` with a ``data_width`` to obtain a
+:py:class:`Parameters` class, which fully defines a CRC computation. The
+:py:class:`Parameters` class provides the :py:meth:`~Parameters.compute` method
+to perform software computations, and the :py:meth:`~Parameters.create` method
+to create a hardware CRC module, :py:class:`Processor`.
+
+.. code-block::
+
+    # Create a predefined CRC16-CCITT hardware module, using the default
+    # 8-bit data width (in other words, bytes).
+    from amaranth.lib.crc.catalog import CRC16_CCITT
+    crc = m.submodules.crc = CRC16_CCITT().create()
+
+    # Create a custom CRC algorithm, specify the data width explicitly,
+    # and use it to compute a CRC value in software.
+    from amaranth.lib.crc import Algorithm
+    algo = Algorithm(crc_width=16, polynomial=0x1021, initial_crc=0xffff,
+                     reflect_input=False, reflect_output=False,
+                     xor_output=0x0000)
+    assert algo(data_width=8).compute(b"123456789") == 0x29b1
+"""
+
+from ... import *
+
+__all__ = ["Algorithm", "Parameters", "Processor", "catalog"]
+
+
+class Algorithm:
+    """
+    Settings for a CRC algorithm, excluding data width.
+
+    The parameter set is based on the Williams model from
+    "A Painless Guide to CRC Error Detection Algorithms":
+    http://www.ross.net/crc/download/crc_v3.txt
+
+    For a reference of standard CRC parameter sets, refer to:
+
+    * `reveng`_'s catalogue, which uses an identical parameterisation,
+    * `crcmod`_'s list of predefined functions, but remove the leading '1'
+      from the polynominal, XOR the "Init-value" with "XOR-out" to obtain
+      ``initial_crc``, and where "Reversed" is True, set both ``reflect_input``
+      and ``reflect_output`` to True,
+    * `CRC Zoo`_, which contains only polynomials; use the "explicit +1"
+      form of polynomial but remove the leading '1'.
+
+    .. _reveng: https://reveng.sourceforge.io/crc-catalogue/all.htm
+    .. _crcmod: http://crcmod.sourceforge.net/crcmod.predefined.html
+    .. _CRC Zoo: https://users.ece.cmu.edu/~koopman/crc/
+
+    Many commonly used CRC algorithms are available in the
+    :py:mod:`~amaranth.lib.crc.catalog` module, which includes
+    all entries in the `reveng`_ catalogue.
+
+    To create a :py:class:`Parameters` instance, call the :py:class:`Algorithm`
+    object with the required data width, which defaults to 8 bits.
+
+    Parameters
+    ----------
+    crc_width : int
+        Bit width of CRC word. Also known as "width" in the Williams model.
+    polynomial : int
+        CRC polynomial to use, ``crc_width`` bits long, without the implicit
+        ``x**crc_width`` term. Polynomial is always specified with the highest
+        order terms in the most significant bit positions; use
+        ``reflect_input`` and ``reflect_output`` to perform a least
+        significant bit first computation.
+    initial_crc : int
+        Initial value of CRC register at reset. Most significant bit always
+        corresponds to the highest order term in the CRC register.
+    reflect_input : bool
+        If True, the input data words are bit-reflected, so that they are
+        processed least significant bit first.
+    reflect_output : bool
+        If True, the output CRC is bit-reflected, so the least-significant bit
+        of the output is the highest-order bit of the CRC register.
+        Note that this reflection is performed over the entire CRC register;
+        for transmission you may want to treat the output as a little-endian
+        multi-word value, so for example the reflected 16-bit output 0x4E4C
+        would be transmitted as the two octets 0x4C 0x4E, each transmitted
+        least significant bit first.
+    xor_output : int
+        The output CRC will be the CRC register XOR'd with this value, applied
+        after any output bit-reflection.
+    """
+    def __init__(self, *, crc_width, polynomial, initial_crc, reflect_input,
+                 reflect_output, xor_output):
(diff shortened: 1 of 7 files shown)


josuah:
> Currently it uses a reset and valid input signal, rather than e.g. relying on users to add EnableInserter and ResetInserter as appropriate.

This could be affected/related to https://github.com/amaranth-lang/rfcs/pull/12

whitequark:
#12 is a major rework that will not be relevant in short term (if ever).

adamgreig:
I've updated to rename the `reset` signal to `rst`.

adamgreig:
This PR has been updated to address the latest review comments in https://github.com/amaranth-lang/rfcs/pull/6#issuecomment-1567391041. I've squashed the commits to a single one suitable for merging as I don't think there's any benefit to preserving the commit history.

whitequark:
Do you think you could reach 100% code coverage in this PR? This is something I strive for in the core language and the standard library, although it is not currently the case everywhere.

whitequark:
We also need a changelog line for the RFC and this PR -- could you add them?

whitequark:
Perfect, thanks!

adamgreig:
Rebased and added tests for 100% coverage of the new files.

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,417 @@
...
+    crc : Signal(crc_width), out
+        Registered CRC output value, updated one clock cycle after ``valid``
+        becomes asserted.
+    match_detected : Signal(), out
+        Asserted if the current CRC value indicates a valid codeword has been
+        received.
+    """
+    def __init__(self, parameters):
+        if not isinstance(parameters, Parameters):
+            raise TypeError("Parameters must be of type Parameters")
```
whitequark, line 357: Could you do this the same way we handle runtime type checks elsewhere, please?
```suggestion
            raise TypeError("Algorithm parameters must be of type Parameters, "
                            "not {!r}"
                            .format(parameters))
```

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,417 @@
...
+        self.crc_width = parameters.crc_width
+        self.data_width = parameters.data_width
+        self.polynomial = parameters.polynomial
+        self.initial_crc = Const(parameters.initial_crc, self.crc_width)
+        self.reflect_input = parameters.reflect_input
+        self.reflect_output = parameters.reflect_output
+        self.xor_output = parameters.xor_output
```
whitequark, line 364: I don't think it's right to have public writable undocumented fields with these parameters--we could have a single read-only `parameters` property, or not have any public properties at all. (I have a weak preference for the latter, but no objection to the former.)

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,417 @@
...
+    ----------
+    algorithm : Algorithm
+        CRC algorithm to use. Specifies the CRC width, polynomial,
+        initial value, whether to reflect the input or output words,
+        and any output XOR.
+    data_width : int
+        Bit width of data words.
+    """
+    def __init__(self, algorithm, data_width=8):
+        self.crc_width = algorithm.crc_width
```
whitequark, line 152: Same comments as below for `Processor`.

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,417 @@
...
+    def __init__(self, *, crc_width, polynomial, initial_crc, reflect_input,
+                 reflect_output, xor_output):
+        self.crc_width = int(crc_width)
+        self.polynomial = int(polynomial)
+        self.initial_crc = int(initial_crc)
+        self.reflect_input = bool(reflect_input)
+        self.reflect_output = bool(reflect_output)
+        self.xor_output = int(xor_output)
+
+        assert self.crc_width > 0
```
whitequark, line 104: These should be runtime type/range checks, not assertions. Please follow the same structure for range checks as elsewhere in the codebase.

whitequark (changes requested):
In addition to inline comments, please also add a changelog entry.

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,417 @@
...
+        with m.If(self.valid):
+            for i in range(self.crc_width):
+                bit = 0
+                for j in range(self.crc_width):
+                    if self._matrix_f[j][i]:
+                        bit ^= source[j]
+                for j in range(self.data_width):
+                    if self._matrix_g[j][i]:
+                        bit ^= data_in[j]
+                m.d.sync += crc_reg[i].eq(bit)
```
hannesha, line 403: The inner loops can be removed if the matrices use bit-vectors instead.
```python
m.d.sync += crc_reg.bit_select(i, 1).eq(
	(source & self._matrix_f[i]).xor() ^ (data_in & self._matrix_g[i]).xor()
)
```
This also allows to skip the bit conversion during the matrix generation step.
However this might reduce the readability of the algorithm.

whitequark: Could you explain a bit more what the impact of this chanage is besides readability?

adamgreig: It gives the synthesis a bit more to chew on but I think optimise to the same thing, i.e., previously:

    crc_reg[3] = 0 ^ source[2] ^ source[4] ^ data_in[3] ^ data_in[5]

instead it would be:

    crc_reg[3] = (source[0] & 0) ^ (source[1] ^ 0) & (source[2] ^ 1) ^ (source[3] & 0) ^ (source[4] & 1) ^ (source[5] & 0) ^ (data_in[0] & 0) ^ (data_in[1] & 0) & (data_in[2] & 0) ^ (data_in[3] & 1) ^ (data_in[4] & 0) & (data_in[5] & 1)

I don't remember if `.xor()` existed as a method when I wrote this the first time or I might have used it.

hannesha: I just uploaded my parallel CRC implementation that uses this approach.
It still needs a bit more documentation, but maybe it can help as a reference.

https://gist.github.com/hannesha/3d8adf48fc85013e8c77d42492611864

amaranth/lib/crc/__init__.py
adamgreig: Sure, I've changed them to private attributes for Parameters and Processor. I've left the ones on `Algorithm` unchanged though, is that OK? I feel like an argument could be made for documenting and making them public on Parameters too since it's just the ones from algorithm and the data width, but I don't think anyone's likely to need to use them -- you can just create/modify Algorithm and then the Parameters instance is quite transient on the way to Processor or computation.

amaranth/lib/crc/__init__.py
adamgreig: I don't think there's any need to change the current implementation, right? The results should be the same, we just do the loop over the bits in the CRC generator instead of later on in the stack. I like that the current way is clear that each new bit is only generated from a small subset of the state/input bits, but I think it's marginal either way.

hannesha: Yes the current implementation is fine. Compared to my approach it only generates the required bits without relying on constant folding to produce the same logic.

amaranth/lib/crc/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+        self._polynomial = algorithm.polynomial
+        self._initial_crc = algorithm.initial_crc
+        self._reflect_input = algorithm.reflect_input
+        self._reflect_output = algorithm.reflect_output
+        self._xor_output = algorithm.xor_output
+        self._data_width = int(data_width)
+        if self._data_width <= 0:
+            raise ValueError("data_width must be greater than 0")
+
+    def algorithm(self):
```
whitequark, line 166: I think this can be a `@property`?

whitequark (commented):
Aside from the one comment in the body, could you add a changelog entry, please?

adamgreig:
Thanks for checking! I thought I had added a changelog entry [here](https://github.com/amaranth-lang/amaranth/pull/681/files#diff-13939719fa5d88ab31dfc99a49861551b8e157a0ead47125cbe28a2ab72a122d), should I do something else too?

I'll rebase now to fix the merge conflict.

whitequark:
> Thanks for checking! I thought I had added a changelog entry [here](https://github.com/amaranth-lang/amaranth/pull/681/files#diff-13939719fa5d88ab31dfc99a49861551b8e157a0ead47125cbe28a2ab72a122d), should I do something else too?

Oh sorry, I completely missed that somehow!

whitequark (approved):
LGTM :tada:

whitequark (approved):
LGTM. Thank you for this excellent addition to Amaranth's standard library!

whitequark:
Could you squash the commits, please?

adamgreig:
Sorry, I missed your comment last week. I've pushed a single squashed commit that's rebased on current main.

PR #685, opened 2022-03-29
Add a Pipeline helper to Module

lachlansneff:
This adds two additional methods to `Module`:

```python
def Pipeline(self, stb, name="pipeline", domain="sync"):
    ...
def Stage(self, name=None):
    ...
```

These can to be used to easily create pipelined hardware, like so:

```python
class MulAdd(Elaboratable):
    def __init__(self, width):
        self.a = Signal(width)
        self.b = Signal(width)
        self.c = Signal(width)
        self.stb = Signal()
        
        self.o = Signal(width * 2 + 1)
        self.o_stb = Signal()
        self.width = width

        m = Module()

        with m.Pipeline(self.stb) as pln:
            with m.Stage():
                pln.mul = self.a * self.b
            with m.Stage("ADD_ONE"):
                pln.added_one = pln.mul + 1
            with m.Stage("ADD"):
                m.d.sync += self.o.eq(pln.added_one + self.c)
            
            m.d.comb += self.o_stb.eq(pln.o_stb)

        return m
```

When the `stb` argument to `m.Pipeline()` is strobed, the pipeline starts on that clock cycle and progresses through the stages. Of course, multiple stages can be running at any given time.

For data to get pipelined through each stage, you set the `pipeline.<whatever name>` property with the value you want propagated forward through the stages. You can access any value you've set in a previous stage in any future stage. The signal is set within the clock domain that the pipeline was instantiated with.

The `pipeline.o_stb` property is a signal that is strobed a cycle after the last stage executes (I think that's the right behavior, let me know if not).

A stage can be made invalid by calling the `pipeline.invalid_stage()` method. This is, of course, propagated forward through the stages.

### DIFF
--- a/amaranth/hdl/dsl.py
+++ b/amaranth/hdl/dsl.py
@@ -156,6 +156,60 @@ class FSM:
             self.encoding[name] = len(self.encoding)
         return Operator("==", [self.state, self.encoding[name]], src_loc_at=0)
 
+class _PipelineState:
+    def __init__(self, output_stb, builder_domain, comb_builder_domain):
+        self._in_stage = False
+        self._output_stb = output_stb
+        self._builder_domain = builder_domain
+        self._comb_builder_domain = comb_builder_domain
+        self._signals = {}
+        self._strobes = []
+        self._stages = set()
+
+class Pipeline(_ModuleBuilderProxy):
+    def __init__(self, output_stb, builder_domain, comb_builder_domain):
+        object.__setattr__(self, "_state", _PipelineState(output_stb, builder_domain, comb_builder_domain))
+
+    def _new_stage(self, name):
+        self._state._stages.add(name)
+        new_stb = Signal(name=f"{name}_stb")
+        if len(self._state._strobes) > 0:
+            previous_stb = self._state._strobes[-1]
+            self._state._builder_domain += new_stb.eq(previous_stb)
+
+        self._state._strobes.append(new_stb)
+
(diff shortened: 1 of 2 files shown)


modwizcode:
A few things about this.

First of all, I think that the idea of a built-in pipeline abstraction is a potentially good contribution to the language. I can see why this would be beneficial and think it's worth discussion on how to target this use case. However, I think that this is best discussed in a preliminary RFC so that other community members can discuss their needs and how the design fits those needs. 

Another thing to note is that for a feature like this, while an example is good to provide, tests would also be required. You can see examples of the kinds of tests that exist for the other DSL features in [amaranth/tests/test_hdl_dsl.py](https://github.com/amaranth-lang/amaranth/blob/main/tests/test_hdl_dsl.py).

lachlansneff:
> First of all, I think that the idea of a built-in pipeline abstraction is a potentially good contribution to the language. I can see why this would be beneficial and think it's worth discussion on how to target this use case. However, I think that this is best discussed in a preliminary RFC so that other community members can discuss their needs and how the design fits those needs.

Oh good idea! Didn't realize that RFCs were used for amaranth. I'll go ahead and open an RFC issue.

> Another thing to note is that for a feature like this, while an example is good to provide, tests would also be required. You can see examples of the kinds of tests that exist for the other DSL features in [amaranth/tests/test_hdl_dsl.py](https://github.com/amaranth-lang/amaranth/blob/main/tests/test_hdl_dsl.py).

Absolutely, I'll add tests if people like the design.

PR #687, opened 2022-03-29
build.plat: use tool_env_var() in _toolchain_env_var.

jfng:
This allows `Platform.toolchain` to contain '+' or  '-' characters.

### DIFF
--- a/amaranth/build/plat.py
+++ b/amaranth/build/plat.py
@@ -70,11 +70,11 @@ class Platform(ResourceManager, metaclass=ABCMeta):
 
     @property
     def _deprecated_toolchain_env_var(self):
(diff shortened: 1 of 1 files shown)


whitequark (approved):

whitequark:
Thanks!

adamgreig:
I guess there's no changelog for keeping track of breaking changes, but I thought it was worth mentioning that this PR means some environment variable names need to be changed to all upper-case in 0.3->master, e.g. `AMARANTH_ENV_Diamond` no longer works and now `AMARANTH_ENV_DIAMOND` is required (which in turn doesn't work on 0.3).

whitequark:
@adamgreig Oh yeah I think that wasn't an intentional one. Could you open an issue, please?

PR #689, opened 2022-03-30
Fix Jinja2 dependency

### LINKED ISSUE #668: Simulated project deviates from real device behavior
When I was implementing a slave Quad SPI controller with amaranth I found two cases, one where the simulation indicates that the device will work, but in reality it does not, and another where in simulation it indicates that device will not work, but in reality it works.
I have simplified the project and created a simplified serial controller with a FSM that is able to reproduce the issue. (Example can be found in the attachment along with examples of the two issues)

- **Issue 1**: The simulation indicates that the device will work, but in reality it does not. This is the scenario in the attached example SimpleFSMClient1. In this case the simulation clock is set to be: 
```
m.d.comb += clkSerial.eq(clkSerialExt & ~simpleFSMClient1.serialCSn)
```
that is, the serial clock signal is only generated during the data transmission, when the CSn (negated Chip Select) signal is active. In this case, in reality, the FSM misses one extra clock cycle to come out of the FSM `Data` state into the `Idle` state, thus failing to make it ready for receiving a second transfer. However in simulation despite the needed and missing extra clock cycle the simulation says it will come to `Idle` state at the end of transfer, when `serialCSn` signal goes high/inactive. This simulation result is generated when the signals between the modules are set like so: 
```
simpleFSM.serialCSn = self.serialCSn
```
The result can be seen here: 
sim1IdleStateOnCSnHigh

- **Issue 2**: The simulation indicates that the device will not work, but in reality it does work. This is the scenario in the attached example SimpleFSMClient2. In this case the simulation clock is set to be:
```
m.d.comb += clkSerial.eq((clkSerialExt & ~simpleFSMClient2.serialCSn) | ((simpleFSMClient2.simpleFSM.fsmState != 0) & simpleFSMClient2.serialCSn))
```
that is, the serial clock is generated only when `serialCSn` when signal is low/active and an extra clock cycle is generated by the `serialCSn` signal itself. Here the `simpleFSMClient2.simpleFSM.fsmState != 0` condition is not required in reality, but only for simulation, because serialCSn starts low by default, despite the simulation process setting it high as early as it can. The condition ensures that the extra cycle is only generated when the FSM machine has its current state set to a non `Idle` state.

This scenario was tested with a Xilinx XC7S15 (Spartan-7) and is working properly in the device. The simulation however indicates that the FSM machine will get stuck in the `Data`state, thus failing the second transfer. This simulation result is caused when the following `serialCSn` signal assignment is used:  
```
m.d.comb += simpleFSM.serialCSn.eq(self.serialCSn)
```
as depicted here:
sim2NonIdleStateOnCSnHigh

In summary:
- The design always works in simulation when `serialCSn` is assigned with:
```
simpleFSM.serialCSn = self.serialCSn
```
- The design never works in simulation when `serialCSn`is assigned with:
```
m.d.comb += simpleFSM.serialCSn.eq(self.serialCSn)
```
- In Issue1 the simulation should get stuck in the `Data` state at the end of the first transfer
- In Issue2 the simulation should reach the `Idle` state at the end of the first transfer

[SimpleFSM.txt](https://github.com/amaranth-lang/amaranth/files/7757679/SimpleFSM.txt)
[SimpleFSMClient1.txt](https://github.com/amaranth-lang/amaranth/files/7757680/SimpleFSMClient1.txt)
[SimpleFSMClient2.txt](https://github.com/amaranth-lang/amaranth/files/7757681/SimpleFSMClient2.txt)

### LINKED ISSUE #688: Dependency issue: "ImportError: cannot import name 'soft_unicode' from 'markupsafe'"
I installed amaranth for the first time (version 0.3), and encountered this error:

> ImportError: cannot import name 'soft_unicode' from 'markupsafe'

I found the [same error](https://github.com/aws/aws-sam-cli/issues/3661) which says that it's due to a change in [version 2.1.0 of markupsafe](https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0).

I fixed it locally by changing to an older version with `pip install markupsafe==2.0.1`.

Stack trace:

```
...snip...
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/amaranth/build/__init__.py", line 3, in <module>
    from .plat import *
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/amaranth/build/plat.py", line 6, in <module>
    import jinja2
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/Users/rossmotley/Library/Python/3.8/lib/python/site-packages/markupsafe/__init__.py)
```

modwizcode:
This PR fixes #688 by upgrading the Jinja2 dependency to minimum of major version 3, which does not have the issue with it's markupsafe dependency being  incorrect. In order to do this, update the uses of `@jinja2.contextfunction` to use the non-deprecated `@jinja2.pass_context` name instead, as this name will be removed in Jinja2 versions `>=3.1.0`.

### DIFF
--- a/amaranth/build/plat.py
+++ b/amaranth/build/plat.py
@@ -375,7 +375,7 @@ class TemplatedPlatform(Platform):
             else:
                 return jinja2.Undefined(name=var)
 
-        @jinja2.contextfunction
+        @jinja2.pass_context
         def invoke_tool(context, name):
             env_var = tool_env_var(name)
             if context.parent["syntax"] == "sh":
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "importlib_metadata; python_version<'3.8'",  # for __version__ and amaranth._toolchain.yosys
         "importlib_resources; python_version<'3.9'", # for amaranth._toolchain.yosys
         "pyvcd~=0.2.2", # for amaranth.pysim
-        "Jinja2>=2.11,<4.0", # for amaranth.build
+        "Jinja2~=3.0",  # for amaranth.build
     ],
     extras_require={
         # this version requirement needs to be synchronized with the one in amaranth.back.verilog!


whitequark (approved):
LGTM

whitequark:
Thanks!

Issue #623, opened 2021-07-16
Option to disable debug verilog generation


rroohhh:
The generation of the `debug.v` can take quite some time (`O(10 seconds)` for some designs), so especially for flows where only a `rtlil` file is needed it would be nice to have a option to disable the its generation.

whitequark:
Having seen just how much time it wastes, I agree that this would be a valuable addition.

PR #694, opened 2022-04-05
Add a flag for disabling debug Verilog generation

modwizcode:
Currently `get_override()` does not have a well defined interface and is not well designed for handling things like a flag to
*disable* a feature cleanly. This PR implements `get_override_flag()` to provide this functionality while `get_override()` additionally gains improved type checking for keyword arguments.

This is used to implement #623 which is now possible by providing a "false-like" value `("0", "no", "n", "false", "")` to it (case-insensitive).


amaranth/build/plat.py
```diff
@@ -315,6 +315,45 @@ def toolchain_prepare(self, fragment, name, **kwargs):
...
+                    var_env_value = os.environ[var_env]
+                elif deprecated_var_env in os.environ:
+                    var_env_value = os.environ[deprecated_var_env]
+                return re.sub(r'^\"\"$', "", var_env_value)
+            elif var in kwargs:
+                if expected_type is None or isinstance(kwargs[var], expected_type):
+                    return kwargs[var]
+                else:
+                    raise TypeError("Invalid keyword argument '{}' of type '{}' when type '{}' was expected."
+                                    .format(var, type(kwargs[var]).__name__, expected_type.__name__))
```
whitequark, line 339: Our type errors are usually more like:

    raise TypeError("Override '{}' must be a {}, not {!r}".format(var, expected_type.__name__, kwargs[var])

Aside from the phrasing differences this shows you what actual object you got in there, not just its type, which is convenient.

A minor stylistic issue, but if this was converted to an early raise, then every branch would lexically end with a `return`.

amaranth/build/plat.py
```diff
@@ -315,6 +315,45 @@ def toolchain_prepare(self, fragment, name, **kwargs):
...
+            return value
+
+        def get_override_flag(var):
+            value = _extract_override(var, expected_type=bool)
+            if isinstance(value, str):
+                value = value.lower()
+                if value in ("0", "no", "n", "false", ""):
+                    return False
+                if value in ("1", "yes", "y", "true"):
+                    return True
```
whitequark, line 354: I think unrecognized values shouldn't be silently swallowed here (and converted to `True` by the fallthrough below).

whitequark (changes requested):

amaranth/build/plat.py
modwizcode: Agreed. I didn't search hard enough when looking for example error messages to follow. I also agree with the stylistic change.

amaranth/build/plat.py
```diff
@@ -315,6 +315,50 @@ def toolchain_prepare(self, fragment, name, **kwargs):
...
+                else:
+                    raise ValueError("Override '{}' must be one of "
+                                     "(\"0\", \"n\", \"no\", \"false\", \"\") "
+                                     "or "
+                                     "(\"1\", \"y\", \"yes\", \"true\") not {!r}"
```
whitequark, line 358: Missing comma:
```suggestion
                                     "(\"1\", \"y\", \"yes\", \"true\"), not {!r}"
```

whitequark (changes requested):

whitequark (approved):
LGTM

whitequark (approved):

whitequark:
Thank you!

PR #698, opened 2022-04-06
Cover `amaranth.build.dsl` with documentation

Lunaphied:
Starting to document the `amaranth.build` module and thought the DSL portion was a useful place to start since it's the most frequently encountered.

Due to Git issues I accidentally forced #696 to not exist so here's the fixed PR.


whitequark:
Thank you for this contribution. The build DSL is something that's widely considered flawed and we plan to (hopefully) replace it in short to medium term future, so there aren't current plans to write and update documentation for it.

PR #710, opened 2022-06-30
tracer: add Python 3.11 support.

mwkmwkmwk:


whitequark (approved):
LGTM

mwkmwkmwk:
It hasn't been released yet — I can try adding the beta version, but it'll likely have to be changed later

whitequark:
It can be a non-required check for the beta version.

mwkmwkmwk:
Just adding "3.11-dev" makes CI die on importlib-resources DeprecationWarning in certifi, which is a dependency of codecov.  I fixed it by unsetting `PYTHONWARNINGS` before running codecov (presumably no amaranth code actually gets run by that command, so we have no use for the warnings).

Issue #359, opened 2020-04-17
Python chokes on simulated memories of depth greater than ~3k


sjolsen:
With Python 3.7.3 (shipped with Debian stable) and 3.8.2 (via pyenv), I encountered the following error when simulating a design with a couple of memory banks:

```
  File "/path/to/nmigen/back/pysim.py", line 765, in __call__
    exec(emitter.flush(), exec_locals)
RecursionError: maximum recursion depth exceeded during compilation
```

Here's a minimal reproduction:

```python
import unittest

from nmigen import *
from nmigen.back.pysim import *

class MemTest(unittest.TestCase):

    def test_mem(self):
        m = Module()
        mem = Memory(width=1, depth=0xC00)
        m.submodules.rport = mem.read_port()
        m.submodules.wport = mem.write_port()
        sim = Simulator(m)
        sim.run()
```

Reducing the depth of the memory (width doesn't seem to matter) makes the failure go away. For the time being that workaround works for me, but it would be nice to use the same design parameters for simulation and synthesis.

whitequark:
Yep, this is a known pysim issue that requires reworking the way memories are translated.

whitequark:
Interesting, I thought this would work on 3.7 (https://bugs.python.org/issue31113) but it doesn't.

twam:
Observed the same problem. Can workaround for now by calling `sys.setrecursionlimit(x)`.

jeremyherbert:
Just to add a few notes to this to help with people finding it from google:

On a Core i9 MBP (2019) this makes simulations of memory with a depth of 2**13 (just a lookup table) take about 1 minute if the recursion limit is set high enough. Python version is `3.9.5 (v3.9.5:0a7dcbdb13, May  3 2021, 13:05:53) \n[Clang 12.0.5 (clang-1205.0.22.9)]`.

You can expect to see an exception which looks something like this:

```
Traceback (most recent call last):
  File "/Users/jeremy/Seafile/misc_projects/amaranth_playing/sine_lookup.py", line 96, in <module>
    sim = Simulator(m)
  File "/Users/jeremy/Seafile/misc_projects/amaranth_playing/venv/lib/python3.9/site-packages/amaranth/sim/core.py", line 68, in __init__
    self._engine   = engine(self._fragment)
  File "/Users/jeremy/Seafile/misc_projects/amaranth_playing/venv/lib/python3.9/site-packages/amaranth/sim/pysim.py", line 282, in __init__
[...]

PR #712, opened 2022-09-13
Translate ArrayProxy to pattern matching when supported

Jimx-:
Current the value compiler translates ArrayProxy into if-elif trees which can cause the compiler to crash due to deep recursion (#359). This patch instead translates them into pattern matching when it is supported (Python >= 3.10) to avoid this problem.

### DIFF
--- a/amaranth/sim/_pyrtl.py
+++ b/amaranth/sim/_pyrtl.py
@@ -11,6 +11,7 @@ from ._base import BaseProcess
 
 __all__ = ["PyRTLProcess"]
 
+_USE_PATTERN_MATCHING = (sys.version_info >= (3, 10))
 
 class PyRTLProcess(BaseProcess):
     __slots__ = ("is_comb", "runnable", "passive", "run")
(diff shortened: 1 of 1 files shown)


whitequark:
This is an interesting solution--I'll need to look at it a bit closer.

What's performance like?

Jimx-:
Pattern matching seems to be slightly faster than if-elif for compiling
```python
m = Module()
mem = Memory(width=32, depth=0xb80)
m.submodules.rport = mem.read_port()
m.submodules.wport = mem.write_port()
sim = Simulator(m)
sim.run()
```
Pattern matching:
`python test.py  0.28s user 0.02s system 99% cpu 0.299 total`

if-elif:
`python test.py  0.36s user 0.02s system 99% cpu 0.380 total`

If-elif crashes for depth over 0xc00 with the default recursion limit

Patter matching with depth 0x4000:
`python test.py  1.48s user 0.10s system 99% cpu 1.581 total`

whitequark:
What about runtime performance?

Jimx-:
They are very close

```python
    def process():
        for i in range(depth):
            yield dut.waddr.eq(i)
            yield dut.wdata.eq(i)
            yield dut.we.eq(1)
            yield

    sim.add_sync_process(process)
    start = time.time()
    sim.run()
    end = time.time()
    print(end - start)
```
It took both of them about 1.5s to finish

amaranth/sim/_pyrtl.py
```diff
@@ -224,20 +225,33 @@ def on_Repl(self, value):
         return f"0"
 
     def on_ArrayProxy(self, value):
+        use_pm = sys.version_info >= (3, 10)
```
whitequark, line 228: Can you make this a private module-level constant? Something like `_USE_PATTERN_MATCHING = (sys.version_info >= (3, 10))` after the `__all__` definition.

whitequark (commented):
Looks good to me; I had another solution planned but there's no reason to not merge this in the interim.

whitequark:
Could you rebase on top of `main`, please? The CI should work afterwards.

whitequark (approved):

whitequark:
Thank you!

PR #716, opened 2022-09-20
sim._pyrtl: work around Python's new integer-string convertion limits

Mrmaxmeier:
Hi,

some of my code recently started failing with this exception after a Python version bump:

```
[...]
  File "/usr/lib/python3.10/site-packages/amaranth/hdl/xfrm.py", line 209, in on_statement
    new_stmt = self.on_Assign(stmt)
  File "/usr/lib/python3.10/site-packages/amaranth/sim/_pyrtl.py", line 349, in on_Assign
    gen_rhs = f"({(1 << len(stmt.rhs)) - 1} & {gen_rhs_value})"
ValueError: Exceeds the limit (4300) for integer string conversion
```

Here's a small reproducer:

```python
from amaranth import Module, Signal, sim
dut = Module()
dut.d.sync += Signal(1).eq(Signal(1) << Signal(14))
sim.Simulator(dut)
```

This fails because the generated code uses a large mask value, which raises an ValueError during decimal formatting in Python versions that include a mitigation for CVE-2020-10735. Formatting to hexadecimal instead avoids the algorithmic complexity and is not impacted by the new conversion limits.

I'm not sure if this is worth merging as affected modules are already near the [existing value limits](https://github.com/amaranth-lang/amaranth/blob/db49294cf722e9463666bd06f81a86f930b8707c/amaranth/sim/_pyrtl.py#L75) enforced by pyrtl, so feel free to close/reject this patch ^^

Note: There's an open proposal that changes int's `repr` implementation to hex for large values and would make this patch a bit more self-explanatory (`{v!r}` vs `{v:#x}`): https://github.com/python/cpython/issues/96601

### DIFF
--- a/amaranth/sim/_pyrtl.py
+++ b/amaranth/sim/_pyrtl.py
@@ -125,11 +125,11 @@ class _RHSValueCompiler(_ValueCompiler):
     def on_Operator(self, value):
         def mask(value):
             value_mask = (1 << len(value)) - 1
-            return f"({value_mask} & {self(value)})"
+            return f"({value_mask:#x} & {self(value)})"
 
         def sign(value):
             if value.shape().signed:
-                return f"sign({mask(value)}, {-1 << (len(value) - 1)})"
+                return f"sign({mask(value)}, {-1 << (len(value) - 1):#x})"
             else: # unsigned
                 return mask(value)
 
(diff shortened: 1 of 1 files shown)


whitequark (dismissed):
LGTM, thank you!

whitequark:
Could you rebase on top of `main`, please? The CI should pass afterwards.

whitequark (approved):

whitequark:
Thank you!

Issue #704, opened 2022-04-21
Detect and reject netlists with combinatorial loops


sporniket:
_Originally posted by @whitequark in https://github.com/amaranth-lang/amaranth/issues/703#issuecomment-1104789392_

> In general, if the simulator doesn't converge, it means an issue with the input netlist. I'd say that the appropriate way to handle this is to reject such netlists before they can be simulated, since by the time the simulator is stuck, the original issue (a logic loop of some kind) becomes very hard to find.

&rArr; Questions

* How to detect such netlists : loops, ... ?
* Cases that can't be spotted ?
* Message to report the problem ?

whitequark:
This is not an RFC: no design is presented that can be commented on.

Also, this issue is not exclusive to the simulator: designs with combinatorial feedback loops are unsound and [may be silently miscompiled during synthesis](https://amaranth-lang.org/docs/amaranth/latest/lang.html#lang-comb) as well.

ekliptik:
As far as I can tell, this requires the following, done on the entire design, or on fragments with combinational domains guaranteed to be disconnected:

```
for every combinational assignment s.eq(e):
    for every signal s' in expression e:
        add edge (s', s)
```

We get an oriented graph. If all is good, it is a [forest](https://en.wikipedia.org/wiki/Tree_(graph_theory)#Forest). If there's a comb loop, there's a cycle. So we look for it with DFS from all states, most of these terminating immediately. I don't know where this traversal should happen. I have looked at xfrm.py, we could use some sort of traversal. Maybe it should be implemented in the `elaborate` method of `Module`? Or at the highest level in cli.py before conversion or sim?

ekliptik:
I'm building a prototype using Visitors from xfrm.py and have determined that the `prepare` method in ir.py. Not sure how to get "fully qualified" signal names or IDs yet.

whitequark:
I think this isn't something I'm considering integrating in the next few development cycles, sorry.

ekliptik:
Understood. Is the project structure likely to change until it may become relevant? I'm still probably going to get it done, it's fine by me if it stays in my fork for half a year or forever, I'm doing it for rather academic reasons

ekliptik:
[...]

PR #726, opened 2022-10-31
lib.coding: remove GrayDecoder comb loop for consistency

ekliptik:
While developing a prototype for #704 at 052647445772cfe83767b0c55dd47c196763ac82, I discovered that bits of `Graydecoder.o` are combinationally dependent on each other. When each bit is thought of as an individual signal, this is no problem, but a comb loop detection pass would be slowed down by that assumption. Regardless of if it's within the scope of the language's "no *indirect* comb loops" policy, it is easily rewritable to never have assignments that have the output signal both on the left and righ-hand sides. That is what this PR contains. All tests pass, including the formal check of gray encoding reversibility, proving correctness of this PR.

### DIFF
--- a/amaranth/lib/coding.py
+++ b/amaranth/lib/coding.py
@@ -178,7 +178,8 @@ class GrayDecoder(Elaboratable):
 
     def elaborate(self, platform):
         m = Module()
-        m.d.comb += self.o[-1].eq(self.i[-1])
-        for i in reversed(range(self.width - 1)):
-            m.d.comb += self.o[i].eq(self.o[i + 1] ^ self.i[i])
+        rhs = Const(0)
+        for i in reversed(range(self.width)):
+            rhs = rhs ^ self.i[i]
+            m.d.comb += self.o[i].eq(rhs)
         return m


whitequark:
> I discovered that bits of `Graydecoder.o` are combinationally dependent on each other.

So... I'm on the fence as to whether this should be allowed. I know, for a fact, that people write such code and expect it to work. (CXXRTL, for example, has to handle this, and so does Amaranth currently.)

I do not know whether it should be allowed in Amaranth. Currently, a literal reading of the language guide [prohibits](https://amaranth-lang.org/docs/amaranth/latest/lang.html#combinatorial-evaluation) such apparent combinatorial loops, so your PR follows its text; however, that section doesn't really talk about word-level combinatorial loops that disappear once you look at the bit-level. Since there was no way to detect or forbid this pattern, I'm sure that eventually forbidding it will break enough real-world code to be a concern.

ekliptik:
If you - or users - believe Amaranth has staying power in the HDL space, it might be worth it to break small parts of people's designs as soon as possible, rather than overconstraining what the implementation can do with users' designs. I mean, Amaranth is on version 0.x, that seems like a signal to users that things can shift around.

As far as this PR goes, I realized that the `Value`s built are bigger with this change. Could be a perf issue if people decide to, um, gray decode thousands of bits, or something, not a big deal, but it's just more IR propagating until yosys does your AIG strashing or something, which doesn't even happen with pysim, if I understand it correctly.

whitequark (approved):

PR #729, opened 2022-11-13
Don't uppercase toolchain in toolchain_env_var. Fixes #728.

### LINKED ISSUE #728: Toolchain environment override environment variable names now must be upper-case
In 0.3 and before #687, environment variable names for overrides such as `AMARANTH_ENV_Diamond` used the same case as the toolchain name (e.g., `Diamond`). After #687 the use of `tool_env_var()` inadvertently(?) causes all names to require upper-case, e.g. `AMARANTH_ENV_DIAMOND`. The upper-case name doesn't work in 0.3, and the previous mixed-case name no longer works.

adamgreig:
See #728 and #687.

Having done this, though, I wonder if having the toolchain env vars be uppercase would be better in general, even if we maintain an old/deprecated mixed-case name? All the tool overrides are already in uppercase, and using all-uppercase environment variables is a widely used convention. We could extend `deprecated_toolchain_env_var` to cover both `NMIGEN_ENV_Diamond` and `AMARANTH_ENV_Diamond`, with `AMARANTH_ENV_DIAMOND` being the nominal name, used first if multiple are set.

I don't especially mind, happy to just take this as-is (or with whatever changes you want), but maybe it would be more consistent to deliberately change to all-uppercase for 0.4.


whitequark:
> We could extend `deprecated_toolchain_env_var` to cover both `NMIGEN_ENV_Diamond` and `AMARANTH_ENV_Diamond`, with `AMARANTH_ENV_DIAMOND` being the nominal name, used first if multiple are set.

LGTM

adamgreig:
I've updated to accept the old case for the name alongside the NMIGEN version and the new uppercase version. If multiple variables are specified then the deprecated versions are executed first, followed by the new uppercase version, which means it should be the one to actually take effect - which is the same behaviour as before when both NMIGEN and AMARANTH versions were specified.

I've only tested the ECP5 toolchain as I don't have tools installed for the others.

whitequark (approved):

whitequark:
Thank you very much!

PR #731, opened 2022-12-01
fix xilinx toolchain scripts for f4pga.

bl0x:
The symbiflow tools in f4pga have changed the required arguments.
This PR fixes the xilinx toolchain (at least for Arty A35).


whitequark (approved):
LGTM

PR #734, opened 2022-12-20
Initial support for gowin vendor via proprietary and apicula toolchain.

bl0x:
Supported devices: gw1n 9c / tang nano 9k

### DIFF
--- a/amaranth/vendor/gowin.py
+++ b/amaranth/vendor/gowin.py
@@ -1,0 +1,594 @@
+from abc import abstractproperty
+from fractions import Fraction
+import re
+
+from ..hdl import *
+from ..lib.cdc import ResetSynchronizer
+from ..build import *
+
+# Acknowledgments:
+#   Parts of this file originate from https://github.com/tcjie/Gowin
+
+
+__all__ = ["GowinPlatform"]
+
+
+class GowinPlatform(TemplatedPlatform):
+    """
+    .. rubric:: Apicula toolchain
+
+    Required tools:
+        * ``yosys``
+        * ``nextpnr-gowin``
+        * ``gowin_pack``
+
+    The environment is populated by running the script specified in the environment variable
+    ``AMARANTH_ENV_APICULA``, if present.
+
+    Build products:
+        * ``{{name}}.fs``: binary bitstream.
+
+    .. rubric:: Gowin toolchain
+
+    Required tools:
+        * ``gw_sh``
+
+    The environment is populated by running the script specified in the environment variable
+    ``AMARANTH_ENV_GOWIN``, if present.
+
+    Build products:
+        * ``{{name}}.fs``: binary bitstream.
+    """
+
+    toolchain = None # selected when creating platform
+
+    part   = abstractproperty()
+    family = abstractproperty()
+
+    def parse_part(self):
+        # These regular expressions match all >900 parts of Gowin device_info.csv
+        reg_series    = r"(GW[12]{1}[AN]{1}[EFNRSZ]{0,3})-"
+        reg_voltage   = r"(ZV|EV|LV|LX|UV|UX)"
+        reg_size      = r"(1|2|4|9|18|55)"
+        reg_subseries = r"(?:(B|C|S|X|P5)?)"
+        reg_package   = r"((?:PG|UG|EQ|LQ|MG|M|QN|CS|FN)(?:\d+)(?:P?)(?:A|E|M|CF|C|D|G|H|F|S|T|U|X)?)"
+        reg_speed     = r"((?:C\d{1}/I\d{1})|ES|A\d{1}|I\d{1})"
+
+        match = re.match(reg_series+reg_voltage+reg_size+reg_subseries+reg_package+reg_speed+"$",
+                         self.part)
+        if not match:
+            raise ValueError("Supplied part name is invalid")
+
+        self.series    = match.group(1)
+        self.voltage   = match.group(2)
+        self.size      = match.group(3)
+        self.subseries = match.group(4) or ""
+        self.package   = match.group(5)
+        self.speed     = match.group(6)
+
+        match = re.match(reg_series+reg_size+reg_subseries+"$", self.family)
+        if not match:
+            raise ValueError("Supplied device family name is invalid")
+
+        self.series_f    = match.group(1)
+        self.size_f      = match.group(2)
+        self.subseries_f = match.group(3) or ""
+
+        # subseries_f is usually more reliable than subseries.
+
+        if self.series != self.series_f:
+            raise ValueError("Series extracted from supplied part name does not match "
+                             "supplied family series")
+        if self.size != self.size_f:
+            raise ValueError("Size extracted from supplied part name does not match "
+                             "supplied family size")
+
+    # _chipdb_device is tied to available chipdb-*.bin files of nextpnr-gowin
+    @property
+    def _chipdb_device(self):
+        # GW1NR series does not have its own chipdb file, but works with GW1N
+        if self.series == "GW1NR":
+            return "GW1N-{}{}".format(self.size, self.subseries_f)
+        return self.family
+
+    _dev_osc_mapping = {
+       "GW1N-1" : "OSCH",
+       "GW1N-1P5" : "OSCO",
+       "GW1N-1P5B" : "OSCO",
+       "GW1N-1S" : "OSCH",
+       "GW1N-2" : "OSCO",
+       "GW1N-2B" : "OSCO",
+       "GW1N-4" : "OSC",
+       "GW1N-4B" : "OSC",
+       "GW1N-9" : "OSC",
+       "GW1N-9C" : "OSC",
+       "GW1NR-1" : "OSCH",
+       "GW1NR-2" : "OSCO",
+       "GW1NR-2B" : "OSCO",
+       "GW1NR-4" : "OSC",
+       "GW1NR-4B" : "OSC",
+       "GW1NR-9" : "OSC",
+       "GW1NR-9C" : "OSC",
+       "GW1NRF-4B" : "OSC",
+       "GW1NS-2" : "OSCF",
+       "GW1NS-2C" : "OSCF",
+       "GW1NS-4" : "OSCZ",
+       "GW1NS-4C" : "OSCZ",
+       "GW1NSE-2C" : "OSCF",
+       "GW1NSER-4C" : "OSCZ",
+       "GW1NSR-2" : "OSCF",
+       "GW1NSR-2C" : "OSCF",
+       "GW1NSR-4" : "OSCZ",
+       "GW1NSR-4C" : "OSCZ",
+       "GW1NZ-1" : "OSCZ",
+       "GW1NZ-1C" : "OSCZ",
+       "GW2A-18" : "OSC",
+       "GW2A-18C" : "OSC",
+       "GW2A-55" : "OSC",
+       "GW2A-55C" : "OSC",
+       "GW2AN-18X" : "OSCW",
+       "GW2AN-55C" : "OSC",
+       "GW2AN-9X" : "OSCW",
+       "GW2ANR-18C" : "OSC",
+       "GW2AR-18" : "OSC",
+       "GW2AR-18C" : "OSC"
+    }
+
+    @property
+    def _osc_type(self):
+        if self.family in self._dev_osc_mapping:
+            return self._dev_osc_mapping[self.family]
+        raise NotImplementedError("Device family {} does not have an assigned oscillator type"
+                                  .format(self.family))
+
+    @property
+    def _osc_base_freq(self):
+        osc = self._osc_type
+        if osc == "OSC":
+            if self.speed == 4 and self.subseries_f in ("B", "D"):
+                return 210_000_000
+            else:
+                return 250_000_000
+        elif osc in ("OSCZ", "OSCO"):
+            if self.series == "GW1NSR" and self.speed == "C7/I6":
+                return 260_000_000
+            else:
+                return 250_000_000
+        elif osc in ("OSCF", "OSCH"):
+            return 240_000_000
+        elif osc == "OSCW":
+            return 200_000_000
+        else:
+            assert False
+
+    @property
+    def _osc_div(self):
+        div_min  = 2
+        div_max  = 128
+        div_step = 2
+        div_frac = Fraction(self._osc_base_freq, self.osc_frequency)
+
+        if div_frac.denominator != 1 or div_frac not in range(div_min, div_max, div_step):
+            raise ValueError(
+                "On-chip oscillator frequency (platform.osc_frequency) must be chosen such that "
+                "the oscillator divider, calculated as ({}/{}), is an integer between {} and {} in "
+                "steps of {}"
+                .format(div_frac.numerator, div_frac.denominator, div_min, div_max, div_step))
+
+        return div_frac.numerator
+
+    # Common templates
+
+    _common_file_templates = {
+        "{{name}}.cst": r"""
+            // {{autogenerated}}
+            {% for port_name, pin_name, attrs in platform.iter_port_constraints_bits() -%}
+                IO_LOC "{{port_name}}" {{pin_name}};
+                {% for attr_name, attr_value in attrs.items() -%}
+                    IO_PORT "{{port_name}}" {{attr_name}}={{attr_value}};
+                {% endfor %}
+            {% endfor %}
+        """,
+    }
+
+    # Apicula templates
+
+    _apicula_required_tools = [
+        "yosys",
+        "nextpnr-gowin",
+        "gowin_pack"
+    ]
+    _apicula_file_templates = {
+        **TemplatedPlatform.build_script_templates,
+        **_common_file_templates,
+        "{{name}}.il": r"""
+            # {{autogenerated}}
+            {{emit_rtlil()}}
+        """,
+        "{{name}}.debug.v": r"""
+            /* {{autogenerated}} */
+            {{emit_debug_verilog()}}
+        """,
+        "{{name}}.ys": r"""
+            # {{autogenerated}}
+            {% for file in platform.iter_files(".v") -%}
+                read_verilog {{get_override("read_verilog_opts")|options}} {{file}}
+            {% endfor %}
+            {% for file in platform.iter_files(".sv") -%}
+                read_verilog -sv {{get_override("read_verilog_opts")|options}} {{file}}
+            {% endfor %}
+            {% for file in platform.iter_files(".il") -%}
+                read_ilang {{file}}
+            {% endfor %}
+            read_ilang {{name}}.il
+            delete w:$verilog_initial_trigger
+            {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
+            synth_gowin {{get_override("synth_opts")|options}} -top {{name}} -json {{name}}.syn.json
+            {{get_override("script_after_synth")|default("# (script_after_synth placeholder)")}}
+        """,
+    }
+    _apicula_command_templates = [
+        r"""
+        {{invoke_tool("yosys")}}
+            {{quiet("-q")}}
+            {{get_override("yosys_opts")|options}}
+            -l {{name}}.rpt
+            {{name}}.ys
+        """,
+        r"""
+        {{invoke_tool("nextpnr-gowin")}}
+            {{quiet("--quiet")}}
+            {{get_override("nextpnr_opts")|options}}
+            --log {{name}}.tim
+            --device {{platform.part}}
+            --family {{platform._chipdb_device}}
+            --json {{name}}.syn.json
+            --cst {{name}}.cst
+            --write {{name}}.pnr.json
+        """,
+        r"""
+        {{invoke_tool("gowin_pack")}}
+            -d {{platform._chipdb_device}}
+            -o {{name}}.fs
+            {{get_override("gowin_pack_opts")|options}}
+            {{name}}.pnr.json
+        """
+    ]
+
+    # Vendor toolchain templates
+
+    _gowin_required_tools = ["gw_sh"]
+    _gowin_file_templates = {
+        **TemplatedPlatform.build_script_templates,
+        **_common_file_templates,
+        "{{name}}.v": r"""
+            /* {{autogenerated}} */
+            {{emit_verilog()}}
+        """,
+        "{{name}}.tcl": r"""
+            # {{autogenerated}}
+            {% for file in platform.iter_files(".v",".sv",".vhd",".vhdl") -%}
+                add_file {{file}}
+            {% endfor %}
+            add_file -type verilog {{name}}.v
+            add_file -type cst {{name}}.cst
+            add_file -type sdc {{name}}.sdc
+            set_device -name {{platform.family}} {{platform.part}}
+            set_option -verilog_std v2001 -print_all_synthesis_warning 1 -show_all_warn 1
+            {{get_override("add_options")|default("# (add_options placeholder)")}}
+            run all
+            file delete -force {{name}}.fs
+            file copy -force impl/pnr/project.fs {{name}}.fs
+        """,
+        "{{name}}.sdc": r"""
+        // {{autogenerated}}
+        {% for net_signal,port_signal,frequency in platform.iter_clock_constraints() -%}
+            create_clock -name {{port_signal.name|tcl_escape}} -period {{1000000000/frequency}}  [get_ports {{port_signal.name|tcl_escape}}]
+        {% endfor %}
+        {{get_override("add_constraints")|default("# (add_constraints placeholder)")}}
+        """,
+    }
+    _gowin_command_templates = [
+        r"""
+        {{invoke_tool("gw_sh")}}
+            {{name}}.tcl
+        """
+    ]
+
+    def __init__(self, *, toolchain="Apicula"):
+        super().__init__()
+
+        assert toolchain in ("Apicula", "Gowin")
+        self.toolchain = toolchain
+
+        self.parse_part()
+
+    @property
+    def required_tools(self):
+        if self.toolchain == "Apicula":
+            return self._apicula_required_tools
+        elif self.toolchain == "Gowin":
+            return self._gowin_required_tools
+        assert False
+
+    @property
+    def file_templates(self):
+        if self.toolchain == "Apicula":
+            return self._apicula_file_templates
+        elif self.toolchain == "Gowin":
+            return self._gowin_file_templates
+        assert False
+
+    @property
+    def command_templates(self):
+        if self.toolchain == "Apicula":
+            return self._apicula_command_templates
+        elif self.toolchain == "Gowin":
+            return self._gowin_command_templates
+        assert False
+
+    def add_clock_constraint(self, clock, frequency):
+        super().add_clock_constraint(clock, frequency)
+        clock.attrs["keep"] = "true"
+
+    @property
+    def default_clk_constraint(self):
+        if self.default_clk == "OSC":
+            if not hasattr(self, "osc_frequency"):
+                raise AttributeError(
+                    "Using the on-chip oscillator as the default clock source requires "
+                    "the platform.osc_frequency attribute to be set")
+            return Clock(self.osc_frequency)
+
+        # Use the defined Clock resource.
+        return super().default_clk_constraint
+
+    def create_missing_domain(self, name):
+        if name == "sync" and self.default_clk is not None:
+            m = Module()
+
+            if self.default_clk == "OSC":
+                clk_i = Signal()
+                if self._osc_type == "OSCZ":
+                    m.submodules += Instance(self._osc_type,
+                                             p_FREQ_DIV=self._osc_div,
+                                             i_OSCEN=Const(1),
+                                             o_OSCOUT=clk_i)
+                elif self._osc_type == "OSCO":
+                    # TODO: Make use of regulator configurable
+                    m.submodules += Instance(self._osc_type,
+                                             p_REGULATOR_EN=Const(1),
+                                             p_FREQ_DIV=self._osc_div,
+                                             i_OSCEN=Const(1),
+                                             o_OSCOUT=clk_i)
+                elif self._osc_type == "OSCF":
+                    m.submodules += Instance(self._osc_type,
+                                             p_FREQ_DIV=self._osc_div,
+                                             o_OSCOUT30M=None,
+                                             o_OSCOUT=clk_i)
+                else:
+                    m.submodules += Instance(self._osc_type,
+                                             p_FREQ_DIV=self._osc_div,
+                                             o_OSCOUT=clk_i)
+
+            else:
+                clk_i = self.request(self.default_clk).i
+
+            if self.default_rst is not None:
+                rst_i = self.request(self.default_rst).i
+            else:
+                rst_i = Const(0)
+
+            m.submodules.reset_sync = ResetSynchronizer(rst_i, domain="sync")
+            m.domains += ClockDomain("sync")
+            m.d.comb += ClockSignal("sync").eq(clk_i)
+
+            return m
+
+    def _get_xdr_buffer(self, m, pin, i_invert=False, o_invert=False):
+
+        def get_ireg(clk,d,q):
+            for bit in range(len(q)):
+                m.submodules += Instance("DFF",
+                    i_CLK=clk,
+                    i_D=d[bit],
+                    o_Q=q[bit],
+                )
+
+        def get_oreg(clk,d,q):
+            for bit in range(len(q)):
+                m.submodules += Instance("DFF",
+                    i_CLK=clk,
+                    i_D=d[bit],
+                    o_Q=q[bit]
+                )
+
+        def get_iddr(clk,d,q0,q1):
+            for bit in range(len(d)):
+                m.submodules += Instance("IDDR",
+                    i_CLK=clk,
+                    i_D=d[bit],
+                    o_Q0=q0[bit],
+                    o_Q1=q1[bit]
+                )
+
+        def get_oddr(clk,d0,d1,tx,q0,q1):
+            for bit in range(len(q0)):
+                m.submodules += Instance("ODDR",
+                    p_TXCLK_POL=0, # default -> Q1 changes on posedge of CLK
+                    i_CLK=clk,
+                    i_D0=d0[bit],
+                    i_D1=d1[bit],
+                    i_TX=tx[bit],
+                    o_Q0=q0[bit],
+                    o_Q1=q1[bit]
+                )
+
+        def get_ineg(y, invert):
+            if invert:
+                a = Signal.like(y, name_suffix="_n")
+                m.d.comb += y.eq(~a)
+                return a
+            else:
+                return y
+
+        def get_oneg(a, invert):
+            if invert:
+                y = Signal.like(a, name_suffix="_n")
+                m.d.comb += y.eq(~a)
+                return y
+            else:
+                return a
+
+
+        if "i" in pin.dir:
+            if pin.xdr < 2:
+                pin_i = get_ineg(pin.i, i_invert)
+            elif pin.xdr == 2:
+                pin_i0 = get_ineg(pin.i0, i_invert)
+                pin_i1 = get_ineg(pin.i1, i_invert)
+        if "o" in pin.dir:
+            if pin.xdr < 2:
+                pin_o = get_oneg(pin.o, o_invert)
+            elif pin.xdr == 2:
+                pin_o0 = get_oneg(pin.o0, o_invert)
+                pin_o1 = get_oneg(pin.o1, o_invert)
+
+        i = o = t = None
+
+        if "i" in pin.dir:
+            i = Signal(pin.width, name="{}_xdr_i".format(pin.name))
+        if "o" in pin.dir:
+            o = Signal(pin.width, name="{}_xdr_o".format(pin.name))
+        if pin.dir in ("oe", "io"):
+            t = Signal(1,         name="{}_xdr_t".format(pin.name))
+
+        if pin.xdr == 0:
+            if "i" in pin.dir:
+                i = pin_i
+            if "o" in pin.dir:
+                o = pin_o
+            if pin.dir in ("oe", "io"):
+                t = ~pin.oe
+        elif pin.xdr == 1:
+            if "i" in pin.dir:
+                get_ireg(pin.i_clk, i, pin_i)
+            if "o" in pin.dir:
+                get_oreg(pin.o_clk, pin_o, o)
+            if pin.dir in ("oe", "io"):
+                get_oreg(pin.o_clk, ~pin.oe, t)
+        elif pin.xdr == 2:
+            if "i" in pin.dir:
+                get_iddr(pin.i_clk, i, pin_i0, pin_i1)
+            if pin.dir in ("o", "oe", "io"):
+                get_oddr(pin.o_clk, pin_o0, pin_o1, ~pin_oe, o, t)
+        else:
+            assert False
+
+        return (i, o, t)
+
+    def get_input(self, pin, port, attrs, invert):
+        self._check_feature("single-ended input", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, i_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name, bit)] = Instance("IBUF",
+                i_I=port.io[bit],
+                o_O=i[bit]
+            )
+        return m
+
+    def get_output(self, pin, port, attrs, invert):
+        self._check_feature("single-ended output", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, port.io, o_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name, bit)] = Instance("OBUF",
+                i_I=o[bit],
+                o_O=port.io[bit]
+            )
+        return m
+
+    def get_tristate(self, pin, port, attrs, invert):
+        self._check_feature("single-ended tristate", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, o_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name, bit)] = Instance("TBUF",
+                i_OEN=t,
+                i_I=o[bit],
+                o_O=port.io[bit]
+            )
+        return m
+
+    def get_input_output(self, pin, port, attrs, invert):
+        self._check_feature("single-ended input/output", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, i_invert=invert, o_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name, bit)] = Instance("IOBUF",
+                i_OEN=t,
+                i_I=o[bit],
+                o_O=i[bit],
+                io_IO=port.io[bit]
+            )
+        return m
+
+    def get_diff_input(self, pin, port, attrs, invert):
+        self._check_feature("differential input", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, i_invert=invert)
+        for bit in range(pin.wodth):
+            m.submodules["{}_{}".format(pin.name,bit)] = Instance("TLVDS_IBUF",
+                i_I=port.p[bit],
+                i_IB=port.n[bit],
+                o_O=i[bit]
+            )
+        return m
+
+    def get_diff_output(self, pin, port, attrs, invert):
+        self._check_feature("differential output", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, o_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name,bit)] = Instance("TLVDS_OBUF",
+                i_I=o[bit],
+                o_O=port.p[bit],
+                o_OB=port.n[bit],
+            )
+        return m
+
+    def get_diff_tristate(self, pin, port, attrs, invert):
+        self._check_feature("differential tristate", pin, attrs,
+                            valid_xdrs=(0, 1, 2), valid_attrs=True)
+        m = Module()
+        i, o, t = self._get_xdr_buffer(m, pin, o_invert=invert)
+        for bit in range(pin.width):
+            m.submodules["{}_{}".format(pin.name,bit)] = Instance("TLVDS_TBUF",
+                i_OEN=t,
+                i_I=o[bit],
+                o_O=port.p[bit],
+                o_OB=port.n[bit]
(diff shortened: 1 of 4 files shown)


amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,217 @@
...
+
+    toolchain = None # selected when creating platform
+
+    device  = abstractproperty()
+    family  = abstractproperty()
+    voltage = abstractproperty()
+    size    = abstractproperty()
+    package = abstractproperty()
+    speed   = abstractproperty()
+    board   = abstractproperty()
```
whitequark, line 35: `board`?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,217 @@
...
+            -d {{platform.device}}
+            -o {{name}}.fs
+            {{name}}.json
+        """
+    ]
+
+    def __init__(self, *, toolchain="Apicula"):
+        super().__init__()
+
+        assert toolchain in ("Apicula")
```
whitequark, line 122: This needs a trailing comma, or you are using string `in` operator.

whitequark (changes requested):

amaranth/vendor/gowin.py
bl0x: oops. Has nothing to do there. Fixed.

amaranth/vendor/gowin.py
bl0x: Yes, pet peeve. Fixed.

whitequark:
Does the Apicula toolchain not support IOB registers at all currently?

bl0x:
Unclear. The blinky did not want them. I can check how far it goes. @pepijndevos certainly knows more. Also, the fuzzer code mentions IOBUF.

whitequark:
I will merge the PR once all of the features of the Apicula toolchain are supported, similar to Lattice, Xilinx, etc ones.

pepijndevos:
Exciting stuff. Looks like the IOB question has been resolved right?

bl0x:
@pepijndevos thanks! What else is supported in apicula? Is there a list of supported and tested primitives somewhere?

pepijndevos:
Tested you say? Bwaha
There is https://github.com/YosysHQ/yosys/wiki/FPGA-family-feature-matrix
Recent development is WIP pll support but not on 9k yet.

bl0x:
Ok, so DDR is something I could at least add and see if it compiles.

bl0x:
I've rebased and added support for xdr>0 from https://github.com/tcjie/Gowin repo.
The next step could be to also add support for the vendor toolchain that is also included in tcjie's repo.
@tcjie: would you want to contribute?

bl0x:
@pepijndevos if you have time, could you perhaps have a look here? Is anything still missing? Testing is still pending for the differential IO, since I have limited means to do this.

tcjie:
@bl0x The answer is yes.

pepijndevos:
Eh, not too familiar with amaranth but I guess it basically looks fine. What do you need for testing differential IO? IIRC we basically run this example as a sanity check https://github.com/YosysHQ/apicula/blob/master/examples/blinky-oddr.v

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,360 @@
...
+            return self._apicula_file_templates
+        assert False
+
+    @property
+    def command_templates(self):
+        if self.toolchain == "Apicula":
+            return self._apicula_command_templates
+        assert False
+
+    @property
```
whitequark, line 150: Does Gowin have on-chip oscillators?

pepijndevos: Yes

bl0x: The latest commits have added support for on-chip oscillators.

bl0x:
@tcjie I've used some code from your repository https://github.com/tcjie/Gowin to add support for the proprietary gowin toolchain.

bl0x:
Support for on chip oscillators is now also in.

josuah:
Here is the result with `pip install .` on bl0x's fork of amaranth (this branch) then running:

```
$ python amaranth_boards/tang_nano_9k.py
Jtag frequency : requested 6.00MHz   -> real 6.00MHz
Parse file Parse /tmp/amaranth_j3bogl4d_top.fs:
Done
DONE
Jtag frequency : requested 2.50MHz   -> real 2.00MHz
erase SRAM Done
Flash SRAM: [==================================================] 100.00%
Done
$
```

VID_20230417_165957

josuah:
Exact same result as above with the Gowin toolchain, after doing this:
```
# ln -sf /opt/Gowin_V1.9.8.09_linux/IDE/bin/gw_sh /usr/local/bin
```

`test.py`
```python
from amaranth_boards.tang_nano_9k import *
from amaranth_boards.test.blinky import *
TangNano9kPlatform(toolchain="Gowin").build(Blinky(), do_program=True)
```

```
$ python test.py
*** GOWIN Tcl Command Line Console  ***
add new file: "top.v"
add new file: "top.cst"
add new file: "top.sdc"
current device: GW1NR-9C  GW1NR-LV9QN88PC6/I5
GowinSynthesis start
Running parser ...
Analyzing Verilog file '/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v'
Compiling module 'top'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":160)
WARN  (EX3780) : Using initial value of '$auto$verilog_backend.cc:2083:dump_module$2' since it is never assigned("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":161)
Compiling module 'cd_sync'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":4)
Compiling module 'reset_sync'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":119)
WARN  (EX3780) : Using initial value of '$auto$verilog_backend.cc:2083:dump_module$1' since it is never assigned("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":120)
Compiling module 'pin_button_0'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":19)
Compiling module 'pin_button_1'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":30)
Compiling module 'pin_clk27_0'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":41)
Compiling module 'pin_led_0'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":53)
Compiling module 'pin_led_1'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":64)
Compiling module 'pin_led_2'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":75)
Compiling module 'pin_led_3'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":86)
Compiling module 'pin_led_4'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":97)
Compiling module 'pin_led_5'("/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.v":108)
NOTE  (EX0101) : Current top module is "top"
[5%] Running netlist conversion ...
Running device independent optimization ...
[10%] Optimizing Phase 0 completed
[15%] Optimizing Phase 1 completed
[25%] Optimizing Phase 2 completed
Running inference ...
[30%] Inferring Phase 0 completed
[40%] Inferring Phase 1 completed
[50%] Inferring Phase 2 completed
[55%] Inferring Phase 3 completed
Running technical mapping ...
[60%] Tech-Mapping Phase 0 completed
[65%] Tech-Mapping Phase 1 completed
[75%] Tech-Mapping Phase 2 completed
[80%] Tech-Mapping Phase 3 completed
[90%] Tech-Mapping Phase 4 completed
[95%] Generate netlist file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/gwsynthesis/project.vg" completed
[100%] Generate report file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/gwsynthesis/project_syn.rpt.html" completed
GowinSynthesis finish
Reading netlist file: "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/gwsynthesis/project.vg"
Parsing netlist file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/gwsynthesis/project.vg" completed
Processing netlist completed
Reading constraint file: "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/top.cst"
Physical Constraint parsed completed
Running placement......
[10%] Placement Phase 0 completed
[20%] Placement Phase 1 completed
[30%] Placement Phase 2 completed
[50%] Placement Phase 3 completed
Running routing......
[60%] Routing Phase 0 completed
[70%] Routing Phase 1 completed
[80%] Routing Phase 2 completed
[90%] Routing Phase 3 completed
Running timing analysis......
[95%] Timing analysis completed
Placement and routing completed
Bitstream generation in progress......
Bitstream generation completed
Running power analysis......
[100%] Power analysis completed
Generate file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/pnr/project.power.html" completed
Generate file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/pnr/project.pin.html" completed
Generate file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/pnr/project.rpt.html" completed
Generate file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/pnr/project.rpt.txt" completed
Generate file "/home/josuah/amaranth-lang/amaranth-boards-bl0x/build/impl/pnr/project.tr.html" completed
Mon Apr 17 17:14:17 2023

Jtag frequency : requested 6.00MHz   -> real 6.00MHz
Parse file Parse /tmp/amaranth_5xhumiex_top.fs:
Done
DONE
Jtag frequency : requested 2.50MHz   -> real 2.00MHz
erase SRAM Done
Flash SRAM: [==================================================] 100.00%
Done
$ 
```

tcjie:
@bl0x  that's very useful

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+            --device {{platform._part}}
+            --family {{platform.device}}
+            --json {{name}}.json
+            --freq 27
+            --cst {{name}}.cst
+            --write {{name}}.json
+        """,
+        r"""
+        {{invoke_tool("gowin_pack")}}
+            {{verbose("-v")}}
```
mpasternacki, line 206: `gowin_pack` doesn't seem to accept `-v`:

```
+ gowin_pack -v -d GW1NZ-1 -o top.fs top.json
usage: gowin_pack [-h] -d DEVICE [-o OUTPUT] [-c] [-s CST] [--allow_pinless_io] [--jtag_as_gpio] [--sspi_as_gpio] [--mspi_as_gpio] [--ready_as_gpio] [--done_as_gpio] [--reconfign_as_gpio] [--png PNG] netlist
gowin_pack: error: unrecognized arguments: -v
```

Other than this minor issue, this code works with my Tang Nano 1k – thank you!

whitequark (commented):
Thank you for your contribution! I do not see any technical issues with this PR. I'm going to test it locally and once everything passes, merge it.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+        """,
+        r"""
+        {{invoke_tool("nextpnr-gowin")}}
+            {{quiet("--quiet")}}
+            {{get_override("nextpnr_opts")|options}}
+            --log {{name}}.tim
+            --device {{platform._part}}
+            --family {{platform.device}}
+            --json {{name}}.json
+            --freq 27
```
whitequark, line 200: This argument shouldn't be there--I don't know what way is there to set constrains for the Gowin toolchain but it is what should be used.

@psentee Unfortunately replacing it with the frequency of the internal oscillator, while convenient for the development boards, isn't the right solution here eiher.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+    """
+    .. rubric:: Apicula toolchain
+
+    Required tools:
+        * ``yosys``
+        * ``nextpnr-gowin``
+        * ``gowin_pack``
+
+    The environment is populated by running the script specified in the environment variable
+    ``AMARANTH_ENV_Apicula``, if present.
```
whitequark, line 24: In `main` branch this should be `AMARANTH_ENV_APICULA`.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+    Build products:
+        * ``{{name}}.fs``: binary bitstream.
+
+    .. rubric:: Gowin toolchain
+
+    Required tools:
+        * ``gw_sh``
+
+    The environment is populated by running the script specified in the environment variable
+    ``AMARANTH_ENV_Gowin``, if present.
```
whitequark, line 35: In `main` branch this should be `AMARANTH_ENV_GOWIN`.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+
+    ### Apicula templates
+
+    _apicula_required_tools = ["yosys", "nextpnr-gowin", "gowin_pack"]
+
+    _apicula_file_templates = {
+        **TemplatedPlatform.build_script_templates,
+        **_common_file_templates,
```
whitequark, line 160: That's not right—you need either `.il`+`.debug.v` or just `.v`; otherwise you are wasting time by doing double work (and this can quickly add up).

Also, I don't think nextpnr uses the SDC file at all.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+            run all
+        """,
+    }
+
+    _gowin_command_templates = [
+        r"""
+        {{invoke_tool("gw_sh")}}
+            {{name}}.tcl
+        """,
+        "cp impl/pnr/project.fs {{name}}.fs"
```
whitequark, line 245: This will not work on Windows. Does Gowin ship Windows versions of the tools?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+            return self._gowin_command_templates
+        assert False
+
+    def add_clock_constraint(self, clock, frequency):
+        super().add_clock_constraint(clock, frequency)
+        clock.attrs["keep"] = "true"
+
+    @property
+    def default_clk_constraint(self):
+        if self.default_clk == "sys_clk0":
```
whitequark, line 290: Why `sys_clk0`?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+                osc = self._oscillator
+                if (not isinstance(self.osc_div, int)
+                    or self.osc_div < props["div_min"]
+                    or self.osc_div > props["div_max"]
+                    or (self.osc_div % props["div_step"]) != 0):
+                    raise ValueError("OSC divider (osc_div) must be an integer between {} and {} in steps of {}, not {!r}".format(
+                        props["div_min"], props["div_max"], props["div_step"], self.osc_div))
+
+                clk_i = Signal()
+                if osc is "OSCZ":
```
whitequark, line 309: It's not correct to compare Python strings using `is`.
```suggestion
                if osc == "OSCZ":
```

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+                    m.submodules += Instance(osc,
+                                             p_FREQ_DIV=self.osc_div,
+                                             i_OSCEN=Const(1),
+                                             o_OSCOUT=clk_i)
+                elif osc is "OSCO":
```
whitequark, line 314:
```suggestion
                elif osc == "OSCO":
```

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+                    # TODO: Make use of regulator configurable
+                    m.submodules += Instance(osc,
+                                             p_REGULATOR_EN=Const(1),
+                                             p_FREQ_DIV=self.osc_div,
+                                             i_OSCEN=Const(1),
+                                             o_OSCOUT=clk_i)
+                elif osc is "OSCF":
```
whitequark, line 321:
```suggestion
                elif osc == "OSCF":
```

whitequark (changes requested):
On second look I was a bit too enthusiastic earlier, sorry about that!

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,549 @@
...
+                clk_i = self.request(self.default_clk).i
+
+            if self.default_rst is not None:
+                rst_i = self.request(self.default_rst).i
+            else:
+                rst_i = Const(0)
+
+            m.domains += ClockDomain("sync")
+            m.d.comb += ClockSignal("sync").eq(clk_i)
+            m.submodules.reset_sync = ResetSynchronizer(rst_i, domain="sync")
```
whitequark, line 341: After reviewing the GW1N datasheet I think this code should be done like Lattice ECP5, using GSR/SGSR.

amaranth/vendor/gowin.py
bl0x: I removed it and nextpnr-gowin ran fine without it. As I understand it, it allows to set the minimum frequency for any clock domain. Once one has more than one clock domain this parameter becomes useless anyhow. Manually checking the timing report is needed and advised in any case.

amaranth/vendor/gowin.py
bl0x: Should the toolchain name be uppercase? Or just uppercase in the _toolchain_env_var?

amaranth/vendor/gowin.py
bl0x: Is there a default way to distinguish between the `.il`+`.debug.v` or `.v` path? Can I run the toolchain in debug mode?

bl0x: I've removed `.debug.v` generation for now, and don't generate `.sdc` for nextpnr.

amaranth/vendor/gowin.py
whitequark: Ideally, nextpnr-gowin would read clock constraints from a file. However, it looks like the Gowin port specifically has no way to do this: https://github.com/YosysHQ/nextpnr/issues/1191

amaranth/vendor/gowin.py
bl0x: Yes, gowin EDA also exists for Windows. Does a template exist already for such a copy operation?
I'll try to put that copy operation into the tcl file that is generated. That ought to be platform independent.

amaranth/vendor/gowin.py
whitequark: Only in the environment variable name. I.e. only the docs need to be changed.

amaranth/vendor/gowin.py
whitequark: Apicula would use `.il`+`.debug.v`, Gowin would use `.v`.

amaranth/vendor/gowin.py
bl0x: Not sure anymore, I think I got that from the quicklogic / quickfeather bits of amaranth-lang and amaranth-boards. Do you have a better suggestion?

amaranth/vendor/gowin.py
bl0x: Fixed

amaranth/vendor/gowin.py
bl0x: Fixed

amaranth/vendor/gowin.py
bl0x: Fixed.

amaranth/vendor/gowin.py
```diff
@@ -221,6 +221,7 @@ def _device_name(self):
             set_option -verilog_std v2001 -print_all_synthesis_warning 1 -show_all_warn 1
             {{get_override("add_options")|default("# (add_options placeholder)")}}
             run all
+            file copy -force impl/pnr/project.fs {{name}}.fs
```
whitequark, line 224: Thanks, this is a really clever workaround!

amaranth/vendor/gowin.py
bl0x: No idea what that is about, but can find out. Need a break though.

whitequark: Thanks! I'm very happy with the progress we're making.

amaranth/vendor/gowin.py
bl0x: Thanks. Let's have TCL take care of this business.

amaranth/vendor/gowin.py
bl0x: Now it is all caps.

amaranth/vendor/gowin.py
bl0x: Also fixed.

amaranth/vendor/gowin.py
bl0x: There is only asynchronous `GSR` on all Gowin chips. No `SGSR`. I'll use something along the lines of what is done for EPC5. The ECP5 `FD1S3AX` should be equivalent to Gowin `DFF`, right?

amaranth/vendor/gowin.py
bl0x: Ugh, project.fs is created without w permission, so can't easily be overwritten by a subsequent copy. Setting file attributes is not OS independent in TCL (throws 'bad option' if using invalid options for the current OS). Trying to instead delete destination before copy.

amaranth/vendor/gowin.py
bl0x: [constraints.md](https://github.com/YosysHQ/nextpnr/blob/master/docs/constraints.md) of nextpnr describes how to use the `--pre-pack` option for adding individual clock constraints.

bl0x: @whitequark you are right. There is no `--pre-pack` option for nextpnr-gowin. Oops.

amaranth/vendor/gowin.py
whitequark: Oh, if we don't have `SGSR` then I think the current implementation is fine. It would only be because of `SGSR`.

amaranth/vendor/gowin.py
whitequark: No, a template does not exist. I was thinking it could be introduced, but your Tcl solution seems good!

amaranth/vendor/gowin.py
whitequark: `--pre-pack` only exists if nextpnr was built with boost-python, which is not always the case, so we can't rely on it. YoWASP for example lacks it.

amaranth/vendor/gowin.py
whitequark: Yes, deleting before copying, or moving instead of copying, should both work.

amaranth/vendor/gowin.py
whitequark: Thinking about it, I wonder if the `GSR` implemenation actually works with nextpnr-gowin. I remember there was some additional work to make it happen for ECP5...

amaranth/vendor/gowin.py
whitequark: `OSC` or `OSCx`

amaranth/vendor/gowin.py
bl0x: Ok, got it. Fixing this now.

amaranth/vendor/gowin.py
bl0x: Ok, fixed.

amaranth/vendor/gowin.py
bl0x: There were no complaints at compile time, but I did not test with hardware. Now back to using previous implementation.

amaranth/vendor/gowin.py
bl0x: Can we leave it like this for now?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+    @property
+    def _oscillator(self):
+        oscillators = []
+        for osc,devs in self.osc_dev_mapping.items():
+            if self.device in devs:
+                oscillators.append(osc)
+        # should this be unique?
+        if len(oscillators) != 1:
+            assert False
+        return oscillators[0]
```
whitequark, line 121: I think there's no especially good reason to do this dict inversion every time at runtime. Can you put the inverted dict in the sources?

whitequark (changes requested):

amaranth/vendor/gowin.py
whitequark: As in, without `--freq`? Yeah.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+            --json {{name}}.json
+            --cst {{name}}.cst
+            --write {{name}}.json
+        """,
+        r"""
+        {{invoke_tool("gowin_pack")}}
+            -d {{platform.device}}
+            -o {{name}}.fs
+            {{get_override("gowin_pack_opts")|options}}
+            {{name}}.json
```
whitequark, line 197: I see that gowin_pack has a `--cst` option--do you know what's that about?

josuah: As far as I remember, "cst" in GOWIN is for the constraint file, like the "pcf" (port configuration file?) files of other platforms.

From source: https://github.com/YosysHQ/apicula/blob/master/apycula/gowin_pack.py#L997
For instance: https://github.com/YosysHQ/apicula/blob/033d1c6bc8cc93d5a24f9ba935fb690b023149a2/examples/honeycomb.cst#L4

whitequark: Yes, I'm aware. I'm asking what function it has in gowin_pack, considering that it is already used for nextpnr-gowin.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+                    o_q1 = q1[bit]
+                )
+
+        def get_oddr(clk,d0,d1,q):
+            for bit in range(len(1)):
+                m.submodules += Instance("ODDR",
+                    i_clk = clk,
+                    i_d0 = d0[bit],
+                    i_d1 = d1[bit],
+                    o_q0 = q[bit],
```
whitequark, line 376:
```suggestion
                    o_q = q[bit],
```

whitequark:
Video of a color bar test running on Tang Nano (not 9K, one without a suffix): 

https://github.com/amaranth-lang/amaranth/assets/54771/c8132c72-0e2c-4ad2-b9bf-9ec1bab81b5e

amaranth/vendor/gowin.py
josuah: My bad for stating the obvious.

It is generated out rather than read in... maybe just for debug?

> Create .cst (constraints) file from gowin_pack and gowin_unpack
> Add the .cst file generation during packing
> Make the .cst building optional
> -- https://github.com/YosysHQ/apicula/commit/a327700441dc367d895e98c0b4d2a7c6fae35024

Nothing about `cst` [before it](https://github.com/YosysHQ/apicula/blob/f0162d0d46d153a5436e904cf50d4be6b3ddc97e/apycula/gowin_pack.py), which suggests it is not required to be there, but convenient to have?

whitequark: Oh, it _generates_ the constraint file. I'm at a loss of why, but this answers it. Thank you!

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+    ]
+
+    def __init__(self, *, toolchain="Apicula"):
+        super().__init__()
+
+        assert toolchain in ("Apicula", "Gowin")
+        self.toolchain = toolchain
+
+    @property
+    def _toolchain_env_var(self):
```
whitequark, line 254: I think you don't need to override this at all, since the default implementation is enough.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+    package   = abstractproperty()
+    speed     = abstractproperty()
+    osc_div   = abstractproperty()
+
+    @property
+    def _part(self):
+            return "{}-{}{}{}{}".format(self.series, self.voltage,
+            self.size, self.package, self.speed)
+
+    osc_dev_mapping = {
```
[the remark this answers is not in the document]
bl0x: Done.

amaranth/vendor/gowin.py
bl0x: Yes. Done.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+            for bit in range(len(d)):
+                m.submodules += Instance("IDDR",
+                    i_clk = clk,
+                    i_d = d[bit],
+                    o_q0 = q0[bit],
+                    o_q1 = q1[bit]
+                )
+
+        def get_oddr(clk,d0,d1,q):
+            for bit in range(len(1)):
```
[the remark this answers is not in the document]
bl0x: Fixed.

amaranth/vendor/gowin.py
bl0x: Fixed. Also for the others.

amaranth/vendor/gowin.py
bl0x: Yes, looks like it. Thanks. I've removed it.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,554 @@
...
+                    o_Q1 = q1[bit]
+                )
+
+        def get_oddr(clk,d0,d1,q):
+            for bit in range(len(q)):
+                m.submodules += Instance("ODDR",
+                    i_CLK = clk,
+                    i_D0 = d0[bit],
+                    i_D1 = d1[bit],
+                    o_Q0 = q[bit],
```
whitequark, line 382: I think that this isn't quite the right implementation. On Gowin, `ODDR` has two flops inside: one for the data input, and one for the tristate input. I think you have to put both `o` and `oe` through the `ODDR` primitive, and then connect `Q1` to the `t` output.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,548 @@
...
+    toolchain = None # selected when creating platform
+
+    device    = abstractproperty()
+    series    = abstractproperty()
+    subseries = abstractproperty()
+    voltage   = abstractproperty()
+    size      = abstractproperty()
+    package   = abstractproperty()
+    speed     = abstractproperty()
+    osc_div   = abstractproperty()
```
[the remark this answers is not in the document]
bl0x: Done.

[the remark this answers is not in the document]
bl0x: Oh, should it be called `osc_frequency` instead of `osc_freq`? And is there a convention on the unit of `osc_freq*` ? It is convenient to use MHz as unit and allow floats. One might also argue that only integers should be allowed and use Hz as base unit.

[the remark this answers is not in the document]
whitequark: It should be called `osc_frequency` and the unit must be Hz. Take a look at `LatticeMachXO2Or3LPlatform`, which can be used as a template here.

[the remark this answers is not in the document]
whitequark: Also, I think a comment on this line might've went missing. It shouldn't be an `abstractproperty` because this means a design that doesn't use the built-in oscillator at all cannot be instantiated (I'm actually hitting this issue). Rather, you can remove the property, and have `AttributeError` propagate when a clock constraint is requested on a design without `osc_frequency` set.

amaranth/vendor/gowin.py
bl0x: Ah, yes! I checked the manual and found the description in section 4.3.1 of [UG289E](http://cdn.gowinsemi.com.cn/UG289E.pdf). Is this mainly to cope with the associated delay in the DDR path so that signals align properly at the input of the io/tristate buffer?

whitequark: Yep

bl0x: The manual claims that Q1 can also be left floating.

whitequark: Yes, if you don't need tristates. But we do have a case where you have a DDR tristate, don't we?

bl0x: Yes, that is when `pin.xdr == 2` and `pin.dir in ("oe", "io")`.

bl0x: Since the ODDR has two flops already, one doesn't need an extra one, and one condition for all of `"o", "oe" and "io"` should be sufficient?

whitequark: Yes, that was essentially my point here.

amaranth/vendor/gowin.py
```diff
@@ -124,17 +123,19 @@ def _osc_props(self):
...
-        if ((not isinstance(self.osc_freq, int))
-            and (not isinstance(self.osc_freq, float))):
-                raise ValueError("osc_freq must be an int or float")
+        if not hasattr(self, "osc_frequency"):
+            raise AttributeError('default_clk "OSC" requires attribute "osc_frequency" to be set.')
 
-        # osc_freq is given in MHz
-        div = int(props["freq"] / (self.osc_freq * 1e6))
+        if not isinstance(self.osc_frequency, int):
+            raise ValueError("osc_frequency must be an int.")
```
whitequark, line 130: This would be a `TypeError`, but in any case we don't really need to enforce it here.

bl0x: Ok.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,565 @@
...
+       "GW2AN-55C" : "OSC",
+       "GW2AN-9X" : "OSCW",
+       "GW2ANR-18C" : "OSC",
+       "GW2AR-18" : "OSC",
+       "GW2AR-18C" : "OSC"
+    }
+
+    # Each on-chip oscillator has unique properties
+    @property
+    def _osc_props(self):
```
whitequark, line 103: Could you simplify this a bit please? Namely:
- extract a `_osc_base_freq` property that only contains the conditional branching on the oscillator type;
- replace `_osc_props` with `_osc_div` that now only computes the divisor;
- and in this way eliminate the use of dicts to pass semi-structured data around.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,565 @@
...
+            props["freq"] = 250000000
+        elif osc in ["OSCF", "OSCH"]:
+            props["freq"] = 240000000
+        elif osc == "OSCW":
+            props["freq"] = 200000000
+        else:
+            assert False
+
+        if not hasattr(self, "osc_frequency"):
+            raise AttributeError('default_clk "OSC" requires attribute "osc_frequency" to be set.')
```
whitequark, line 127: In Amaranth we do not end exception messages with a dot.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,565 @@
...
+
+        # osc_frequency is given in Hz
+        div = props["freq"] // self.osc_frequency
+        props["div"] = div
+
+        if (div < props["div_min"] or div > props["div_max"]
+            or (div % props["div_step"]) != 0):
+                raise ValueError("OSC frequency (osc_frequency) must be chosen such that the oscillator divider is an integer between {} and {} in steps of {}. The calculated value of {!r} is not valid. The undivided oscillator frequency for this device is {} Hz".format(
```
whitequark, line 135: Could you split this line to fit a 120 character max length?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,565 @@
...
+
+        return props
+
+
+    @property
+    def _oscillator(self):
+        if self.device in self._dev_osc_mapping:
+            return self._dev_osc_mapping[self.device]
+        else:
+            raise ValueError("Device {} not yet in _dev_osc_mapping".format(
```
whitequark, line 147: Please use `NotImplementedError` for this. (Also you don't need an `else:`)

whitequark (changes requested):
Very happy how this is coming together! I think these are some of the last changes to make.

Could you also add a changelog entry, please? You can copy the format from some earlier entries introducing other platforms.

amaranth/vendor/gowin.py
bl0x: Fixed.

amaranth/vendor/gowin.py
bl0x: Fixed.

amaranth/vendor/gowin.py
bl0x: Ok.

amaranth/vendor/gowin.py
bl0x: The compulsive `else:` is a remainder from my Haskell days, I think.

amaranth/vendor/gowin.py
bl0x: Done. The dict came up because I thought that other chips might have other constraints on the divider. Turned out to not be the case.

amaranth/vendor/gowin.py
whitequark: Oh yeah, I can see that. I use `assert False` a lot as a holdover from OCaml.

bl0x:
> Could you also add a changelog entry, please? You can copy the format from some earlier entries introducing other platforms.

Is there an order for the items in the changelog? Should I append/prepend, fit somewhere

whitequark:
@bl0x I've done some minor changes (mainly whitespace fixes and improvemen of diagnosics) that I pushed into your PR, could you take a look please? Other than that I have two comments, which I'll add inline.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,556 @@
...
+    The environment is populated by running the script specified in the environment variable
+    ``AMARANTH_ENV_GOWIN``, if present.
+
+    Build products:
+        * ``{{name}}.fs``: binary bitstream.
+    """
+
+    toolchain = None # selected when creating platform
+
+    device    = abstractproperty()
```
whitequark, line 44: Setting all of these properties individually is quite confusing: (from Tang Nano)
```
    # GW1N-LV1QN48C6/I5
    device      = "GW1N-1"
    series      = "GW1N"
    size        = "1"
    subseries   = ""
    voltage     = "LV"
    package     = "QN48"
    speed       = "C6/I5"
```

Can we do better here?

Also, and this is related, I'm not entirely sure why we have both `_device_name` and `_part`; the naming makes it especially difficult to understand what's going on.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,556 @@
...
+        if self.device in self._dev_osc_mapping:
+            return self._dev_osc_mapping[self.device]
+        raise NotImplementedError("Device {} does not have an assigned oscillator type"
+                                  .format(self.device))
+
+    @property
+    def _osc_base_freq(self):
+        osc = self._osc_type
+        if osc == "OSC":
+            if "-4" in self.device:
```
whitequark, line 115: Related to the issue raised above but can we do better than `"-4" in self.device`?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,556 @@
...
+                return 210_000_000
```
whitequark, line 116: Am I reading it correctly that all oscillators are in hundreds of MHz range?

whitequark:
> Is there an order for the items in the changelog? Should I append/prepend, fit somewhere in the middle?

We should have a policy for this but don't... I've added it somewhere it looks right >_>

amaranth/vendor/gowin.py
bl0x: The internal ones, yes. And then you are supposed to divide down from there.

whitequark: OK, just making sure.

amaranth/vendor/gowin.py
bl0x: One could list the devices explicitly as stated in [UG286E](http://cdn.gowinsemi.com.cn/UG286E.pdf)

amaranth/vendor/gowin.py
bl0x: Their part/family naming scheme is bonkers. Also the option names of the tools are not quite straightforward. 

We could just have the long part name as the sole property and then try to derive everything that's necessary inside of GowinPlatform.

`_part` and `_device_name` are different concatenations of the part name constituents and needed by the different tools in exactly this format.

amaranth/vendor/gowin.py
whitequark: I guess if we're looking for `xxx-4` then `self.device.endswith("-4")` would be much more reasonable than matching anywhere in the string.

amaranth/vendor/gowin.py
whitequark: > We could just have the long part name as the sole property and then try to derive everything that's necessary inside of GowinPlatform.

This is what `XilinxPlatform` does. If you're up for that I think that's ideal, but considering how bonkers it is I don't feel like I can _insist_ on this.

> `_part` and `_device_name` are different concatenations of the part name constituents and needed by the different tools in exactly this format.

If you're _not_ up for it then I propose workshopping a better naming for these two, for any unlucky soul who has to work on this after us.

whitequark:
@bl0x Let me know once you're happy with this PR.

bl0x:
`part` and `family` are needed to unambiguously fix the part. I've gone through device_list.csv that comes with Gowin IDE and checked. Neither `part` nor `family` alone has the full information needed.

Therefore, it is probably best (now and for the future) to use these two as input and check that they are somwhat sane and at least agree with each other.

bl0x:
@whitequark  I'm happier now with the PR =)
Thanks for all the comments.

whitequark:
@bl0x Thanks so much, you're a hero! I did one last round of small changes, mostly focused on making the regexp use a little more idiomatic, though I think `"\d"` will become invalid in newer Python versions as well so it has to be `r"\d"`.

Could you check that my edits are correct? Then I'm happy to merge this at last!

whitequark (approved):
Really happy with this PR!

bl0x (commented):
All fine with me! Thanks!

It's good to go then?

whitequark:
@bl0x Yep! I'm going to squash all the commits in this PR and then merge.

whitequark:
@bl0x Actually, I'm hitting an odd issue with hardware testing. I'm using [this](https://github.com/amaranth-lang/amaranth-boards/pull/227) Tang Nano platform file and it just doesn't blink at all. Do you know why?

bl0x:
Did it blink before at all? You've showed some video demo before. Was that the same device?

whitequark:
I've never used the built-in oscillator before since it was kind of awkward. I decided to try it out now that the support for it seems to be done properly.

bl0x:
I'll try and check with tang nano 9k later.

bl0x:
Using "OSC" and 25 MHz with tang nano 9k the 6 LEDs flash.
No buttons pressed: LED 0 and 1 are inverted vs. the rest
Buttons pressed: All LEDs flash with same phase.

With 12.5 MHz the flash speed increases (by factor 2?)
With 62.5 MHz the flash speed decreases (also by a factor of ~2), now at ~8 Hz toggle rate.

whitequark:
> No buttons pressed: LED 0 and 1 are inverted vs. the rest

This means you need to pass `invert=True` to the button resource.

> With 12.5 MHz the flash speed increases (by factor 2?) With 62.5 MHz the flash speed decreases (also by a factor of ~2), now at ~8 Hz toggle rate.

This seems wrong; the LEDs are supposed to always flash with 1 Hz period.

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,594 @@
...
+        r"""
+        {{invoke_tool("nextpnr-gowin")}}
+            {{quiet("--quiet")}}
+            {{get_override("nextpnr_opts")|options}}
+            --log {{name}}.tim
+            --device {{platform.part}}
+            --family {{platform._chipdb_device}}
+            --json {{name}}.json
+            --cst {{name}}.cst
+            --write {{name}}.json
```
whitequark, line 247: Just noticed it overwrites its input file. >_<

whitequark:
This looks like an upstream bug: https://github.com/YosysHQ/apicula/issues/193

@bl0x I'll fix the filename overwrite issue. Could you take a look at why blinky doesn't get the right frequency?

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,594 @@
...
+        clock.attrs["keep"] = "true"
+
+    @property
+    def default_clk_constraint(self):
+        if self.default_clk == "OSC":
+            if not hasattr(self, "osc_frequency"):
+                raise AttributeError(
+                    "Using the on-chip oscillator as the default clock source requires "
+                    "the platform.osc_frequency attribute to be set")
+            return Clock(self.osc_frequency / self._osc_div)
```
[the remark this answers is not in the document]
bl0x: That appears to fix it.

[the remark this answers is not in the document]
bl0x: I've tested with 62.5 MHz, 25 MHz and 2.5 MHz and the blinky always has the same frequency (~2 Hz toggle rate, i.e. 1 Hz rising-edge rate).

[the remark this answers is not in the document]
whitequark: Yep my bad!

whitequark:
As blinky works on both 1k and 9k, I'm merging this :tada:

bl0x:
@whitequark nice work!

amaranth/vendor/gowin.py
```diff
@@ -0,0 +1,594 @@
...
+            raise ValueError("Supplied part name is invalid")
+
+        self.series    = match.group(1)
+        self.voltage   = match.group(2)
+        self.size      = match.group(3)
+        self.subseries = match.group(4) or ""
+        self.package   = match.group(5)
+        self.speed     = match.group(6)
+
+        match = re.match(reg_series+reg_size+reg_subseries+"$", self.family)
```
whitequark, line 69: Hold on. This matches series, size, and subseries, *all of which are also matched above*. So what information does `family` actually add?

pepijndevos: It's... Complicated. Iirc device names don't actually uniquely define the actual device. I think I have an email from their support that's supposed to explain it but honestly it just doesn't make any sense.

whitequark: Have you looked at the code here?

josuah: It might be related to *revisions* of the chips, as shown on the package marking, as opposed to the datasheets:

https://github.com/brilliantlabsAR/docs/blob/e075bbe5784e6a61d23e10cbb86e26fe74a1831b/monocle/monocle.md#fpga-chip-revisions

So while a full part order name might be `GW1N-LV9 MG100C6/I5` the device family name *as shown in GOWIN EDA* software tool might be different according to... the extra chip marking information "xxxx**C**A0N" also shown here: `GW1N-9C`

Many thanks to @siliconwitch who found out about it.

And as you can see, this would also have a different Device ID.

whitequark: That is nuts.

bl0x: @josuah That is exactly why I'm doing the double-matching and why one also has to supply `family` to uniquely identify the part.

bl0x: Just take a look at the contents of `<path-to-gowin-IDE>/share/device/device_info.csv`.
Run this:

`awk -F"," '{print $2}' device_info.csv  | sort | uniq -c`

and you'll see that the part name in column 2 is *not unique* enough.

josuah: And the device family lacks the package information used for the pin mapping.

I suppose they would have preferred to avoid this split themselves.
Maybe a difference in the process with the same masks.
Maybe a fix that lead to more consequences than intended.

whitequark: @bl0x But the RE that matches `family` is a subset of the RE that matches `part`... what am I missing?

bl0x: @whitequark Nothing! Because part can also contain all information (and oftentimes does), but not always. We could use the same RE to also match `family`, but that would be too general.

PR #736, opened 2023-01-16
vendor.xilinx: Add support for more parts when using Symbiflow.

bl0x:
Adds support for parts used in Digilent CMOD boards.

### DIFF
--- a/amaranth/vendor/xilinx.py
+++ b/amaranth/vendor/xilinx.py
@@ -343,14 +343,20 @@ class XilinxPlatform(TemplatedPlatform):
 
     _symbiflow_part_map = {
         "xc7a35ticsg324-1L": "xc7a35tcsg324-1", # Arty-A7
+        "xc7a15tcpg236-1": "xc7a15tcpg236-1", # Cmod-A7-15
+        "xc7a35tcpg236-1": "xc7a35tcpg236-1", # Cmod-A7-35
     }
 
     _symbiflow_bitstream_device_map = {
             "xc7a35ticsg324-1L": "artix7",
+            "xc7a15tcpg236-1": "artix7",
+            "xc7a35tcpg236-1": "artix7",
     }
 
     _symbiflow_device_map = {
             "xc7a35ticsg324-1L": "xc7a50t_test",
+            "xc7a15tcpg236-1": "xc7a50t_test",
+            "xc7a35tcpg236-1": "xc7a50t_test",
     }
 
     _symbiflow_required_tools = [
--- a/amaranth/vendor/xilinx.py
+++ b/amaranth/vendor/xilinx.py
@@ -4,6 +4,7 @@ from ..hdl import *
 from ..lib.cdc import ResetSynchronizer
 from ..build import *
 
+import re
 
 __all__ = ["XilinxPlatform"]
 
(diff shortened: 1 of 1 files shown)


amaranth/vendor/xilinx.py
```diff
@@ -343,14 +343,20 @@ def vendor_toolchain(self):
...
     _symbiflow_part_map = {
         "xc7a35ticsg324-1L": "xc7a35tcsg324-1", # Arty-A7
+        "xc7a15tcpg236-1": "xc7a15tcpg236-1", # Cmod-A7-15
+        "xc7a35tcpg236-1": "xc7a35tcpg236-1", # Cmod-A7-35
     }
 
     _symbiflow_bitstream_device_map = {
             "xc7a35ticsg324-1L": "artix7",
+            "xc7a15tcpg236-1": "artix7",
+            "xc7a35tcpg236-1": "artix7",
```
whitequark, line 353: I'm quite sure this map could be replaced with an `if`/`else` chain, since a prefix of `xc7a` implies the device will be Artix-7.

amaranth/vendor/xilinx.py
```diff
@@ -343,14 +343,20 @@ def vendor_toolchain(self):
 
     _symbiflow_part_map = {
         "xc7a35ticsg324-1L": "xc7a35tcsg324-1", # Arty-A7
```
whitequark, line 345: Does this map just cut off the last letter of the speed grade? (cc @mwkmwkmwk, I don't quite have Xilinx part numbering scheme memorized)

amaranth/vendor/xilinx.py
```diff
@@ -343,14 +343,20 @@ def vendor_toolchain(self):
...
     _symbiflow_bitstream_device_map = {
             "xc7a35ticsg324-1L": "artix7",
+            "xc7a15tcpg236-1": "artix7",
+            "xc7a35tcpg236-1": "artix7",
     }
 
     _symbiflow_device_map = {
             "xc7a35ticsg324-1L": "xc7a50t_test",
+            "xc7a15tcpg236-1": "xc7a50t_test",
+            "xc7a35tcpg236-1": "xc7a50t_test",
```
whitequark, line 359: This one seems fine though.

whitequark (changes requested):

amaranth/vendor/xilinx.py
mwkmwkmwk: yes, specifically cuts the "low-power" designation

whitequark: Ok yeah we can definitely use a function here.

amaranth/vendor/xilinx.py
bl0x: This looks a bit like a mess (applies to all three comments here):

The [Arty A7 reference manual](https://digilent.com/reference/programmable-logic/arty-a7/reference-manual) mentions a part called `xc7a35ticsg324-1L`, but a speed grade of '1L' does not exist in the [Artix 7 family](https://docs.xilinx.com/v/u/en-US/ds181_Artix_7_Data_Sheet), only '1LI'. The actual Arty A7 I have on hand *does not even have* a speed grade marking on the package, so I can neither confirm nor disprove.

To me it looks as if Digilent has managed to swap around some letters in the part name, while the proper part name should most probably be `xc7a35t-l1csg324i`, which is what you can order from the part vendors.

Symbiflow on the other hand has a naming scheme that is still different and requires the `_symbiflow_part_map` to map to their internal names.

The part names mentioned on the [CMOD A7 reference manual](https://digilent.com/reference/programmable-logic/cmod-a7/reference-manual) and the [CMOD S7 reference manual](https://digilent.com/reference/programmable-logic/cmod-s7/reference-manual) correctly refer to existing Xilinx parts.

Perhaps it is best - at least for new additions - to use the proper part names to avoid any more confusion.

To answer this particular comment: The `_symbiflow_bitstream_device_map` can be replaced by matching the prefix of the part name, yes. I'll do just that.

amaranth/vendor/xilinx.py
bl0x: I don't know exactly how symbiflow part naming is done, but having a map here is probably safest.

amaranth/vendor/xilinx.py
bl0x: Here a prefix match should also be able to do the job. The used architecture definition does not depend on package or speed grade afaict.

amaranth/vendor/xilinx.py
bl0x: Hmm, of course the maps rely on whatever is returned from the `_part` property. The format returned from there is also not what one would expect from the vendor datasheet.

bl0x: Oh: of course Vivado has a naming scheme different from both the datasheets and the device markings. The `_part` property returns something that Vivado understands. To use Symbiflow, the maps exist to transform into Symbiflow part names.

amaranth/vendor/xilinx.py
bl0x: The symbiflow [part db](https://github.com/chipsalliance/f4pga/blob/main/f4pga/flows/part_db.yml) seems to just drop the speed grade from the Vivado part name.

amaranth/vendor/xilinx.py
bl0x: I've added here whatever is also supported in f4pga project.

amaranth/vendor/xilinx.py
bl0x: I turned this into a re.sub call.

amaranth/vendor/xilinx.py
bl0x: This can be joined with the `_xray_family` property. Can't decide which style is nicer. `symbiflow_bitstream_device` is slightly more generic, while `_xray_family` is more concise. Also I prefer to raise an exception here instead of relying on useful error messages in the called tool.

amaranth/vendor/xilinx.py
```diff
@@ -341,17 +342,45 @@ def vendor_toolchain(self):
 
     # Symbiflow templates
 
-    _symbiflow_part_map = {
-        "xc7a35ticsg324-1L": "xc7a35tcsg324-1", # Arty-A7
-    }
+    # symbiflow does not distinguish between speed grades
+    # TODO: join with _xray_part
+    @property
+    def symbiflow_part(self):
```
whitequark, line 348: I think we don't necessarily need to export these as public names, unless we commit to maintaining them as a stable API. Could you make them private, please?

whitequark (changes requested):

amaranth/vendor/xilinx.py
bl0x: Done.

whitequark (dismissed):
LGTM

whitequark (approved):

bl0x:
Not sure, if rebasing to main is needed? It does mess with the history of the PR.

whitequark:
> Not sure, if rebasing to main is needed? It does mess with the history of the PR.

I don't tend to keep track of history, so it's fine. I am also going to squash on my side before merging since this is a simple PR, but for more complex I would ask to rebase and squash into self-contained changes.

bl0x:
Got it.

whitequark:
Thanks!

Issue #730, opened 2022-11-16
lib.data: enum Field width cannot be user-defined


jfng:
Repro:
```python3
import enum
from amaranth import *
from amaranth.lib import data


class B(enum.Enum):
    FOO = 0b00
    BAR = 0b01


class A(data.Struct):
    b: B
    c: unsigned(1)


a = A()
print(len(a.b)) # 1
```
The width of an enum Field is currently inferred from ``Shape.cast()``, as the smallest width required to represent all of its members.

In the above example, a user may actually want the ``A.b`` field to have a width of 2 bits. Currently, the enum would need a third member with a value of `0b10` or `0b11`.

whitequark:
I'm wondering if there are other cases for defining Enums that are wider when shape-cast than what is inferred from their variants alone, and if so, whether this should actually be fixed in the core language.

jfng:
Signals with enum shapes have the same issue, but it can be worked around by using an explicit width, with the enum as a decoder (e.g. ``s = Signal(unsigned(2), decoder=B)``).

Records also have this issue. Their (non-record) fields are signals, so the above workaround can work, with more awkwardness (e.g. ``r = Record([("b", unsigned(2))]); r.b.decoder = B``).

whitequark:
[Discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-02-06#1675706095-1675709437;) on 2023-02-06 weekly meeting. This feature needs an RFC; the discussion can be continued there.

whitequark:
I wrote an RFC for this feature: https://github.com/amaranth-lang/rfcs/pull/3

whitequark:
This works now:

```python
from amaranth import *
from amaranth.lib import data, enum


class B(enum.Enum, shape=unsigned(2)):
    FOO = 0b00
    BAR = 0b01


class A(data.Struct):
    b: B
    c: unsigned(1)


a = A()
print(len(a.b)) # 2
```

Issue #693, opened 2022-04-05
[RFC] Aggregate data structure library


whitequark:
# Aggregate data structure library RFC

## Summary

Add a rich set of standard library classes for accessing hierarchical aggregate data an idiomatic way, to fill one of the two major use cases of `Record` while avoiding its downsides.

See also #???.

## Motivation

Amaranth has a single data storage primitive: `Signal`. A `Signal` can be referenced on the left hand and the right hand side of an assignment, and so can be its slices. Although a signal serves the role of a numeric and bit container type fine, designs often include signals used as bit containers whose individual bits are named and have unique meanings. A mechanism that allows referring to such bit fields by name is essential.

Currently, the role of this mechanism is served by `Record`. However, `Record` has multiple major drawbacks:

  1. `Record` attempts to do too much: it is both a mechanism for _controlling representation_ (including implicitly casting a record to a value) and a mechanism for _defining interfaces_ (specifying signal directions and facilitating connections between records).

     These mechanisms should be defined separately, since the only aspect they have in common is using a container class that consists of multiple named fields. Conflating the two mechanisms constraints the design space, making addressing the other drawbacks impossible, and the ill-defined scope encourages bugs in downstream code.

  2. `Record` has limited composability: records can only be nested within each other. Practical designs (in particular, implementations of protocols) use data with complex representation beyond nested sequences of fields: these include overlaid sequences of fields (where interpretation alternates based on some discriminant) and arrays of fields (where a field can be indexed by a non-constant), where any individual field can have a complex representation itself.

     `Record` is structured as a sequence of `Signal`s, which is a part of its API. As such, it cannot support overlaid fields, and implementing support for arrays of fields is challenging.

  3. `Record` has limited introspectability: while its `layout` member can be accessed to enumerate its fields, the results do not include field boundaries, and the types of the returned shape-castable objects are preserved only as an implementation detail. `Layout` objects themselves are also not shape-castable.

     `Record` and `Layout` are structured as a sequence of `Signal`s rather than a view into an underlying bit container, which is reflected in its API. Thus, `Layout` does not fit into Amaranth's data model, which concerns individual values.

  4. `Record` comes with its own storage: while its `fields` argument can be used to substitute the signals that individual fields are pointing to (in an awkward and error-prone way), it is still a collection of `Signal`s. Using `Record` to impose structure on an existing `Value` requires a `Module` and a combinatorial assignment. This is an unnecessary complication, especially in helper functions.

[...]

PR #739, opened 2023-01-21
hdl.ast: allow IntEnum/IntFlag in Cat.

Arusekk:
Example code that can make use of such changes:

```py
class Something(IntEnum):
    A = 0b1000
    B = 0b0000

...

class Component(Elaboratable):
        m = Module()
        with m.Switch(...):
             with m.Case(Cat(Something.A, Something.B)):
                 ...
```

The first commit is more important to me than the second one (you can always work it around using `Case(Cat(...)._as_const())`), but they are pretty much independent, so if it is desirable to only merge one of them, you can do it.

### DIFF
--- a/docs/lang.rst
+++ b/docs/lang.rst
@@ -235,6 +235,9 @@ Casting a value from an integer ``i`` is a shorthand for ``Const(i)``:
    >>> Value.cast(5)
    (const 3'd5)
 
+.. note::
+
+   If a value subclasses :class:`enum.IntEnum` or its class otherwise inherits from both :class:`int` and :class:`Enum`, it is treated as an enumeration.
 
 Values from enumeration members
 -------------------------------
@@ -246,6 +249,10 @@ Casting a value from an enumeration member ``m`` is a shorthand for ``Const(m.va
    >>> Value.cast(Direction.LEFT)
    (const 2'd1)
 
+
+.. note::
+
+   If a value subclasses :class:`enum.IntEnum` or its class otherwise inherits from both :class:`int` and :class:`Enum`, it is treated as an enumeration.
 
 .. _lang-signals:
 
--- a/amaranth/hdl/ast.py
+++ b/amaranth/hdl/ast.py
@@ -168,12 +168,12 @@ class Value(metaclass=ABCMeta):
         while True:
             if isinstance(obj, Value):
                 return obj
-            elif isinstance(obj, int):
-                return Const(obj)
-            elif isinstance(obj, Enum):
-                return Const(obj.value, Shape.cast(type(obj)))
             elif isinstance(obj, ValueCastable):
                 new_obj = obj.as_value()
+            elif isinstance(obj, Enum):
+                return Const(obj.value, Shape.cast(type(obj)))
+            elif isinstance(obj, int):
+                return Const(obj)
             else:
                 raise TypeError("Object {!r} cannot be converted to an Amaranth value".format(obj))
             if new_obj is obj:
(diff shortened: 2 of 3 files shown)


whitequark:
Please split this PR into two and add tests.

whitequark:
So there are actually two distinct changes in this PR that I'd like to have addressed separately:
  1. Fixing a bug in value casting code that would treat `IntEnum` values as integers and not enumerations. This is currently unspecified, but since the [documentation](https://amaranth-lang.org/docs/amaranth/latest/lang.html#shapes-from-enumerations) lists that the Enum behavior specializes the `int` behavior (widens the result), I think they should be indeed treated as enumerations.
  2. Changing the behavior of the diagnostic that prints a warning if a bare integer is used in a `Cat` context. This is actually still a problem with enums because it's not possible to specify width of an enumeration explicitly. That's tracked in #730 though.

In addition you only have tests for `IntEnum` while the change is applied to `Enum` too.

Could you please split this into two commits? One introducing the change (1), adding a test for bare `Enum`, and adding a line into both of the sections of documentation [here](https://amaranth-lang.org/docs/amaranth/latest/lang.html#value-casting) to mention that if a value is both `int` and `Enum`, `Enum` behavior takes preference. Another introducing the change (2) silencing the warning for `IntEnum`, the way it's already silenced for `Enum`.

Arusekk:
ad 1: Sure.
ad 2: ...and you got me on this one.  The whole PR is mainly because I managed to make a proof-of-concept implementation of an enum with width; would you like me to add it in a next PR?
```py
class BitEnum(ValueCastable, Enum):
    """
    A helper class that defines Amaranth enums with fixed width
    """
    def __init_subclass__(cls, *, width):
        cls.__width = width

    @ValueCastable.lowermethod
    def as_value(self):
        return Const(self.value, self.__width)

class Color(BitEnum, width=5):
    RED  = 0b00001
    BLUE = 0b00010
```
Another option would be to insert such logic to Const and define enums as `class Color(Const, Enum, width=5):`, possibly with an extra check that the subclass is a subclass of Enum, to still prevent reckless Const subclassing.

(actually the above does not work due to Python incorrectly assuming that `ValueCastable` is a data type, not a mix-in class, because `ValueCastable` defines a `__new__` method—maybe the checks could be moved to `Value.cast`? the below does work, however)
<details>

```py
class ValueCastableHack(int, ValueCastable):
    @ValueCastable.lowermethod
    def as_value(self):
        raise NotImplementedError("width information lost!")


class BitEnum(ValueCastableHack, Enum):
    """
    A helper class that defines Amaranth enums with fixed width
    """
    def __init_subclass__(cls, *, width):
        cls.__width = width

    @ValueCastable.lowermethod
    def as_value(self):
        return Const(self.value, self.__width)

class Color(BitEnum, width=5):
    RED  = 0b00001
    BLUE = 0b00010
```

</details>

whitequark:
> would you like me to add it in a next PR?

I think such a change needs an RFC first. Please do write one (#693 can be used as an example re: format) and submit it together with your implementation as the proposed one, especially if you do already have it.

We don't have to wait until the RFC is merged to merge these changes though.

docs/lang.rst
```diff
@@ -235,6 +235,9 @@ Casting a value from an integer ``i`` is a shorthand for ``Const(i)``:
    >>> Value.cast(5)
    (const 3'd5)
 
+.. note::
+
+   If a value subclasses :class:`enum.IntEnum` or is otherwise both int and Enum, Enum behavior takes precedence.
```
whitequark, line 240: I'd phrase this as:

    If a value subclasses :class:`enum.IntEnum` or its class otherwise inherits from both :class:`int` and :class:`enum.Enum`, it is treated as an enumeration.

whitequark (changes requested):
One minor change. Looks great otherwise!

whitequark (approved):
LGTM

whitequark:
Thanks for this excellent PR! I'm looking forward to your future work on Amaranth, would such arise.

PR #740, opened 2023-01-22
hdl.{ast,dsl}: allow Cat in matches/Case.

Arusekk:
Split from #739

### DIFF
--- a/amaranth/hdl/ast.py
+++ b/amaranth/hdl/ast.py
@@ -423,6 +423,8 @@ class Value(metaclass=ABCMeta):
         """
         matches = []
         for pattern in patterns:
+            if isinstance(pattern, Value):
+                pattern = pattern._as_const()
             if not isinstance(pattern, (int, str, Enum)):
                 raise SyntaxError("Match pattern must be an integer, a string, or an enumeration, "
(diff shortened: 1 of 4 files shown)


whitequark:
I'm not sure how much I want to add this to the language. We don't currently have a concept of a "constant expression" (`_as_const` is internal for this reason) and it would be a substantial new concept to introduce. On the flipside, I've encountered the need for this functionality myself and it's not always clear what to replace it with.

Feedback from others here is welcome. (@jfng?)

whitequark:
@Arusekk, thank you for submitting the PR. At the moment Amaranth does not have the concept of a constant expression. It would be valuable if it did; however such a change requires going through the [RFC process](https://github.com/amaranth-lang/rfcs#when-you-need-to-follow-this-process). Since this change did not go through the RFC process, unfortunately I have to close this pull request at the moment.

I've opened an issue https://github.com/amaranth-lang/amaranth/issues/751 to discuss and track progress on this change.

whitequark:
Eventually, Amaranth did get an RFC for constant-castable expressions (https://amaranth-lang.org/rfcs/0004-const-castable-exprs.html), and this change made it in.

PR #758, opened 2023-03-01
add Lattice Nexus vendor platform (using nextpnr-nexus and prjoxide)

slagernate:
tested on crosslink-nx board, blinking all 14 LEDs


amaranth/vendor/lattice_nexus.py
```diff
@@ -0,0 +1,675 @@
...
+
+    @property
+    def default_clk_constraint(self):
+        if self.default_clk == "OSC":
+            return Clock(150e6 / self.osc_div) # lscc/radiant/2022.1/docs/webhelp/eng/index.htm#page/Reference%20Guides/FPGA%20Libraries%20Reference%20Guide/Alphanumeric_Primitives_List.11.155.htm        
+        # For HFCLKOUT (created clock): ADJUSTED_PERIOD=6666.66667ps * HF_CLK_DIV
+        return super().default_clk_constraint
+
+    def create_missing_domain(self, name):
+        # Lattice ECP5 devices have two global set/reset signals: PUR, which is driven at startup
```
whitequark, line 302: Should this be changed to mention a different family?

amaranth/vendor/lattice_nexus.py
```diff
@@ -0,0 +1,675 @@
...
+            if self.default_clk == "OSC":
+                if not hasattr(self, "osc_div"):
+                    raise ValueError("OSC divider (osc_div) must be an integer between 2 "
+                                     "and 128")
+                if not isinstance(self.osc_div, int) or self.osc_div < 2 or self.osc_div > 128:
+                    raise ValueError("OSC divider (osc_div) must be an integer between 2 "
+                                     "and 128, not {!r}"
+                                     .format(self.osc_div))
+                clk_i = Signal()
+                m.submodules += Instance("OSC", p_HF_CLK_DIV=self.osc_div, o_HFCLKOUT=clk_i)
```
whitequark, line 319: Have you checked the oscillator?

amaranth/vendor/lattice_nexus.py
```diff
@@ -0,0 +1,675 @@
...
+            # There is no end-of-startup signal on ECP5, but PUR is released after IOB enable, so
+            # a simple reset synchronizer (with PUR as the asynchronous reset) does the job.
+            # LIFCL UPDATE: GSR() has a parameter (SYNCMODE) to perform synchronization of the reset to the user clock
+            m.submodules += [
+                #Instance("FD1S3AX", p_GSR="DISABLED", i_CK=clk_i, i_D=~rst_i, o_Q=gsr0),
+                #Instance("FD1S3AX", p_GSR="DISABLED", i_CK=clk_i, i_D=gsr0,   o_Q=gsr1),
+                ## Although we already synchronize the reset input to user clock, SGSR has dedicated
+                ## clock routing to the center of the FPGA; use that just in case it turns out to be
+                ## more reliable. (None of this is documented.)
+                #Instance("SGSR", i_CLK=clk_i, i_GSR=gsr1),
```
whitequark, line 338: Looks good--I think you can delete the pre-LIFCL code and only leave the current GSR configuration

amaranth/vendor/lattice_nexus.py
```diff
@@ -0,0 +1,675 @@
...
+        "LVCMOS12", "LVCMOS12H", 
+        "LVCMOS10", "LVCMOS10H", "LVCMOS10R", 
+        "SSTL135_I", "SSTL135_II", "SSTL15_I", "SSTL15_II",
+        "HSUL12", "MIPI D-PHY LP Input (LVCMOS12)"
+    ]
+    _differential_io_types = [
+        "LVDS", "LVDSE", "subLVDS", "subLVDSEH", "SLVS", "MIPI D-PHY", "LVCMOS33D",
+        "LVTTL33D", "LVCMOS25D", "SSTLD_I", "SSTL135D_I", "SSTL15D_I", "SSTL15D_II",
+        "HSUL15D_I", "HSUL12D", 
+    ]
```
whitequark, line 363: Have you checked this list against the datasheet?

whitequark (commented):
Looks very good--I've highlighted a few areas where I'd like clarification

amaranth/vendor/lattice_nexus.py
slagernate: yes

amaranth/vendor/lattice_nexus.py
slagernate: with 
```
        platform.osc_div = 2
        platform.default_clk = "OSC"
```
in Blinky.py, seems I'm getting 
```
ERROR: Cell type 'OSC' instantiated as 'cd_sync.U$$0' is not supported by this device.
```
`top.debug.v` looks like:
```
  OSC #(
    .HF_CLK_DIV("2")
  ) \U$$0  (
    .HFCLKOUT(clk_i)
  );
 ```
 LMK if I'm missing something

amaranth/vendor/lattice_nexus.py
slagernate: ok

amaranth/vendor/lattice_nexus.py
slagernate: yes

amaranth/vendor/lattice_nexus.py
whitequark: Let's just remove oscillator support then, it can be added later.

amaranth/vendor/lattice_nexus.py
gatecat: I don't think this is correct, I would expect things like `MIPI_DPHY` instead of `MIPI D-PHY` (although you could check if radiant actually supports both I doubt it?)

there's a list in nextpnr: https://github.com/YosysHQ/nextpnr/blob/2f509734dff676b7543a574b64a46be4224a5aa4/nexus/io.cc#L26 but this might not be definitive

amaranth/vendor/lattice_nexus.py
gatecat: Oops, I think I forgot to include the mapping from `OSC` to `OSC_CORE` in nextpnr

amaranth/vendor/lattice_nexus.py
gatecat: I don't think nextpnr-nexus ever gained GSR support, fyi

amaranth/vendor/lattice_nexus.py
slagernate: My bad, followed the datasheet a little too closely (https://www.latticesemi.com/view_document?document_id=52780, pg. 59). I'll fix the lowercase, and move `MIPI D-PHY` to `MIPI_DPHY` as shown in nextpnr. Will leave 'MIPI D-PHY LP Input (LVCMOS12)` as is --more so an fyi (that you can lvcmos12 for soft mipi LS, see note 6 in datasheet) than something people would actually use.

amaranth/vendor/lattice_nexus.py
slagernate: I see. I was just checking the primitives reference guide for the crosslink (`lscc/radiant/2022.1/docs/webhelp/eng/Reference Guides/`). You're right tho, pressing rst on the board does nothing.

amaranth/vendor/lattice_nexus.py
slagernate: I noticed LIteX uses OSCA --let me know if I should just use that

PR #761, opened 2023-03-03
build.run: Handle UTF-8 encoding errors in SSH output gracefully

neuschaefer:
In some cases, a toolchain might produce shell output that isn't correct UTF-8. To avoid crashing in such cases, pass errors="replace" to bytes.decode.

For example, Lattice Diamond uses the Latin-1 encoding for some reason. This recently broke my setup because the month turned to "März" in a German locale:

--- Start Time: Fr. M�r 3 20:01:41 2023


whitequark:
Thanks!

PR #767, opened 2023-03-21
hdl.ast: Test *Castable subclasses on definition.

Arusekk:
The `__init_subclass__` method fires on class definition rather than use. It also has the bonus impact that no `__new__` method is defined, so the classes can be correctly detected as mix-in classes by modules such as enum.


whitequark (approved):
Thank you! This is a very good idea that didn't occur to me for some reason.

This will also improve the documentation, because at the moment `__new__` is confusing sphinx-autodoc in a way that's fairly irritating. Good things all around!

PR #799, opened 2023-06-01
Fix tests on 3.12

whitequark:


mwkmwkmwk:
> It looks like (according to @mwkmwkmwk) the following test failure is a Python regression and not an Amaranth issue:

(to be clear, I think it's a coverage.py issue on new Python version, not a Python regression)

whitequark:
Ah sorry, I was distracted and missed that part!

mwkmwkmwk:
For reference, the remaining test failure is caused by https://github.com/python/cpython/issues/105162 , which has just been fixed in git main; the fix will presumably be included in the next 3.12 beta.

PR #801, opened 2023-06-04
Allow pyvcd 0.4

psentee:
The only breaking change in 0.4 is dropping support for Python 3.6, which Amaranth has already dropped. It also adds official support for 3.10 and 3.11.

https://github.com/westerndigitalcorporation/pyvcd/blob/3ad7340072791e695e32cfb51135b751a8ddab41/CHANGELOG.rst


whitequark (approved):
Thanks!

PR #814, opened 2023-06-07
hdl.ast: fix shape for subtraction.

### LINKED ISSUE #813: Subtract of unsigned `Value`s should have signed shape
```
>>> a = Const(1) - Const(2)                                                                                                                                                                     
>>> a.shape()
unsigned(3)
```

The correct result here (`-1`) is obviously not representable in an unsigned shape.

mwkmwkmwk:
Fixes #813.


whitequark (approved):
LGTM

PR #820, opened 2023-06-07
hdl.ast: make `Value.__abs__` return unsigned shape.

mwkmwkmwk:


amaranth/hdl/ast.py
```diff
@@ -263,7 +263,7 @@ def __ge__(self, other):
 
     def __abs__(self):
         if self.shape().signed:
-            return Mux(self >= 0, self, -self)
```
whitequark, line 266: I think this could use `as_unsigned`?

mwkmwkmwk: slicing already implicitly converts to unsigned, and `as_unsigned` without slicing would result in the shape being needlessly one bit wider than input

whitequark (approved):
LGTM

PR #826, opened 2023-06-22
sim._pyrtl: py3.12+: convert to int before bitwise negating.

charlottia:
Amaranth bitwise negation `~` compiles to Python bitwise negation `~` in simulation; the same holds for comparison operators such as `==`. Thus an expression such as `~(a == b)` in simulation will compile to Python that takes the bitwise negation of the comparison result, which will be an actual bool.

On 3.12, the result is a `DeprecationWarning` emitted only at simulation run-time.

When negating in simulation, coerce the value to an int. `mask` is sufficient as we do no further arithmetic here.

### DIFF
--- a/amaranth/sim/_pyrtl.py
+++ b/amaranth/sim/_pyrtl.py
@@ -138,7 +138,7 @@ class _RHSValueCompiler(_ValueCompiler):
         if len(value.operands) == 1:
             arg, = value.operands
             if value.operator == "~":
-                return f"(~{self(arg)})"
+                return f"(~{mask(arg)})"
             if value.operator == "-":
                 return f"(-{sign(arg)})"
             if value.operator == "b":
--- a/tests/test_sim.py
+++ b/tests/test_sim.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from contextlib import contextmanager
 
 from amaranth._utils import flatten
@@ -871,6 +872,15 @@ class SimulatorIntegrationTestCase(FHDLTestCase):
             with open(os.path.devnull, "w") as f:
                 with sim.write_vcd(f):
                     pass
+
+    def test_no_negated_boolean_warning(self):
+        m = Module()
+        a = Signal()
+        b = Signal()
+        m.d.comb += a.eq(~(b == b))
+        with warnings.catch_warnings(record=True) as warns:
+            Simulator(m).run()
+            self.assertEqual(warns, [])
 
 
 class SimulatorRegressionTestCase(FHDLTestCase):


whitequark (approved):
LGTM, thank you!

Do you happen to know why we aren't hitting that in existing tests? We have `PYTHONWARNINGS=error` but the build succeeds silently for some reason...

whitequark:
Huh that's an interesting failure! Let me retry.

charlottia:
> Huh that's an interesting [failure](https://github.com/amaranth-lang/amaranth/actions/runs/5348336375/jobs/9698085573)!

O_O. Worrying.

> Do you happen to know why we aren't hitting that in existing tests? We have `PYTHONWARNINGS=error` but the build succeeds silently for some reason...

I believe it's as simple as no existing simulation test covering this. Here's an example I've come up with which actually brings us some new questions:

```python
    def test_invert(self):
        stmt = lambda y, a: y.eq(~a)
        self.assertStatement(stmt, [C(0b0000, 4)], C(0b1111, 4))
        self.assertStatement(stmt, [C(0b1010, 4)], C(0b0101, 4))
        self.assertStatement(stmt, [C(0,      4)], C(-1,     4))
```

This is the existing `tests.test_sim.SimulatorUnitTestCase.test_invert`. Consider this statement instead:

```python
y.eq(~a | ~(a == a))
```

Firstly, this will trigger the warning (on 3.12, without the commit from this branch).

Secondly, should this cause the assertions to fail, or not? `a == a` should effectively reduce to `C(1, 1)`, `~(a == a)` to `C(0, 1)`, and so the whole expression should reduce to `~a | C(0, 1)` and make no difference to the result.

That doesn't happen, though:

```
======================================================================
FAIL: test_invert (tests.test_sim.SimulatorUnitTestCase.test_invert)
  File "/home/charlotte/g/amaranth/tests/test_sim.py", line 47, in test_invert
    self.assertStatement(stmt, [C(0b1010, 4)], C(0b0101, 4))
[...]
  File "/home/charlotte/g/amaranth/tests/test_sim.py", line 39, in process
    self.assertEqual((yield osig), output.value)
AssertionError: 15 != 5
```

The reason is the use of `self(lhs)` and `self(rhs)` in compiling bitwise binary operators. Because we don't mask the operands, we end up with:

```python
((~(0xf & slots[1].curr)) | (~(0x1 & ((0xf & slots[1].curr) == (0xf & slots[1].curr)))))
# =>
((~(0xf & slots[1].curr)) | (~(0x1 & True)))
# =>
((~(0xf & slots[1].curr)) | ~0x1)
# =>
((~(0xf & slots[1].curr)) | -2)
# =>
((~0b1010) | -2)
# =>
(0b0101 | -2)
# =>
0b0101 | 0b1110
# =>
0b1111
```

When correctly masked, we end up with `(0x1 & (~0x1))` on the RHS, which gives `0` as expected.

If you concur, happy to open a PR to further test and address this.

whitequark:
> The reason is the use of `self(lhs)` and `self(rhs)` in compiling bitwise binary operators. Because we don't mask the operands, we end up with:

Oh no!! This is *definitely* a bug. Thank you for discovering it!

PR #827, opened 2023-06-24
sim._pyrtl: mask bitwise binary operands.

charlottia:
Addresses the bug surfaced in https://github.com/amaranth-lang/amaranth/pull/826#issuecomment-1603634393.

I read through the remaining value compiler branches (and poked at a few things to see if I could break them), but I believe it's all OK from here.


whitequark (approved):
LGTM, thank you!

PR #829, opened 2023-06-29
lib.enum: allow `import * from amaranth.lib.enum`.

charlottia:
There's an actual `py_enum.member` (which we briefly overwrite our loop index with (!)).  We delete our `member`, but it's still in the `__all__` that came from `py_enum`, so `import *` fails.


whitequark (approved):
Oops! Thanks ^^

PR #831, opened 2023-07-02
lib.data: remove unnecessary loop.

charlottia:
Found during #830 simplification pass; copy-paste leftover from similar sites that repeatedly `.as_shape()` to unwrap.


whitequark (approved):
LGTM

PR #832, opened 2023-07-02
lib.enum: allow empty enums.

charlottia:
Fixes the error noted in https://github.com/amaranth-lang/amaranth/pull/830#issuecomment-1613690529 where empty Python `Enum`s are shape-castable to `unsigned(0)`.

~~We move the member check from the Amaranth `Enum` implicit shape-cast into the Python `Enum` shape-cast, which the former backs onto.~~

Allows empty Amaranth `Enum`s to shape-cast to `unsigned(0)`, like Python `Enum`s, per https://github.com/amaranth-lang/amaranth/pull/832#issuecomment-1617639806.

### DIFF
--- a/amaranth/lib/enum.py
+++ b/amaranth/lib/enum.py
@@ -112,16 +112,11 @@ class EnumMeta(ShapeCastable, py_enum.EnumMeta):
         if hasattr(cls, "_amaranth_shape_"):
             # Shape was provided explicitly; return it.
             return cls._amaranth_shape_
-        elif cls.__members__:
-            # Shape was not provided explicitly, but enumeration has members; treat it
-            # the same way `Shape.cast` treats standard library enumerations, so that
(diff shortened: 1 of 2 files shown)


whitequark:
I was actually thinking of allowing it everywhere instead. (This would technically be a breaking change, but even ignoring that, I can't think of any reason to disallow such enums. Do you?)

charlottia:
I think it's totally fine! I'll see if anything comes up.

charlottia:
Seems OK. https://github.com/amaranth-lang/amaranth/pull/830#issuecomment-1613690529 mentions inheritance concerns (presumably because the base enum is also an empty enum and so things can get weird), but I can't think of a scenario that's properly impacted by allowing this.

whitequark (approved):
Thank you~!

PR #833, opened 2023-07-03
lib.io: allow 0-width `Pin`

mwkmwkmwk:
The main purpose of this change is migrating glasgow from the compat `TSTriple` (which allows 0 width) to `Pin`.  This sort of change would normally require a RFC, but `Pin` is already slated for removal/replacement, so that was deemed to be unnecessary.

### DIFF
--- a/amaranth/lib/io.py
+++ b/amaranth/lib/io.py
@@ -12,8 +12,8 @@ def pin_layout(width, dir, xdr=0):
 
     See :class:`Pin` for details.
     """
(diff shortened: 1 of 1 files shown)


whitequark:
Could you clarify it in the commit message that this change happens for Glasgow and because we intend to update/deprecate the Pin abstraction in the future anyway? Normally this would require an RFC but `Pin` goes away so I'd say that's just useless paperwork...

whitequark (approved):

PR #840, opened 2023-07-18
(Merge after 0.4 is released) Remove the Migen compatibility layer

### LINKED ISSUE #692: Remove the Migen compatibility layer
It has (technically) been deprecated since day 1, and hasn't seen much uptake. This interface was considered important for migration of existing codebases, but the (intentional and unintentional) mismatches with Migen behavior meant that it only had a limited usefulness; most code that needed Migen compatibility combines the LiteX and the Amaranth parts of the design through Verilog.

Removing the compatibility layer allows a number of cleanups: the AST visitors can dispatch faster using `type() is` rather than `isinstance()`, and several AST types can be marked `@final`.

It also deemphasizes the relation to Migen. Many of the remaining design choices in Amaranth that directly derive from Migen (records, fragments, AST transformations, board DSL, ...) are load bearing liabilities and there's no reason to have a gratuitous one on top of that.

whitequark:
Fixes #692.

This will stay a draft until a resolution on #692 is reached.


whitequark:
Note: this will need another PR to add a changelog entry after merging.

PR #841, opened 2023-07-18
lib.data: improve diagnostics for field access on array layout view

### LINKED ISSUE #837: TypeError when accessing reset attribute from a Signal created with an ArrayLayout
Repro:
```python3
from amaranth import *
from amaranth.lib import data

layout = data.ArrayLayout(unsigned(1), 4)
a = Signal(layout, reset=[0,0,0,0])
a.reset
```

Output:
```
  File "/tmp/repro.py", line 7, in <module>
    a.reset
  File "/home/jf/src/amaranth/amaranth/lib/data.py", line 701, in __getattr__
    item = self[name]
           ~~~~^^^^^^
  File "/home/jf/src/amaranth/amaranth/lib/data.py", line 666, in __getitem__
    value = self.__target.word_select(key, Shape.cast(self.__layout.elem_shape).width)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jf/src/amaranth/amaranth/hdl/ast.py", line 400, in word_select
    offset = Value.cast(offset)
             ^^^^^^^^^^^^^^^^^^
  File "/home/jf/src/amaranth/amaranth/hdl/ast.py", line 174, in cast
    raise TypeError("Object {!r} cannot be converted to an Amaranth value".format(obj))
TypeError: Object 'reset' cannot be converted to an Amaranth value
```

whitequark:
Fixes #837.

### DIFF
--- a/amaranth/lib/data.py
+++ b/amaranth/lib/data.py
@@ -662,6 +662,10 @@ class View(ValueCastable):
             If ``ShapeCastable.__call__`` does not return a value or a value-castable object.
         """
         if isinstance(self.__layout, ArrayLayout):
+            if not isinstance(key, (int, Value, ValueCastable)):
+                raise TypeError("Views with array layout may only be indexed with an integer "
+                                "or a value, not {!r}"
+                                .format(key))
             shape = self.__layout.elem_shape
             value = self.__target.word_select(key, Shape.cast(self.__layout.elem_shape).width)
         else:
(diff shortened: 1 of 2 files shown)


jfng:
LGTM. The assumption behind #837 was that an object returned by the `Signal` constructor was an instance of `Signal`. This fix clarifies that it is in fact not the case, unless `.as_value()` is called.

PR #847, opened 2023-07-22
lib.data: fix `Layout.const` masking for signed fields.

### LINKED ISSUE #846: `Layout.const` doesn't properly mask signed values
```
from amaranth.lib import data
from amaranth import *

class Abc(data.Struct):
    a: signed(8)
    b: signed(8)
    c: signed(8)

print(hex(Abc.const({
    'a': 1,
    'c': 1,
    'b': -1,
}).value))

print(hex(Abc.const({
    'a': 1,
    'b': -1,
    'c': 1,
}).value))
```

Result:

```
0xffff01
0x1ff01
```

mwkmwkmwk:
Fixes #846.


whitequark (approved):
LGTM

PR #858, opened 2023-08-07
Create Glossary

AllysonProlisko:
First draft of Glossary document

### DIFF
--- a/docs/Glossary
+++ b/docs/Glossary
@@ -1,0 +1,42 @@
+Glossary
+========
+
++-----------------------+----------------------------------------------+
+| .. rubric:: Term      | .. rubric:: Definition                       |
+|    :name: term        |    :name: definition                         |
++=======================+==============================================+
++-----------------------+----------------------------------------------+
+
++-------------------+--------------------------------------------------+
+| Amaranth          | An open-source toolchain for developing hardware |
+|                   | based on synchronous digital logic using the     |
+|                   | Python programming language                      |
++===================+==================================================+
+| ASIC              | Application-specific integrated circuit          |
++-------------------+--------------------------------------------------+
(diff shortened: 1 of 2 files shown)


docs/Glossary
```diff
@@ -0,0 +1,42 @@
+Glossary
+========
+
++-----------------------+----------------------------------------------+
```
whitequark, line 4: Rather than using a table, the glossary should use a [definition list](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists), which is a syntax that is specially suited for e.g. a glossary.

docs/Glossary
```diff
@@ -0,0 +1,42 @@
...
+|                   | — inputs and one output                          |
++-------------------+--------------------------------------------------+
+| LUT               | Look up table                                    |
++-------------------+--------------------------------------------------+
+| MCU               | Microcontroller unit                             |
++-------------------+--------------------------------------------------+
+| Propagation delay | The time it takes for a signal to move from      |
+|                   | source to destination                            |
++-------------------+--------------------------------------------------+
+| RTL               | Register transfer language                       |
```
whitequark, line 41: This should be ["Register-transfer level"](https://en.wikipedia.org/wiki/Register-transfer_level)

whitequark (changes requested):

docs/Glossary
AllysonProlisko: That's done now.

AllysonProlisko:
@whitequark I've update this with the changes from the other PR and closed that one.

whitequark:
Superseded by https://github.com/amaranth-lang/amaranth/pull/869.

PR #865, opened 2023-08-12
Implement RFC #2

### LINKED ISSUE #342: Deprecate `Record.connect`

whitequark:
Fixes #342.

- [x] Implement all of the RFC functionality
  - [ ] Resolve FIXMEs
- [ ] Write tests for all of the RFC functionality
- [ ] Document all of the RFC functionality


whitequark:
The entire RFC's functionality (sans unresolved questions in the current document, which cover a very minor part of it) was implemented and tested to 100% coverage.

amaranth/lib/wiring.py
```diff
@@ -0,0 +1,1041 @@
...
+import inspect
+import re
+import warnings
+
+from ..hdl.ast import Shape, ShapeCastable, Const, Signal, Value, ValueCastable
+from ..hdl.ir import Elaboratable
+from .._utils import final
+
+
+__all__ = ["In", "Out", "Signature", "connect", "transpose", "Component"]
```
charlottia, line 13:
```suggestion
__all__ = ["In", "Out", "Signature", "connect", "flipped", "Component"]
```

PR #884, opened 2023-08-31
docs: emphasize that `shape=` is optional

whitequark:


whitequark:
@cr1901 r?

cr1901:
LGTM! Already am using exclusively `amaranth.lib.enum` locally now :).

Issue #879, opened 2023-08-22
Deprecate `Record`


whitequark:
Now that we have [RFC 1](https://amaranth-lang.org/rfcs/0001-aggregate-data-structures.html) and [RFC 2](https://amaranth-lang.org/rfcs/0002-interfaces.html), we can finally deprecate (and, later, remove) that awful hack.

It should probably not be removed before Amaranth 0.6, since it is quite widely used.

- [x] Deprecate `amaranth.hdl.rec` in Amaranth 0.4: https://github.com/amaranth-lang/amaranth/pull/890
- [x] Make `amaranth.lib.io.Pin` an interface object usable with `connect`: https://github.com/amaranth-lang/amaranth/pull/891
- [ ] Remove `amaranth.hdl.rec` in Amaranth **0.6**

PR #891, opened 2023-09-01
lib.io: make `Pin` an interface object

whitequark:
Tracking #879.

The directions of signals in `Pin` make it convenient to use a pin signature in a component, such as in:

```python
class LEDDriver(Component):
    pins: Out(Signature({"o": Out(1)}))
    # same as `Out(Pin(1, dir="o").signature)`

led_driver = LEDDriver()
connect(led_driver.pins, platform.request("led"))
````

The `platform.request` call, correspondingly, returns a flipped `Pin` object.


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-09-04#1693847414-1693848200;) this PR on the 2023-09-04 weekly meeting. The decision was to **merge**.

PR #893, opened 2023-09-01
hdl.mem: lower `Memory` directly to `$mem_v2` RTLIL cell

whitequark:
The design decision of using split memory ports in the internal representation (copied from Yosys) was misguided and caused no end of misery. Remove any uses of `$memrd`/`$memwr` and lower memories directly to a combined memory cell, currently the RTLIL one.


whitequark:
This PR is feature-complete but requires additional testing. Testing on designs with memories is quite welcome!

crzwdjk:
Somewhat belatedly, this does work fine for me on an iCE40 UP platform.

whitequark:
Thanks!

Issue #26, opened 2019-06-15
Clean up Xilinx platform code


nmigen-issue-migration:
<a href="https://github.com/whitequark"><img src="https://avatars1.githubusercontent.com/u/54771?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Saturday Jun 15, 2019 at 16:11 GMT_
_Originally opened as https://github.com/m-labs/nmigen/pull/104_

----

@jfng Can you please review this, and make sure I didn't break anything? The main part here is the commit that removes the inverter between the register (FDCE) and tristate input of the IOB. The rest is just misc cleanups that remove intermediate signals and ensure that if a pin doesn't have e.g. the `i` signal, there is no `i` signal returned from `_get_xdr_buffer`.

----

_**[whitequark](https://github.com/whitequark)** included the following code: https://github.com/m-labs/nmigen/pull/104/commits_

PR #900, opened 2023-09-03
Add a strobe generator (rfc #26)

galibert:
Possible implementation for the proposed rfc #26


whitequark:
(Converted to a draft since a PR waiting on an RFC by definition cannot be merged. This makes it easier to parse the list of PRs.)

PR #905, opened 2023-09-09
vendor._lattice_ice40: add an icepack_opts override

crzwdjk:
Add an icepack_opts override in case the user wants to pass  extra options to icepack as part of the build process. For example can be used to pass the "-s" option to disable putting the SPI flash to sleep after the FPGA is configured.


whitequark (changes requested):
Thanks! Could you add a changelog entry please?

whitequark (approved):

PR #906, opened 2023-09-11
pyproject: migrate to PDM build backend

whitequark:
This simplifies the build process a bit, and eliminates all of the legacy packaging code in favor of fully declarative metadata (extended in a build hook that's ignored by everything except pdm).

This degrades functionality (`.git_archival.txt` is no longer usable) but maybe we could live with that until https://github.com/pdm-project/pdm-backend/issues/194 is implemented.


whitequark:
For some reason this installs Amaranth with the wrong version:

```
  ✔ Install amaranth 0.1.dev1312+gfb3c63d.editable successful
```

whitequark:
I like the consolidation of packaging functionality and the new hook enough that I think we can live without `.git_archival.txt`.

PR #908, opened 2023-09-13
Elaboratable does not need ABCMeta as its metaclass

whitequark:
This was introduced in commit 44711b7d, and was never used within Amaranth itself. While technically a breaking change I think this will not cause enough breakage to warrant a deprecation cycle (nor can we make this a deprecation this without a lot of work).

### DIFF
--- a/amaranth/hdl/ir.py
+++ b/amaranth/hdl/ir.py
@@ -20,7 +20,7 @@ class UnusedElaboratable(UnusedMustUse):
     _MustUse__silence = True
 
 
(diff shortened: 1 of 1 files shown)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-09-25#1695662288-1695661229;) this item on the 2023-09-25 weekly meeting. The conclsion was to merge and there were no objections. Any breakage (unlikely) is easily addressed.

PR #912, opened 2023-09-23
ast: fix const-castable expression handling in `Signal(reset=)`.

### LINKED ISSUE #911: Signal reset cannot be set to a const value
Consider this code

```
CYCLE_OVERFLOW = Const(742500, Shape(width=22))
audgen_accum = Signal(22, reset=CYCLE_OVERFLOW) # Master clock
```

Amaranth prints this error/exception from the build/generate step:

```
    audgen_accum = Signal(22, reset=CYCLE_OVERFLOW) # Master clock
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
TypeError: Reset value must be a constant-castable expression, not (const 22'd742500)
```

**Expected behavior:**
Consts of matching shape should be allowed as a reset value. If they are not allowed, then the error message should be changed (language "constant-castable" is unclear and sounds wrong).

I do not have an opinion on what happens if you set a signal's reset value to a const of non-matching shape.

wanda-phi:
The code to accept const-castable expressions was previously added in 0c4fda92fecb0f6b1a3dce8960f8d13459c7eef1, but it was untested and had a few bugs.

Fixes #911.

### DIFF
--- a/amaranth/hdl/ast.py
+++ b/amaranth/hdl/ast.py
@@ -1056,12 +1056,15 @@ class Signal(Value, DUID, metaclass=_SignalMeta):
                                  .format(orig_shape, Shape.cast(orig_shape),
                                          reset.shape()))
         else:
+            if reset is None:
+                reset = 0
             try:
-                reset = Const.cast(reset or 0)
+                reset = Const.cast(reset)
             except TypeError:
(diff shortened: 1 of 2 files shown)


whitequark (approved):
LGTM, thank you!

PR #929, opened 2023-10-05
ast: allow overriding Value operators.

zyp:
Proposed implementation for amaranth-lang/rfcs#28.

I'm not entirely happy with the location of the decorator implementation. The natural place would have been in `_utils` if not for the dependence on `ValueCastable`.


amaranth/hdl/ast.py
```diff
@@ -239,29 +263,38 @@ def __rrshift__(self, other):
...
     def __ror__(self, other):
         return Operator("|", [other, self])
 
+    @overridable_by_reflected("__eq__")
     def __eq__(self, other):
         return Operator("==", [self, other])
+    @overridable_by_reflected("__ne__")
     def __ne__(self, other):
         return Operator("!=", [self, other])
+    @overridable_by_reflected("__ge__")
```
wanda-phi, line 288: I think this should be `__gt__` (and similar for the next 3 operators as well

zyp: You're right, I had a brainfart and was thinking inverted result, not reflected operation. Thanks!

whitequark: I don't think this is right. Look up the `NotImplemented` value and its semantics.

zyp: The intention here is to get behavior consistent with how Python treats derived classes that overrides the operators. The derived class' implementation will be tried first and if it returns `NotImplemented`, the parent's implementation will be tried.

Example:
```python
class Left:
    def __mul__(self, other):
        print("Left.__mul__ called")

    def __rmul__(self, other):
        print("Left.__rmul__ called")

class Right(Left):
    def __mul__(self, other):
        print("Right.__mul__ called")
        return NotImplemented

    def __rmul__(self, other):
        print("Right.__rmul__ called")
        return NotImplemented

Left() * Right()

'''
$ ./test_ops.py 
Right.__rmul__ called
Left.__mul__ called
'''
```

whitequark: So why can we not use the built-in behavior? I'm concerned about implementing a chunk of what Python should be handling itself, and its correctness.

whitequark: Oh, no, I misunderstood what you're doing. I think this is fine.

zyp: Because `ValueCastable` is not a subclass of `Value`, and when the types are unrelated, the left operand will be tried first:
```python
class Left:
    def __mul__(self, other):
        print("Left.__mul__ called")

    def __rmul__(self, other):
        print("Left.__rmul__ called")

class Right:
    def __mul__(self, other):
        print("Right.__mul__ called")
        return NotImplemented

    def __rmul__(self, other):
        print("Right.__rmul__ called")
        return NotImplemented

Left() * Right()

'''
$ ./test_ops.py 
Left.__mul__ called
'''
```

whitequark: Yep, you're right. I was misremembering how the mechanism works. Apologies!

zyp:
The changes are causing two tests to fail:

```
======================================================================
FAIL: test_operators (tests.test_hdl_rec.RecordTestCase)
  File "/home/runner/work/amaranth/amaranth/tests/test_hdl_rec.py", line 288, in test_operators
    self.assertEqual(repr(s1 == r1), "(== (sig s1) (cat (sig r1__a)))")
AssertionError: '(== (cat (sig r1__a)) (sig s1))' != '(== (sig s1) (cat (sig r1__a)))'
- (== (cat (sig r1__a)) (sig s1))
+ (== (sig s1) (cat (sig r1__a)))
```
This is a natural consequence of `ValueCastable.__eq__()` now being preferred over `Value.__eq__()` and I think it's fine to just update the test case with the new expected result.

```
======================================================================
FAIL: test_bug_588 (tests.test_sim.SimulatorRegressionTestCase)
OverflowError: Value defined at /home/runner/work/amaranth/amaranth/amaranth/hdl/ast.py:172 is 4294967327 bits wide, which is unlikely to simulate in reasonable time

During handling of the above exception, another exception occurred:

  File "/home/runner/work/amaranth/amaranth/tests/test_sim.py", line 1031, in test_bug_588
    Simulator(dut)
AssertionError: "^Value defined at .+?[\\/]test_sim\.py:\d+ is 4294967327 bits wide, which is unlikely to simulate in reasonable time$" does not match "Value defined at /home/runner/work/amaranth/amaranth/amaranth/hdl/ast.py:172 is 4294967327 bits wide, which is unlikely to simulate in reasonable time"
```
This is a real problem, as `ast.py:172` is the temporary variable in the decorator that stores the result for checking against `NotImplemented` before returning, which is a useless location to report. I'm not sure yet how to fix this.

zyp:
Updated to pass the tests by updating the expected result in `test_operators` and setting `src_loc_at=1` in the decorated operators.

amaranth/hdl/ast.py
```diff
@@ -156,6 +156,23 @@ def signed(width):
     return Shape(width, signed=True)
 
 
+def overridable_by_reflected(operator):
```
whitequark, line 159: This should not be a public function as the RFC did not include a public API addition.

whitequark: Re: the suggestion to put it in `_utils`, you could take `ValueCastable` by argument of decorator.

amaranth/hdl/ast.py
```diff
@@ -156,6 +156,23 @@ def signed(width):
...
+    def decorator(f):
+        @functools.wraps(f)
+        def wrapper(self, other):
+            if isinstance(other, ValueCastable) and hasattr(other, operator):
+                res = getattr(other, operator)(self)
+                if res is not NotImplemented:
+                    return res
+            return f(self, other)
+        return wrapper
+    return decorator
```
whitequark, line 174: Two newlines between toplevel definitions.

whitequark (changes requested):

amaranth/hdl/ast.py
zyp: I just made an attempt at this. It runs into the issue that `ValueCastable` doesn't yet exist when the decorator is called, since it's defined after `Value`.

whitequark: Ah right. The current location is fine then.

whitequark:
Please add a line to the changelog (in the existing format) announcing the implementation of RFC 28.

PR #932, opened 2023-10-08
lib.wiring: fix `Component.signature` on subclasses without annotations.

wanda-phi:
On Python <3.10, classes without annotations do not get an `__annotations__` member at all, so the `getattr` on a subclass falls back to the parent class `__annotations__`, attempting to create signature members twice.  Fix that by looking at the `__dict__` instead.

### DIFF
--- a/amaranth/lib/wiring.py
+++ b/amaranth/lib/wiring.py
@@ -786,7 +786,7 @@ class Component(Elaboratable):
         cls = type(self)
         signature = Signature({})
         for base in cls.mro()[:cls.mro().index(Component)]:
-            for name, annot in getattr(base, "__annotations__", {}).items():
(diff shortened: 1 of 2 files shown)


mcclure:
Can confirm: Without this I cannot build my design, with this patch, it works

Repro for bug:
Running with:  Python 3.8.10
Code:
https://github.com/mcclure/analogue-core-template-amaranth/tree/control-test (2354a5e, or branch screen-test f055ce0)
Run instructions in README.md

Error message: amaranth.lib.wiring.SignatureError: Member 'clk' already exists in the signature and cannot be replaced

whitequark (approved):
Thanks! Yes, this is a subtle one.

Issue #875, opened 2023-08-22
Implement RFC 20: Deprecate and remove non-FWFT FIFOs


whitequark:
RFC: https://amaranth-lang.org/rfcs/0020-deprecate-non-fwft-fifos.html
- [x] Deprecate non-FWFT FIFOs: https://github.com/amaranth-lang/amaranth/pull/945
- [ ] Remove non-FWFT FIFOs: https://github.com/amaranth-lang/amaranth/pull/949
  - [ ] Changelog entry for removal

PR #945, opened 2023-10-24
Implement RFC 20: Deprecate non-FWFT FIFOs.

wanda-phi:
Tracking issue #875.


whitequark (approved):
Thanks a lot! This PR completes one of the 0.4 release milestone items, so it's really helpful ^_^

PR #954, opened 2023-11-15
FlippedInterface should propagate flipped to subinterfaces.

### LINKED ISSUE #950: Should `Signature.__getattr__` propagate `flipped` to subinterfaces?
Consider the following code:

```python
from amaranth import Module, Elaboratable
from amaranth.lib.wiring import Component, Signature, In, Out, connect, flipped

import enum


class MyAction(enum.Enum):
    DO_NOTHING = 0
    DO_STUFF = 1


ControlSignature = Signature({
    "action": Out(MyAction)
})


PeripheralSignature = Signature({
    "my_flag": Out(1),
    "ctrl": Out(ControlSignature)
})

# PeripheralSignature is flipped because Signatures are from Initiator POV.
# Peripherals are not Initiators. 
class Peripheral(Component):
    signature = PeripheralSignature.flip()

        m = Module()

        # print(Peripheral.signature.members)

        return m


class Forwarder(Component):
    periph: In(PeripheralSignature)

    def __init__(self):
        super().__init__()

        self.periph_internal = Peripheral()

        m = Module()

        m.submodules.periph_internal = self.periph_internal

        # Forward to inner peripheral.
        connect(m, self.periph_internal, flipped(self.periph))

        return m


InitiatorSignature = Signature({
    "periph": Out(ControlSignature),
})


class Initiator(Component):
    # Pretend that "my_flag" part of PeriphSignature comes from elsewhere,
    # and that it's not appropriate to expose the entire PeripheralSignature
    # to Initiator.
    signature = InitiatorSignature

    def __init__(self):
        super().__init__()

        m = Module()

        return m


class Glue(Elaboratable):
    def __init__(self):
        super().__init__()

        ###

        self.fwd = Forwarder()
        self.ini = Initiator()

        m = Module()

        m.submodules.fwd = self.fwd
        m.submodules.ini = self.ini

        print("self.fwd.signature.members:", self.fwd.signature.members)
        print("self.fwd.periph.signature.members:", self.fwd.periph.signature.members)

        print("self.fwd.periph.ctrl.signature:", self.fwd.periph.ctrl.signature)
        print("type(self.fwd.periph.ctrl.signature):", type(self.fwd.periph.ctrl.signature))

        print("self.fwd.signature.members[\"periph\"].signature.members[\"ctrl\"].signature:",
              self.fwd.signature.members["periph"].signature.members["ctrl"].signature)
        print("type(self.fwd.signature.members[\"periph\"].signature.members[\"ctrl\"].signature):",
              type(self.fwd.signature.members["periph"].signature.members["ctrl"].signature))

        # connect(m, flipped(self.fwd.periph.ctrl), self.ini.periph)
        connect(m, self.fwd.periph.ctrl, self.ini.periph)

        return m


if __name__ == "__main__":
    Glue().elaborate(None)
```

When run as-is, the `connect(m, self.fwd.periph.ctrl, self.ini.periph)` line errors out with `ConnectionError`:

```sh
self.fwd.signature.members: SignatureMembers({'periph': In(Signature({'ctrl': Out(Signature({'action': Out(<enum 'MyAction'>)})), 'my_f

lag': Out(1)}))})
self.fwd.periph.signature.members: SignatureMembers({'my_flag': Out(1), 'ctrl': Out(Signature({'action': Out(<enum 'MyAction'>)}))}).fl

ip()
self.fwd.periph.ctrl.signature: Signature({'action': Out(<enum 'MyAction'>)})
type(self.fwd.periph.ctrl.signature): <class 'amaranth.lib.wiring.Signature'>
self.fwd.signature.members["periph"].signature.members["ctrl"].signature: Signature({'action': Out(<enum 'MyAction'>)}).flip()
type(self.fwd.signature.members["periph"].signature.members["ctrl"].signature): <class 'amaranth.lib.wiring.FlippedSignature'>
  File "C:/msys64/home/William/Projects/MCVE/amaranth-flipped/flipped.py", line 107, in <module>
    Glue().elaborate(None)
  File "C:/msys64/home/William/Projects/MCVE/amaranth-flipped/flipped.py", line 101, in elaborate
    connect(m, self.fwd.periph.ctrl, self.ini.periph)
  File "C:/msys64/mingw64/lib/python3.11/site-packages/amaranth/lib/wiring.py", line 728, in connect
    raise ConnectionError(
amaranth.lib.wiring.ConnectionError: Cannot connect several output members 'arg0.action', 'arg1.action' together
```

It is required to use ` connect(m, flipped(self.fwd.periph.ctrl), self.ini.periph)` to correctly connect the `Initiator` and `Peripheral`s buses described by `ControlSignature`. I thought `self.fwd.periph.ctrl.action` should have the `In()` direction thanks to the `In()` direction of `self.fwd.periph`, and did not understand what's going on. So I added some `print` calls as above to give me some information:

```sh
self.fwd.signature.members: SignatureMembers({'periph': In(Signature({'ctrl': Out(Signature({'action': Out(<enum 'MyAction'>)})), 'my_f

lag': Out(1)}))})
self.fwd.periph.signature.members: SignatureMembers({'my_flag': Out(1), 'ctrl': Out(Signature({'action': Out(<enum 'MyAction'>)}))}).fl

ip()
```

The above two look reasonable to me; accessing `self.fwd.periph` constitutes a `flip()` thanks to `Forwarder`'s `Signature`. Consequently, I would expect access `self.fwd.periph.ctrl` to be flipped and return a `FlippedSignature`. When introspecting the subinterfaces `Signatures` via the `members` mappings, this is exactly what happens:

```
self.fwd.signature.members["periph"].signature.members["ctrl"].signature: Signature({'action': Out(<enum 'MyAction'>)}).flip()
type(self.fwd.signature.members["periph"].signature.members["ctrl"].signature): <class 'amaranth.lib.wiring.FlippedSignature'>
```

However, when access the `periph.ctrl` subinterface signature via `__getattr__`, an unflipped `Signature` is returned. `self.fwd.periph.ctrl` is the input to the failing `connect`, and consequently this is why I need `flipped` in the above snippet to make `connect` work:
```sh
self.fwd.periph.ctrl.signature: Signature({'action': Out(<enum 'MyAction'>)})
type(self.fwd.periph.ctrl.signature): <class 'amaranth.lib.wiring.Signature'>
```

_I think the `Flow`s returned by `member`s introspection and attribute accesses contradict each other, and that the `member`s introspection is correct._ But if these two ways of accessing the `Signature` members _should_ return opposite `Signature` `Flow`s<sup>1</sup>, I find the behavior incredibly unintuitive. Perhaps there should be a diagnostic that goes beyond "pepper your code with `flipped` until your components connect successfully".

## Other Context

This code worked before a90bc7b91a5a4 was merged. After [discussion](https://libera.irclog.whitequark.org/amaranth-lang/2023-10-07#1696708633-1696710343), I was told it's probably a bug in my code. After [discussing](https://libera.irclog.whitequark.org/amaranth-lang/2023-10-12#1697143652-1697146551;) with @zyp, however, I am no longer sure it's a bug in my code. So I filed this issue for clarification.

## Footnotes
1. And this necessitating the use of `flipped` to make the attribute access `Flow` match the `Flow` returned by `members`

cr1901:
Fixes #950.


amaranth/lib/wiring.py
```diff
@@ -570,6 +570,8 @@ def __getattr__(self, name):
         value = getattr(self.__unflipped, name)
         if inspect.ismethod(value):
             return types.MethodType(value.__func__, self)
+        elif isinstance(value, Interface) and name in self.__unflipped.signature.members:
```
whitequark, line 573: I don't think that's right. An interface object isn't an instance of `Interface`, it's an object with a `signature` attribute that is a signature. I think you should just check the type of the member instead.

whitequark (changes requested):

wanda-phi:
I'm not convinced a direct check for `Interface` is correct here — an interface doesn't actually have to inherit from the `Interface` class (and, for one, `FlippedInterface` doesn't), so this could break when a fancy `Signature` subclass with a custom interface class is involved.

A better check might be to check if 1) the name is in the signature, and 2) the relevant signature member is itself a signature.

cr1901:
Improved added test after feedback from @wanda-phi (tyvm!).

amaranth/lib/wiring.py
```diff
@@ -570,6 +570,9 @@ def __getattr__(self, name):
         value = getattr(self.__unflipped, name)
         if inspect.ismethod(value):
             return types.MethodType(value.__func__, self)
+        elif name in self.__unflipped.signature.members and \
```
wanda-phi, line 573: it's usually prefered to put long conditions in parentheses instead of using backslash continuation

cr1901: Should I make a new PR to add `flake8` to `dev-dependencies`, a `.flake8` file (since `flake8` does not support `pyproject.toml` and it is apparently [nontrivial](https://github.com/PyCQA/flake8/issues/234) to add) and: https://github.com/planetlabs/flake8-continuation?

whitequark: No flake8.

whitequark (approved):

PR #964, opened 2023-11-25
lib.data: implement `__eq__`, `__ne__`, `__repr__` for `View`.

wanda-phi:
There is also an alternate version of this PR for discussion purposes: #966.


whitequark:
Superseded by https://github.com/amaranth-lang/amaranth/pull/966.

PR #966, opened 2023-11-25
lib.data: implement equality for `View`, reject all other operators.

wanda-phi:
This is an alternative of #964 for discussion purposes.

### DIFF
--- a/amaranth/lib/data.py
+++ b/amaranth/lib/data.py
@@ -580,6 +580,10 @@ class View(ValueCastable):
     an Amaranth value instead of a constant integer. The returned element is chosen dynamically
     in that case.
 
+    A view can only be compared for equality with another view of the same layout,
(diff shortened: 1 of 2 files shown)


whitequark:
If accepted, this PR should also add a documentation change to `data.rst`.

whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-11-27#1701106625-1701106831;) this PR, which is a breaking change of a feature that was never released nor specified one way another in an RFC, on the 2023-11-27 weekly meeting. The disposition was to **merge**.

whitequark (approved):

whitequark (changes requested):
I think we need at least two more tests (for an arithmetic and a bitwise operator) in this PR.

Issue #790, opened 2023-05-23
Can't get simulation VCD dumps for View fields


tilk:
Using Views instead of Records has an unfortunate consequence for debugging. In Record, each of the fields had a separate Signal, which made them available for VCD dumping. In contrast, Views doesn't introduce new signals, but allows to access fragments of an existing value (including a Signal). In practice, this means that there is typically one Signal for the entire View. Unfortunately, VCD dumps display only whole Signals, and the `traces` argument for `write_vcd` handles only Signals, too. In consequence, there is no way to see View fields in VCD dumps.

whitequark:
> In practice, this means that there is typically one Signal for the entire View.

This is an expected and intended consequence of the design. (There is no way to implement View, in general, with multiple signals due owing to the existence of union views.)

> Unfortunately, VCD dumps display only whole Signals, and the `traces` argument for `write_vcd` handles only Signals, too. In consequence, there is no way to see View fields in VCD dumps.

In the short term the only way to fix this is to add a decoder for all signals created from views. Would that help?

whitequark:
I'll see how I can fit this into our development timeline and get back to you.

For now (i.e. to make sure you do not lose this functionality as Record gets removed) my suggestion is to design a wrapper that introspects the View and adds assign statements to the module corresponding to every field.

whitequark:
Effectively this issue is asking for a decoupling of the Signals in the source code and the VCD lines, which isn't something that we can easily, or without a design cycle, add right now.

tilk:
> Effectively this issue is asking for a decoupling of the Signals in the source code and the VCD lines, which isn't something that we can easily, or without a design cycle, add right now.

After inspecting the implementation of the VCD dump code, I figured as much. So, unfortunately, it looks like we need to work around this issue for now.

whitequark:
It's actually worse than just the Python VCD dump code--we need to figure out how to handle this for CXXRTL as well.

whitequark:
So I'm considering adding a hack specifically to enable migration to views to the Python simulator and dealing with it properly later, but I'm not sure yet. How invasive would a workaround be?

whitequark:
I think I'll prioritize this for 0.4 so you don't have to spend effort on a workaround.

bieganski:
the issue is real for me as well. I attach the workaround that I use, in hope that someone will find it useful.

```py
    def to_record(v: data.View) -> rec.Record:
[...]

PR #967, opened 2023-11-26
Add support for dumping structure fields in VCD

whitequark:
See #790.

This PR adds an entirely private API for describing formatting of values that is used in the standard library, in departure from our standing policy of not using private APIs in the standard library.

This is a temporary measure intended to get the version 0.4 released faster, as it has been years in the making. It is expected that this API will be made public in the version 0.5 after going through the usual RFC process.

This PR only adds VCD lines for fields defined in `lib.data.Layout` when using `sim.pysim`. The emitted RTLIL and Verilog remain the same. It is expected that when `sim.cxxsim` lands, RTLIL/Verilog output will include aliases for layout fields as well.

The value representation API also handles formatting of enumerations, with no changes visible to the designer. The implementation of `Signal(decoder=)` is changed as well to use the new API, with full backwards compatibility and no public API changes.

### DIFF
--- a/amaranth/back/rtlil.py
+++ b/amaranth/back/rtlil.py
@@ -5,7 +5,7 @@ import warnings
 import re
 
 from .._utils import bits_for, flatten
-from ..hdl import ast, ir, mem, xfrm
+from ..hdl import ast, ir, mem, xfrm, _repr
 from ..lib import wiring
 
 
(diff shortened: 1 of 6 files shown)


whitequark:
We have [discussed](https://libera.irclog.whitequark.org/amaranth-lang/2023-11-27#1701106897-1701107124;) this PR, which adds a private API between the Amaranth AST and the Amaranth standard library, on the 2023-11-27 weekly meeting. The disposition was to **merge** with the understanding that the API will go through the RFC process before Amaranth 0.5.

PR #968, opened 2023-11-26
Expand flipped object forwarding in `lib.wiring` to respect `@property` and `del`

whitequark:
Although `@property` is the most common case, any descriptors are now properly supported.

The special casing of methods goes away as they work by having functions implement the descriptor protocol. (`__get__` has some special behavior to make this possible.)

### DIFF
--- a/amaranth/lib/wiring.py
+++ b/amaranth/lib/wiring.py
@@ -491,6 +491,19 @@ class Signature(metaclass=SignatureMeta):
         return super().__repr__()
 
 
+def _gettypeattr(obj, attr):
+    # Resolve the attribute on the object's class, without triggering the descriptor protocol for
+    # attributes that are class methods, etc.
+    for cls in type(obj).__mro__:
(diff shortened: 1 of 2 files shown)


amaranth/lib/wiring.py
```diff
@@ -491,6 +491,17 @@ def __repr__(self):
         return super().__repr__()
 
 
+def _gettypeattr(obj, attr):
+    for cls in type(obj).__mro__:
+        try:
+            return cls.__dict__[attr]
+        except KeyError:
+            pass
+    # In case there is `__setattr__` on the metaclass, or just to generate an `AttributeError` with
```
wanda-phi, line 502: getattr

wanda-phi (changes requested):

amaranth/lib/wiring.py
whitequark: fixed

PR #969, opened 2023-11-26
lib.io: fix Pin.eq to work when FlippedInterface is involved.

wanda-phi:
This was broken by #915, when platform started handing out `FlippedInterface` versions of `Pin`.


whitequark:
Thanks!

PR #978, opened 2023-12-01
Implement RFC 30: Component metadata.

jfng:


amaranth/lib/meta.py
```diff
@@ -0,0 +1,109 @@
...
+        if cls.name != parsed_name:
+            raise ValueError(f"Annotation name '{cls.name}' must be obtainable from the '$id' "
+                             f"URL ('{cls.schema['$id']}'), but does not match '{parsed_name}'")
+
+    @property
+    @abstractmethod
+    def origin(self):
+        """Annotation origin.
+
+        A Python object described by this :class:`Annotation` instance.
```
whitequark, line 80:
```suggestion
        The Python object described by this :class:`Annotation` instance.
```

amaranth/lib/wiring.py
```diff
@@ -891,3 +896,173 @@ def signature(self):
...
+                            "^[A-Za-z][A-Za-z0-9_]*$": {
+                                "oneOf": [
+                                    {
+                                        "type": "object",
+                                        "properties": {
+                                            "type": {
+                                                "enum": ["port"],
+                                            },
+                                            "name": {
+                                                "type": "string",
```
whitequark, line 927: This should have a pattern `"^[A-Za-z][A-Za-z0-9_]*$"`.

amaranth/lib/meta.py
```diff
@@ -0,0 +1,109 @@
+import jsonschema
+from abc import abstractmethod, ABCMeta
+from collections.abc import Mapping
+from urllib.parse import urlparse
```
whitequark, line 4: Imports are usually put into three groups: stdlib, third party, first party. Like this:

```suggestion
from abc import abstractmethod, ABCMeta
from collections.abc import Mapping

import jsonschema
from urllib.parse import urlparse
```

jfng: Updated, except urllib which is part of stdlib.

jfng:
Update to track [changes](https://github.com/amaranth-lang/rfcs/pull/30#issuecomment-1893644341) to the RFC.

whitequark (approved):
Looks good to me. The `lib.meta` module still needs reference documentation, and `lib.wiring` module's reference documentation needs to be expanded, before the feature is considered fully done.

jfng:
Rebased on current main.

Updated `Signature.annotations(self, obj)` to `Signature.annotations(self, obj, /)`; This prevents `obj` from being set as a kwarg, which allows subclasses to rename it (e.g. to `interface`).

jfng:
> reference

Agreed.

amaranth/lib/wiring.py
```diff
@@ -705,6 +706,18 @@ def members(self):
         """
         return self.__members
 
+    def annotations(self, obj, /):
+        """Get annotations of an interface object.
+
+        Subclasses of :class:`Signature` may override this method to return annotations attached
+        to an interface object compatible with this signature.
```
whitequark, line 713:
```suggestion
        to this signature, or an interface object compatible with this signature.
```

whitequark:
I've added a task list to the tracking issue: https://github.com/amaranth-lang/amaranth/issues/1047#issue-2094460017

jfng:
Updated with documentation and changelog entry.

The CI link check will fail until the `ComponentMetadata` schema is published to https://amaranth-lang.org/schema/amaranth/0.5/component.json.

whitequark (approved):
LGTM

PR #990, opened 2023-12-06
[WIP/PoC] Add async testbench functions

zyp:


whitequark (changes requested):
So I don't think we can really commit to this approach because it implements the new interface in terms of the old one and we want to deprecate and remove the old one. (We also want the new one to have *less* overhead, not more.)

zyp:
I'm not intending to *commit* to this approach, but I think it'd be useful to prototype a simple working implementation first and write a suite of passing tests against it before starting on the «real» implementation.

whitequark:
Yeah, no objection to that, of course.

whitequark:
Thank you for this work, @zyp! PR #1344 supersedes this PR with a more complete implementation. We have ended up rewriting a lot of the simulator internals, as well as filling in the blanks in a few places where the RFC was unimplementable or underspecified.

PR #994, opened 2023-12-11
lib.wiring: preserve insertion order in `SignatureMembers.__iter__`.

jfng:
The order in which members of a `Signature` are defined can communicate developer intent, which is useful to preserve when generating documentation.


whitequark:
I am pretty sure this only does not break `connect` by chance. You need to verify that the logic in `connect`, which relies on iteration orders, is still valid.

jfng:
Updated to fix `connect` breakage and test for unordered signature members.

whitequark (approved):
Thanks!

PR #999, opened 2023-12-11
Implement RFC 38: Component signature immutability.

### LINKED ISSUE #996: Implement RFC 38: Component signature immutability
- RFC PR: https://github.com/amaranth-lang/rfcs/pull/38
- Implementation PR: N/A

wanda-phi:
Fixes #996.

### DIFF
--- a/amaranth/lib/wiring.py
+++ b/amaranth/lib/wiring.py
@@ -806,43 +806,41 @@ def connect(m, *args, **kwargs):
 
 
 class Component(Elaboratable):
-    def __init__(self):
-        for name in self.signature.members:
-            if hasattr(self, name):
-                raise NameError(f"Cannot initialize attribute for signature member {name!r} "
-                                f"because an attribute with the same name already exists")
-        self.__dict__.update(self.signature.members.create(path=()))
-
-    @property
-    def signature(self):
+    def __init__(self, signature=None):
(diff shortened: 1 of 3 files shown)


amaranth/lib/wiring.py
```diff
@@ -842,7 +834,29 @@ def signature(self):
                         raise SignatureError(f"Member '{name}' is redefined in {base.__module__}.{base.__qualname__}")
                     members[name] = annot
         if not members:
-            raise NotImplementedError(
-                f"Component '{cls.__module__}.{cls.__qualname__}' does not have signature member "
-                f"annotations")
-        return Signature(members)
+            if signature is None:
```
whitequark, line 837: The warning above will fire even in `if not members:` case. I think it can be just removed; there are legitimate reasons to have an annotation for some value-castable attribute that is not a part of the signature (maybe it's exposed for debugging and then typed).

whitequark (changes requested):
Looks good other than the warning.

whitequark (approved):

amaranth/lib/wiring.py
wanda-phi: OK, warning removed.

whitequark:
Thanks!

PR #1005, opened 2023-12-24
[WIP] Implement RFC 41: Fixed point types.

zyp:


amaranth/lib/fixed.py
```diff
@@ -0,0 +1,251 @@
...
+    @property
+    def f_width(self):
+        return self._shape.f_width
+
+    @property
+    def signed(self):
+        return self._shape.signed
+
+    @ast.ValueCastable.lowermethod
+    def as_value(self):
```
ld-cd, line 86: I think this should be:

```python
    @ast.ValueCastable.lowermethod
    def as_value(self):
        if self.signed:
            return self._target.as_signed()
        return self._target
```

or similar, otherwise multiplying two signed fixed point numbers results in an unsigned fixed point number

ld-cd: This approach does not work, it seems to break initialization of wires but I'm not quite sure how

whitequark: > otherwise multiplying two signed fixed point numbers results in an unsigned fixed point number

Why would that be the case?

ld-cd: I dug into this a bit more; I believe (with my limited understanding of the type system here), is that the signdedness of _target is given by the signdedness of the underlying raw value that the number was constructed with which is not necessarily the same as the signdedness of the fixed point type itself. [`__mul__` uses .as_value](https://github.com/zyp/amaranth/blob/7c1d8c43b7636a1bc30959ca89d70753254708b9/amaranth/lib/fixed.py#L143) and then constructs the new fixed point type based on the signdedness of the result of the multiplication when it is cast back to a fixed point type:

```python
In [19]: x = fixed.SQ(8, 8)(unsigned(17))

In [20]: x, x.as_value()
Out[20]: ((fixedpoint SQ8.8 unsigned(17)), unsigned(17))

In [21]: y = fixed.Shape.cast(x.as_value())

In [22]: y, y.signed
Out[22]: (fixed.Shape(17, 0, signed=False), False)
```

several of the other arithmetic operators do the same thing and this seems to break proper sign handling.

zyp: You're not really supposed to be able to make a `fixed.Value` that differs in width/signedness from the underlying `Value`. Expect your first line to raise an error.

ld-cd: Ok I've checked my notes, one place where I had a conversion to unsigned from signed was because of doing what you describe intentionally (a hacky attempt at implementing complex numbers in a struct and slicing the underlying unsigned value).

The other two I believe come from [here](https://github.com/amaranth-lang/amaranth/blob/7c1d8c43b7636a1bc30959ca89d70753254708b9/amaranth/lib/fixed.py#L65), and [here](https://github.com/amaranth-lang/amaranth/blob/7c1d8c43b7636a1bc30959ca89d70753254708b9/amaranth/lib/fixed.py#L183) (because `Cat` appears to be treated as unsigned?), eg:

```python
In [53]: x = Signal(fixed.SQ(8, 8))

In [54]: x << 9
Out[54]: (fixedpoint UQ18.0 (cat (const 1'd0) (sig x)))

In [55]: y = x.round(f_width = 9)

In [56]: y, y * y
Out[56]:
((fixedpoint SQ8.9 (cat (const 1'd0) (sig x))),
 (fixedpoint UQ18.18 (* (cat (const 1'd0) (sig x)) (cat (const 1'd0) (sig x)))))
```

whitequark: The signedness rules are available [for all of the operations](https://amaranth-lang.org/docs/amaranth/latest/reference.html#amaranth.hdl.Value).

PR #1010, opened 2024-01-02
build.run: add BuildPlan.execute_docker()

anuejn:
One usecase for this is using amaranth with vivado on MacOs.

### DIFF
--- a/amaranth/build/run.py
+++ b/amaranth/build/run.py
@@ -115,6 +115,28 @@ class BuildPlan:
 
         return LocalBuildProducts(build_dir)
 
+
+    def execute_local_docker(self, image, *, root="build", docker_args=[]):
+        """
+        Execute build plan inside a Docker container. Files from the build plan are placed in the 
+        build root directory ``root`` on the local filesystem. This directory is bind mounted to 
+        ``/build`` in a container and the script ``{script}.sh`` is executed inside it.
+        ``docker_args`` is a list containing additional arguments to docker.
+
+        Returns :class:`LocalBuildProducts`.
+        """
+        build_dir = self.extract(root)
+        subprocess.check_call([
+            "docker", "run", *docker_args,
+            "--rm", # remove the container after running
+            "--mount", f"type=bind,source={build_dir},target=/build",
+            "--workdir", "/build",
+            image,
+            "sh", f"{self.script}.sh",
+        ])
+        return LocalBuildProducts(build_dir)
+
+
     def execute_remote_ssh(self, *, connect_to={}, root, run_script=True):
         """
         Execute build plan using the remote SSH strategy. Files from the build
--- a/amaranth/build/run.py
+++ b/amaranth/build/run.py
@@ -5,6 +5,7 @@ import os
 import sys
 import subprocess
 import tempfile
+import warnings
 import zipfile
 import hashlib
 import pathlib
@@ -91,7 +92,7 @@ class BuildPlan:
         finally:
             os.chdir(cwd)
 
-    def execute_local(self, root="build", *, run_script=True, env=None):
+    def execute_local(self, root="build", *, run_script=None, env=None):
         """
         Execute build plan using the local strategy. Files from the build plan are placed in
         the build root directory ``root``, and, if ``run_script`` is ``True``, the script
@@ -99,10 +100,13 @@ class BuildPlan:
         is executed in the build root. If ``env`` is not ``None``, the environment is replaced
         with ``env``.
 
+        The ``run_script`` argument is deprecated. If you only want to extract the files
+        into a local folder, use the ``extract`` method.
+
         Returns :class:`LocalBuildProducts`.
         """
         build_dir = self.extract(root)
-        if run_script:
+        if run_script is None or run_script:
             if sys.platform.startswith("win32"):
                 # Without "call", "cmd /c {}.bat" will return 0.
                 # See https://stackoverflow.com/a/30736987 for a detailed explanation of why.
@@ -112,6 +116,13 @@ class BuildPlan:
             else:
                 subprocess.check_call(["sh", f"{self.script}.sh"],
                                         env=os.environ if env is None else env)
+        # TODO(amaranth-0.5): remove
+        if run_script is not None:
+            warnings.warn("The `run_script` argument is deprecated. If you only want to "
+                            "extract the files from the BuildPlan, use the .extract() method",
+                            DeprecationWarning, stacklevel=2)
+
+
 
         return LocalBuildProducts(build_dir)
 
--- a/amaranth/build/run.py
+++ b/amaranth/build/run.py
@@ -64,15 +64,10 @@ class BuildPlan:
             for filename in sorted(self.files):
                 archive.writestr(zipfile.ZipInfo(filename), self.files[filename])
 
-    def execute_local(self, root="build", *, run_script=True, env=None):
-        """
-        Execute build plan using the local strategy. Files from the build plan are placed in
-        the build root directory ``root``, and, if ``run_script`` is ``True``, the script
-        appropriate for the platform (``{script}.bat`` on Windows, ``{script}.sh`` elsewhere)
-        is executed in the build root. If ``env`` is not ``None``, the environment is replaced
-        with ``env``.
+    def extract(self, root="build"):
+        """Extracts the files from the build plan into the local build directory ``root``.
 
-        Returns :class:`LocalBuildProducts`.
+        Returns :class:`pathlib.Path`
         """
         os.makedirs(root, exist_ok=True)
         cwd = os.getcwd()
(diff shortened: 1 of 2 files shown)


amaranth/build/run.py
```diff
@@ -187,6 +187,52 @@ def mkdirs(path):
 
         return RemoteSSHBuildProducts(connect_to, root)
 
+    def execute_docker(self, root="build", *, container_image, container_workdir="/amaranth_build/",
```
whitequark, line 190: `execute_local_docker`

amaranth/build/run.py
```diff
@@ -187,6 +187,52 @@ def mkdirs(path):
...
+                       docker_args=[], run_script=True):
```
whitequark, line 191: The `run_script` argument shouldn't exist (even for `execute_local`, where we should deprecate it).

amaranth/build/run.py
```diff
@@ -187,6 +187,52 @@ def mkdirs(path):
...
+        """
+        Execute build plan inside a docker container. Files from the build plan are placed in the 
+        build root directory ``root``, on the local filesystem. if ``run_script`` is ``True``, the
+        build root directory is is bind mounted to ``/amaranth_build`` in the container and the
```
whitequark, line 195: Just `/build`.

amaranth/build/run.py
```diff
@@ -187,6 +187,52 @@ def mkdirs(path):
...
+        ``docker_args`` is a dictionary that holds all arguments to docker
+
+        Returns :class:`LocalBuildProducts`.
+        """
+        os.makedirs(root, exist_ok=True)
+        cwd = os.getcwd()
+        try:
+            os.chdir(root)
+
+            for filename, content in self.files.items():
```
whitequark, line 207: This should be factored out into its own function `extract`

whitequark (changes requested):

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+                subprocess.check_call(["cmd", "/c", f"call {self.script}.bat"],
+                                        env=os.environ if env is None else env)
+            else:
+                subprocess.check_call(["sh", f"{self.script}.sh"],
+                                        env=os.environ if env is None else env)
+
+        return LocalBuildProducts(build_dir)
+
+
+    def execute_local_docker(self, root="build", *, container_image, container_workdir="/build",
```
whitequark, line 119:
```suggestion
    def execute_local_docker(self, root="build", *, image, mountpoint="/build",
```

I'd also probably just drop the ability to customize the mounpoint.

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+                       docker_args=[]):
+        """
+        Execute build plan inside a docker container. Files from the build plan are placed in the
```
whitequark, line 122:
```suggestion
        Execute build plan inside a Docker container. Files from the build plan are placed in the 
```

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+                                        env=os.environ if env is None else env)
+
+        return LocalBuildProducts(build_dir)
+
+
+    def execute_local_docker(self, root="build", *, container_image, container_workdir="/build",
+                       docker_args=[]):
+        """
+        Execute build plan inside a docker container. Files from the build plan are placed in the 
+        build root directory ``root``, on the local filesystem. if ``run_script`` is ``True``, the
```
whitequark, line 123: No `run_script` anymore.

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+        build root directory is is bind mounted to ``/build`` in the container and the
```
whitequark, line 124: `is is`

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+
+    def execute_local_docker(self, root="build", *, container_image, container_workdir="/build",
+                       docker_args=[]):
+        """
+        Execute build plan inside a docker container. Files from the build plan are placed in the 
+        build root directory ``root``, on the local filesystem. if ``run_script`` is ``True``, the
+        build root directory is is bind mounted to ``/build`` in the container and the 
+        script ``{script}.sh`` is executed inside the container.
+
+        ``docker_args`` is a dictionary that holds all arguments to docker
```
whitequark, line 127: It's not a dictionary, it's an array. Also, I would also probably just drop it. It's a single `subprocess.check_call` invocation; if you need to customize the behavior you can just do it directly. Let's not complicate lives for ourselves. I can see people wanting to e.g. remove `--rm` and that's not something this interface allows.

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+
+        Returns :class:`LocalBuildProducts`.
+        """
+        build_dir = self.extract(root)
+        subprocess.check_call([
+            "docker", "run", *docker_args,
+            "--rm", # remove the container after running
+            "--mount", f"type=bind,source={os.getcwd()},target={container_workdir}",
```
whitequark, line 135:
```suggestion
            "--mount", f"type=bind,source={build_dir},target={container_workdir}",
```

whitequark (changes requested):

amaranth/build/run.py
```diff
@@ -92,23 +87,59 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+            "--workdir", container_workdir,
+            container_image,
+            "sh", f"{self.script}.sh",
```
whitequark, line 138:
```suggestion
            "--entrypoint={self.script}.sh",
```

amaranth/build/run.py
anuejn: for me this is necessary, since I have my vivado installation on the local disk and the most popular vivado on macos "distribution" I found does it also like this (https://github.com/ichi4096/vivado-on-silicon-mac). However, you are right, that this whole pr might be a bit superfluous now that the extract code is factored out and that might have been the proper approach all along :D.

anuejn: I am also very happy to just drop the second commit and not have any docker-specific code in amaranth

amaranth/build/run.py
anuejn: this does not work for me.
```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "--entrypoint=build_sloth.sh": executable file not found in $PATH: unknown.
```

whitequark: `"--entrypoint={mountpoint}/{self.script}.sh"` then

anuejn: this also does not work since the build scripts amaranth generates do not have a shebang

whitequark: Oh. I thought they do. That... seems a problem in its own right, actually.

whitequark: Should be fine after https://github.com/amaranth-lang/amaranth/pull/1012.

amaranth/build/run.py
```diff
@@ -92,23 +87,56 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+                subprocess.check_call(["cmd", "/c", f"call {self.script}.bat"],
+                                        env=os.environ if env is None else env)
+            else:
+                subprocess.check_call(["sh", f"{self.script}.sh"],
+                                        env=os.environ if env is None else env)
+
+        return LocalBuildProducts(build_dir)
+
+
+    def execute_local_docker(self, root="build", *, image, docker_args=[]):
```
whitequark, line 119: This signature makes a lot more sense:

```suggestion
    def execute_local_docker(self, image, *, root="build", docker_args=[]):
```

amaranth/build/run.py
```diff
@@ -64,15 +64,10 @@ def archive(self, file):
...
-        Execute build plan using the local strategy. Files from the build plan are placed in
-        the build root directory ``root``, and, if ``run_script`` is ``True``, the script
-        appropriate for the platform (``{script}.bat`` on Windows, ``{script}.sh`` elsewhere)
-        is executed in the build root. If ``env`` is not ``None``, the environment is replaced
-        with ``env``.
+    def extract(self, root="build"):
+        """Extracts the files from the build plan into the local build directory ``root``.
 
-        Returns :class:`LocalBuildProducts`.
+        Returns: str: The absolute path of the build directory
```
whitequark, line 70: If we're changing the API here to return a path, this should be a `pathlib.Path`.

amaranth/build/run.py
whitequark: All right. Let's just drop that part then.

whitequark (changes requested):
Can you deprecate `run_script` please?

amaranth/build/run.py
```diff
@@ -92,23 +88,66 @@ def execute_local(self, root="build", *, run_script=True, env=None):
...
+                # Without "call", "cmd /c {}.bat" will return 0.
+                # See https://stackoverflow.com/a/30736987 for a detailed explanation of why.
+                # Running the script manually from a command prompt is unaffected.
+                subprocess.check_call(["cmd", "/c", f"call {self.script}.bat"],
+                                        env=os.environ if env is None else env)
+            else:
+                subprocess.check_call(["sh", f"{self.script}.sh"],
+                                        env=os.environ if env is None else env)
+
+        if run_script is not None:
```
whitequark, line 120: This would help this deprecation not get lost:

```suggestion
        # TODO(amaranth-0.5): remove
        if run_script is not None:
```

whitequark (changes requested):
Looks good other than for the one remark.

whitequark (approved):

whitequark:
Thanks!

PR #1013, opened 2024-01-03
hdl.ast: fix Array not being indexable by ValueCastable

anuejn:


amaranth/hdl/ast.py
```diff
@@ -1382,6 +1382,8 @@ def __init__(self, iterable=()):
         self._mutable  = True
 
     def __getitem__(self, index):
+        if isinstance(index, ValueCastable):
```
whitequark, line 1385: You don't need this check.

whitequark (changes requested):
You should add a test.

amaranth/hdl/ast.py
anuejn: If we remove this check, we change the semantics of the method. currently if we give the method an int, we will not get an ArrayProxy. With this change, we would.

whitequark: Oh, you are right.

whitequark (approved):

PR #1015, opened 2024-01-04
docs: fix nav-bar on mobile

esden:
The side nav-bar tweaks should only be valid when the side nav bar is present. It is only present on displays wider than 769px.


whitequark (approved):

whitequark:
Thanks!

PR #1025, opened 2024-01-10
Implement RFC 17: Remove `log2_int`.

wanda-phi:
Reexports of `amaranth.utils` functions are removed from `amaranth._utils` to avoid a circular import issue (for `deprecated`). Since this is a private module, this should not be a problem.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -10,13 +10,31 @@ Version 0.5 (unreleased)
 The Migen compatibility layer has been removed.
 
 
+Migrating from version 0.4
+--------------------------
+
+Apply the following changes to code written against Amaranth 0.4 to migrate it to version 0.5:
+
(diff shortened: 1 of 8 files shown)


docs/changes.rst
```diff
@@ -10,13 +10,30 @@ Version 0.5 (unreleased)
 The Migen compatibility layer has been removed.
 
 
+Migrating from version 0.4
+--------------------------
+
+Apply the following changes to code written against Amaranth 0.4 to migrate it to version 0.5:
+
+* Update uses of :func:`amaranth.utils.log2_int` to :func:`amaranth.utils.ceil_log2` or :func:`amaranth.utils.exact_log2`
```
whitequark, line 18: Worth writing out which `need_pow2=` value each case corresponds to.

whitequark (approved):
Thanks!

whitequark:
RFC 17 links to this PR and not a tracking issue, so we have missed actually removing `log2_int`.

PR #1032, opened 2024-01-13
hdl.ir: collect source location for `Instance`.

wanda-phi:
Extracted from the new IR branch and expanded a bit, as it is useful on its own.


whitequark (approved):
LGTM

PR #1033, opened 2024-01-13
hdl.ast: raise a sensible error for `xxx in Value`

wanda-phi:


amaranth/hdl/ast.py
```diff
@@ -377,6 +377,9 @@ def __getitem__(self, key):
         else:
             raise TypeError(f"Cannot index value with {key!r}")
 
+    def __contains__(self, other):
+        raise TypeError(f"'{self.__class__.__name__}' object is not a container")
```
whitequark, line 381: Maybe something like `Cannot use 'in' with an Amaranth value`? That would match our other messages.

whitequark (approved):
LGTM, thank you!

PR #1036, opened 2024-01-16
Implement RFC 39: Warn on `Case()` and `matches()` with no patterns.

wanda-phi:


whitequark (approved):
LGTM, thank you!

whitequark:
This does seem to break `examples/basic/pmux.py` when run with `-Werror`...

PR #1054, opened 2024-01-29
hdl.ast: change warning on out-of-range reset to an error, improve it.

### LINKED ISSUE #1019: `Signal(range(0))` raises `SyntaxWarning`
`Signal(range(0))` raises 
```
SyntaxWarning: Reset value 0 equals the non-inclusive end of the signal shape range(0, 0); this is likely an off-by-one error
```
Is this warning intended? 
To me this warning is quite confusing, as amaranth implicitly converts `range(0)` (which would be something like a never type) to `unsigned(0)`, and `Signal(unsigned(0))` does not raise the same warning. 

Some more discussion can be found in the [IRC logs](https://libera.irclog.whitequark.org/amaranth-lang/2024-01-07#1704643458-1704644743;).

wanda-phi:
Fixes #1019.

### DIFF
--- a/amaranth/hdl/ast.py
+++ b/amaranth/hdl/ast.py
@@ -1170,13 +1170,14 @@ class Signal(Value, DUID, metaclass=_SignalMeta):
         self.reset = reset.value
         self.reset_less = bool(reset_less)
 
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM, thank you!

PR #1063, opened 2024-01-30
Implement RFC 42: `Const` from shape-castable.

zyp:


whitequark (approved):
Code looks good. This will require a changelog entry and reference documentation update (if I get to writing reference docs for `Const` by the point this is merged).

whitequark (approved):

PR #1076, opened 2024-01-31
lib.fifo: add `Memory` as submodules instead of its ports. [NFC]

wanda-phi:
This makes the generated netlist very slightly nicer.


whitequark (commented):
LGTM

whitequark (approved):
LGTM

PR #1077, opened 2024-02-03
Implement RFC 46: Change `Shape.cast(range(1))` to `unsigned(0)`.

wanda-phi:
See amaranth-lang/rfcs#46.

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -93,8 +93,8 @@ class Shape:
 
         * a :class:`Shape`, where the result is itself;
         * an :class:`int`, where the result is :func:`unsigned(obj) <unsigned>`;
-        * a :class:`range`, where the result is wide enough to represent any element of the range,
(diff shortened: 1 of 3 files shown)


whitequark (approved):
Thanks!

whitequark:
Do we need any documentation changes here?

wanda-phi:
> Do we need any documentation changes here?

Good question. I think the wording in lang reference could be somewhat stronger and say that it's not just "wide enough to represent any element", but is explicitly the minimal possible?

But mostly my feeling is that this was an implementation-documentation mismatch fix in the first place.

whitequark:
> I think the wording in lang reference could be somewhat stronger and say that it's not just "wide enough to represent any element", but is explicitly the minimal possible?

I think that would be a good change to include here.

wanda-phi:
Done.

whitequark (approved):
LGTM, thank you!

PR #1091, opened 2024-02-08
xilinx: use FDPE instances to implement get_async_ff_sync()

### LINKED ISSUE #721: ResetSynchronizer does not work properly with Vivado
I am attempting to use a [`ResetSynchronizer`](https://github.com/amaranth-lang/amaranth/blob/main/amaranth/lib/cdc.py#L182) with Vivado, and I am finding that the timing analysis for the implementation is not being done properly (and arguably the implementation schematic is also wrong).

My minimal example to show this is:
```python
#!/usr/bin/env python3

from amaranth import *
import amaranth.cli
from amaranth.lib.cdc import ResetSynchronizer
from amaranth.vendor.xilinx import XilinxPlatform


class Test(Elaboratable):
    def __init__(self):
        self.rst_in = Signal()
        self.toggle = Signal()
        self.rstdomain = ClockDomain()

        m = Module()
        m.domains += self.rstdomain
        rst_in_q = Signal(reset=1)
        m.submodules.sync_rst = ResetSynchronizer(rst_in_q)
        m.d.rstdomain += rst_in_q.eq(self.rst_in)
        m.d.sync += self.toggle.eq(~self.toggle)
        return m


class MyPlatform(XilinxPlatform):
    device = "xc7z010"
    package = "clg400"
    speed = "1"
    resources = []
    connectors = []


if __name__ == '__main__':
    top = Test()
    platform = MyPlatform()
    amaranth.cli.main(
        top, platform=platform,
        ports=[top.rst_in, top.toggle,
               top.rstdomain.clk, top.rstdomain.rst])
```

This has a `rstdomain` clock domain for an external reset input `rst_in`, which is registered by the flip-flop `rst_in_q`. Then a `ResetSynchronizer` is used to generate the reset of the `sync` domain, in which we find a flip-flop that keeps toggling.

The generated Verilog code is:
```verilog
/* Generated by Yosys 0.21+18 (git sha1 0ab726e20, clang 10.0.0-4ubuntu1 -fPIC -Os) */

(* \amaranth.hierarchy  = "top.sync_rst" *)
(* generator = "Amaranth" *)
module sync_rst(rst, clk, rst_in_q);
  reg \$auto$verilog_backend.cc:2083:dump_module$1  = 0;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1034" *)
  wire async_ff_clk;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1034" *)
  wire async_ff_rst;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/ir.py:527" *)
  input clk;
  wire clk;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/ir.py:527" *)
  output rst;
  wire rst;
  (* src = "/hdl/test_resetsynchronizer.py:18" *)
  input rst_in_q;
  wire rst_in_q;
  (* ASYNC_REG = "TRUE" *)
  (* \amaranth.vivado.false_path  = "TRUE" *)
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1035" *)
  reg stage0 = 1'h1;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1035" *)
  reg \stage0$next ;
  (* ASYNC_REG = "TRUE" *)
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1035" *)
  reg stage1 = 1'h1;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/vendor/xilinx.py:1035" *)
  reg \stage1$next ;
  always @(posedge async_ff_clk, posedge async_ff_rst)
    if (async_ff_rst) stage0 <= 1'h1;
    else stage0 <= \stage0$next ;
  always @(posedge async_ff_clk, posedge async_ff_rst)
    if (async_ff_rst) stage1 <= 1'h1;
    else stage1 <= \stage1$next ;
  always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$1 ) begin end
    \stage0$next  = 1'h0;
    (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/xfrm.py:519" *)
    casez (async_ff_rst)
      1'h1:
          \stage0$next  = 1'h1;
    endcase
  end
  always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$1 ) begin end
    \stage1$next  = stage0;
    (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/xfrm.py:519" *)
    casez (async_ff_rst)
      1'h1:
          \stage1$next  = 1'h1;
    endcase
  end
  assign rst = stage1;
  assign async_ff_clk = clk;
  assign async_ff_rst = rst_in_q;
endmodule

(* \amaranth.hierarchy  = "top" *)
(* top =  1  *)
(* generator = "Amaranth" *)
module top(toggle, rstdomain_clk, rstdomain_rst, clk, rst, rst_in);
  reg \$auto$verilog_backend.cc:2083:dump_module$2  = 0;
  (* src = "/hdl/test_resetsynchronizer.py:21" *)
  wire \$1 ;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/ir.py:527" *)
  input clk;
  wire clk;
  (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/ir.py:527" *)
  output rst;
  wire rst;
  (* src = "/hdl/test_resetsynchronizer.py:11" *)
  input rst_in;
  wire rst_in;
  (* src = "/hdl/test_resetsynchronizer.py:13" *)
  input rstdomain_clk;
  wire rstdomain_clk;
  (* src = "/hdl/test_resetsynchronizer.py:13" *)
  input rstdomain_rst;
  wire rstdomain_rst;
  (* src = "/hdl/test_resetsynchronizer.py:18" *)
  reg sync_rst_rst_in_q = 1'h1;
  (* src = "/hdl/test_resetsynchronizer.py:18" *)
  reg \sync_rst_rst_in_q$next ;
  (* src = "/hdl/test_resetsynchronizer.py:12" *)
  output toggle;
  reg toggle = 1'h0;
  (* src = "/hdl/test_resetsynchronizer.py:12" *)
  reg \toggle$next ;
  assign \$1  = ~ (* src = "/hdl/test_resetsynchronizer.py:21" *) toggle;
  always @(posedge rstdomain_clk)
    sync_rst_rst_in_q <= \sync_rst_rst_in_q$next ;
  always @(posedge clk)
    toggle <= \toggle$next ;
  sync_rst sync_rst (
    .clk(clk),
    .rst(rst),
    .rst_in_q(sync_rst_rst_in_q)
  );
  always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$2 ) begin end
    \sync_rst_rst_in_q$next  = rst_in;
    (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/xfrm.py:519" *)
    casez (rstdomain_rst)
      1'h1:
          \sync_rst_rst_in_q$next  = 1'h1;
    endcase
  end
  always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$2 ) begin end
    \toggle$next  = \$1 ;
    (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/xfrm.py:519" *)
    casez (rst)
      1'h1:
          \toggle$next  = 1'h0;
    endcase
  end
endmodule

```

I'm using the following XDC to perform out-of-context implementation of this Verilog in Vivado:
```
create_clock -period 5 -name clk [get_ports clk]
create_clock -period 5.01 -name rstdomain_clk [get_ports rstdomain_clk]
set_false_path -to [get_cells -hier -filter {amaranth.vivado.false_path == "TRUE"}]
```

Here is the schematic of the Vivado implementation:
vivado_schematic

Vivado is doing timing analysis of the path from `sync_rst_rst_in_q_reg/C` to `syn_rst/stage1_reg/D`, which is an inter-clock path from `rstdomain_clk` to `clk`, and thus failing timing, because I've purposely set the clock periods of `clk` and `rstdomain_clk` so that their greatest common divisor is very small.

I think that Vivado should not analyze any inter-clock paths for this example (that's the whole point of the synchronizer). It seems that the problem is that the schamatic is not really what we intended. Rather than directly connecting `stage0_reg/Q` to `stage1_reg/D`, we are getting a LUT `stage1_i_1` in the middle which has `sync_rst_rst_in_q`. This is what causes Vivado to analyze this inter-clock path (flop `stage0_reg` has a false-path constraint, but `stage1_reg` doesn't have a false-path constraint).

I think that we wouldn't like the LUT `stage1_i_1` to be there. The schematic would work well with `sync_rst_rst_in_q` only going to the `PRE` inputs of `stage0_reg` and `stage1_reg`.

It seems that Vivado is getting confused by this part of the Verilog:
```verilog
  always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$1 ) begin end
    \stage1$next  = stage0;
    (* src = "/usr/local/lib/python3.10/dist-packages/amaranth/hdl/xfrm.py:519" *)
    casez (async_ff_rst)
      1'h1:
          \stage1$next  = 1'h1;
    endcase
  end
```
Having `\stage1$next` depend on `async_ff_rst` is somewhat redundant, because we already model the asynchronous reset in this other block:
```verilog
always @(posedge async_ff_clk, posedge async_ff_rst)
    if (async_ff_rst) stage1 <= 1'h1;
    else stage1 <= \stage1$next ;
```

In fact, if I replace the `always @*` block by only
```verilog
 always @* begin
    if (\$auto$verilog_backend.cc:2083:dump_module$1 ) begin end
    \stage1$next  = stage0;
  end
```
I get the following schematic, which looks more sensible:
vivado_schematic2
However, this still gives timing failures, because Vivado is analyzing the path from `sync_rst_rst_in_q_reg/C` to `sync_rst/stage1_reg/PRE`. In this analysis, it is considering the `Recov_fdpe_C_PRE` for `stage1_reg`. I'm not sure if it is necessary to consider this recovery time for PRE, or if we can ignore it completely, so that `stage1_reg/PRE` should also have a false-path constraint to avoid analyzing this path.

daniestevez:
This closes #721 by implementing `get_async_ff_sync()` using FDPE primitives to obtain exactly the netlist that we want. This consits of a chain of `N` FPDEs (by default `N = 2`) with all their PRE pins connected to the reset for a positive edge `reset` or to the `~reset` for a negative edge reset. The D pin of the first FDPE in the chain is connected to GND.

To make timing analysis work correctly, two new attributes are introduced: `amaranth.vivado.false_path_pre` and
`amaranth.vivado.max_delay_pre`. These work similarly to `amaranth.vivado.false_path` and `amaranth.vivado.max_delay`, but affect only the PRE pin, which is what is needed for this synchronizer. The TCL has been modified to generate constraints using these attributes, and there are comments explaining how to use the attributes directly in an XDC file in case the user wants to manage their XDC file manually instead of using the TCL.

---

I have tested this with the following example.

```python
#!/usr/bin/env python3

from amaranth import *
import amaranth.cli
from amaranth.lib.cdc import ResetSynchronizer
from amaranth.vendor import XilinxPlatform


class Test(Elaboratable):
    def __init__(self):
        self.rst_in = Signal()
        self.toggle = Signal()
        self.toggle2 = Signal()
        self.rstdomain = ClockDomain()
        self.sync2 = ClockDomain()

        m = Module()
        m.domains += [self.rstdomain, self.sync2]
        rst_in_q = Signal(reset=1)
        m.submodules.sync_rst = ResetSynchronizer(rst_in_q)
        m.submodules.sync_rst_maxdel = ResetSynchronizer(
                rst_in_q, domain="sync2", max_input_delay=3e-9)
        m.d.rstdomain += rst_in_q.eq(self.rst_in)
        m.d.sync += self.toggle.eq(~self.toggle)
        m.d.sync2 += self.toggle2.eq(~self.toggle2)
        return m


class MyPlatform(XilinxPlatform):
    device = "xc7z010"
    package = "clg400"
    speed = "1"
    resources = []
    connectors = []


if __name__ == '__main__':
    top = Test()
    platform = MyPlatform()
    amaranth.cli.main(
        top, platform=platform,
        ports=[top.rst_in, top.toggle, top.toggle2,
               top.rstdomain.clk, top.rstdomain.rst,
               top.sync2.clk, top.sync2.rst])
```

These are the constraints I'm using
```tcl
create_clock -period 5 -name clk [get_ports clk]
create_clock -period 5 -name sync2_clk [get_ports sync2_clk]
create_clock -period 5.01 -name rstdomain_clk [get_ports rstdomain_clk]
set_false_path -to [get_cells -hier -filter {amaranth.vivado.false_path == "TRUE"}]

set_false_path -to [ \
  get_pins -of [get_cells -hier -filter {amaranth.vivado.false_path_pre == "TRUE"}] \
    -filter { REF_PIN_NAME == PRE } ]

set_max_delay -to [ \
  get_pins -of [get_cells -hier -filter {amaranth.vivado.max_delay_pre == "3.0"}] \
    -filter { REF_PIN_NAME == PRE } ] \
  3.0
```

The implementation schematic is as follows. Paths that are timed normally are highlighted in blue. Paths timed of a 3 ns max delay are highlighted in red. The paths that are not highlighted are not timed.

implementation_results

This looks correct to me. I haven't found detailed information about the timing characteristics of the FDPE, but my thinking is that if PRE is asserted then the output of `stage1` will go high in a short time regardless of what is happening with the clock. If PRE is deasserted, the worst that can happen is that this happens close to a clock edge. In this case I think that the output of `stage0` might go metastable, but the output of `stage1` should be a stable 1, because when this happens, `stage1` has a stable 1 at its input. At the next clock edge, hopefully the metastability of `stage0` should have resolved to a stable 0 or a stable 1 (and if not, use a longer chain). If it has resolved to a 0, then the output of `stage1` will pick the 0 in this clock cycle. If it has resolved to a 1, then `stage0` will pick the 0 in this clock cycle, and `stage1` will pick it in the next clock cycle.

Besides testing this example with the constraints shown above, I have deleted the `set_false_path` and `set_max_delay` constraints and checked that running the TCL commands that I have added generates the correct constraints.

A question to make sure I understand what `async_edge="neg"` means: my interpretation is that when the input is 0, then the output should be 1 immediately. When the input is 1, then the output should be 0 at the next clock edge. Is this right?

By the way, now that I have given some more thought to this reset synchronizer, I wonder how useful it is in a typical Xilinx design. Flip-flops are normally synchronous reset, so if the clock is off, they won't reset even if the reset is asserted (I guess that the main use case of `AsyncFFSynchronizer` instead of `FFSynchronizer` is to be able to assert the output even when there is no clock).

### DIFF
--- a/amaranth/vendor/_xilinx.py
+++ b/amaranth/vendor/_xilinx.py
@@ -172,6 +172,11 @@ class XilinxPlatform(TemplatedPlatform):
             foreach cell [get_cells -quiet -hier -filter {amaranth.vivado.false_path == "TRUE"}] {
                 set_false_path -to $cell
             }
+            foreach pin [get_pins -of \
+                             [get_cells -quiet -hier -filter {amaranth.vivado.false_path_pre == "TRUE"}] \
+                             -filter {REF_PIN_NAME == PRE}] {
+                set_false_path -to $pin
+            }
             foreach cell [get_cells -quiet -hier -filter {amaranth.vivado.max_delay != ""}] {
                 set clock [get_clocks -of_objects \
                     [all_fanin -flat -startpoints_only [get_pin $cell/D]]]
@@ -179,6 +184,11 @@ class XilinxPlatform(TemplatedPlatform):
                     set_max_delay -datapath_only -from $clock \
                         -to [get_cells $cell] [get_property amaranth.vivado.max_delay $cell]
                 }
+            }
+            foreach cell [get_cells -quiet -hier -filter {amaranth.vivado.max_delay_pre != ""}] {
+                set_max_delay \
+                    -to [get_pins -of [get_cells $cell] -filter {REF_PIN_NAME == PRE}] \
+                    [get_property amaranth.vivado.max_delay_pre $cell]
             }
             {{get_override("script_after_synth")|default("# (script_after_synth placeholder)")}}
             report_timing_summary -file {{name}}_timing_synth.rpt
@@ -1191,29 +1201,53 @@ class XilinxPlatform(TemplatedPlatform):
     def get_async_ff_sync(self, async_ff_sync):
         m = Module()
         m.domains += ClockDomain("async_ff", async_reset=True, local=True)
-        flops = [Signal(1, name=f"stage{index}", reset=1,
-                        attrs={"ASYNC_REG": "TRUE"})
-                 for index in range(async_ff_sync._stages)]
-        if self.toolchain == "Vivado":
-            if async_ff_sync._max_input_delay is None:
-                flops[0].attrs["amaranth.vivado.false_path"] = "TRUE"
-            else:
-                flops[0].attrs["amaranth.vivado.max_delay"] = str(async_ff_sync._max_input_delay * 1e9)
-        elif async_ff_sync._max_input_delay is not None:
-            raise NotImplementedError("Platform '{}' does not support constraining input delay "
-                                      "for AsyncFFSynchronizer"
-                                      .format(type(self).__name__))
-        for i, o in zip((0, *flops), flops):
-            m.d.async_ff += o.eq(i)
+        # Instantiate a chain of async_ff_sync._stages FDPEs with all
+        # their PRE pins connected to either async_ff_sync.i or
+        # ~async_ff_sync.i. The D of the first FDPE in the chain is
+        # connected to GND.
+        flops_q = Signal(async_ff_sync._stages, reset_less=True)
+        flops_d = Signal(async_ff_sync._stages, reset_less=True)
+        flops_pre = Signal(reset_less=True)
+        for i in range(async_ff_sync._stages):
+            flop = Instance("FDPE", p_INIT=1, o_Q=flops_q[i],
+                            i_C=ClockSignal(async_ff_sync._o_domain),
+                            i_CE=Const(1), i_PRE=flops_pre, i_D=flops_d[i],
+                            a_ASYNC_REG="TRUE")
+            m.submodules[f"stage{i}"] = flop
+            if self.toolchain == "Vivado":
+                if async_ff_sync._max_input_delay is None:
+                    # This attribute should be used with a constraint of the form
+                    #
+                    # set_false_path -to [ \
+                    #   get_pins -of [get_cells -hier -filter {amaranth.vivado.false_path_pre == "TRUE"}] \
+                    #     -filter { REF_PIN_NAME == PRE } ]
+                    #
+                    flop.attrs["amaranth.vivado.false_path_pre"] = "TRUE"
+                else:
+                    # This attributed should be used with a constraint of the form
+                    #
+                    # set_max_delay -to [ \
+                    #   get_pins -of [get_cells -hier -filter {amaranth.vivado.max_delay_pre == "3.0"}] \
+                    #     -filter { REF_PIN_NAME == PRE } ] \
+                    #   3.0
+                    #
+                    # A different constraint must be added for each different _max_input_delay value
+                    # used. The same value should be used in the second parameter of set_max_delay
+                    # and in the -filter.
+                    flop.attrs["amaranth.vivado.max_delay_pre"] = str(async_ff_sync._max_input_delay * 1e9)
+            elif async_ff_sync._max_input_delay is not None:
+                raise NotImplementedError("Platform '{}' does not support constraining input delay "
+                                          "for AsyncFFSynchronizer"
+                                          .format(type(self).__name__))
+
+        for i, o in zip((0, *flops_q), flops_d):
+            m.d.comb += o.eq(i)
 
         if async_ff_sync._edge == "pos":
-            m.d.comb += ResetSignal("async_ff").eq(async_ff_sync.i)
+            m.d.comb += flops_pre.eq(async_ff_sync.i)
         else:
-            m.d.comb += ResetSignal("async_ff").eq(~async_ff_sync.i)
+            m.d.comb += flops_pre.eq(~async_ff_sync.i)
 
-        m.d.comb += [
-            ClockSignal("async_ff").eq(ClockSignal(async_ff_sync._o_domain)),
-            async_ff_sync.o.eq(flops[-1])
-        ]
+        m.d.comb += async_ff_sync.o.eq(flops_q[-1])
 
         return m


whitequark:
> (I guess that the main use case of `AsyncFFSynchronizer` instead of `FFSynchronizer` is to be able to assert the output even when there is no clock).

Yes. It's only useful in a situation where a clock domain has async reset.

amaranth/vendor/_xilinx.py
```diff
@@ -1191,29 +1201,56 @@ def get_ff_sync(self, ff_sync):
...
+        flops_d = Signal(async_ff_sync._stages, reset_less=True)
+        flops_pre = Signal(reset_less=True)
+        for i in range(async_ff_sync._stages):
+            flops = [Instance("FDPE", p_INIT=1, o_Q=flops_q[i],
+                              i_C=ClockSignal(async_ff_sync._o_domain),
+                              i_CE=Const(1), i_PRE=flops_pre, i_D=flops_d[i],
+                              a_ASYNC_REG="TRUE")
+                     for i in range(async_ff_sync._stages)]
+        for i, f in enumerate(flops):
+            setattr(m.submodules, f"stage{i}", f)
```
whitequark, line 1218:
```suggestion
            m.submodules[f"stage{i}"] = f
```

daniestevez: Committing suggestion via the web UI and then I'll squash when the review is done.

amaranth/vendor/_xilinx.py
```diff
@@ -1191,29 +1201,56 @@ def get_ff_sync(self, ff_sync):
...
+        # connected to GND.
+        flops_q = Signal(async_ff_sync._stages, reset_less=True)
+        flops_d = Signal(async_ff_sync._stages, reset_less=True)
+        flops_pre = Signal(reset_less=True)
+        for i in range(async_ff_sync._stages):
+            flops = [Instance("FDPE", p_INIT=1, o_Q=flops_q[i],
+                              i_C=ClockSignal(async_ff_sync._o_domain),
+                              i_CE=Const(1), i_PRE=flops_pre, i_D=flops_d[i],
+                              a_ASYNC_REG="TRUE")
+                     for i in range(async_ff_sync._stages)]
```
whitequark, line 1216: I think the outer `for` isn't doing anything?

whitequark: I think the three `for` loops here should probably be combined into one.

daniestevez: Yeah. Silly me. This is a leftover of how I was rewriting the code. I'll fix this.

daniestevez: Fixed and force-pushed as a single commit.

whitequark (approved):
Fantastic, thank you!

PR #1093, opened 2024-02-08
Implement RFC 27 amendment: deprecate `add_sync_process`, not `add_process`.

wanda-phi:


whitequark (changes requested):
I think this is insufficient since we ban `yield Settle()` in `add_process`: https://amaranth-lang.org/rfcs/0027-simulator-testbenches.html#guidance-on-simulator-modalities

wanda-phi:
> I think this is insufficient since we ban `yield Settle()` in `add_process`: https://amaranth-lang.org/rfcs/0027-simulator-testbenches.html#guidance-on-simulator-modalities

I don't quite understand? Using `Settle` in `add_process` raises a deprecation warning right on `Settle` construction.

whitequark:
> I don't quite understand? Using `Settle` in `add_process` raises a deprecation warning right on `Settle` construction.

Oh sorry, I completely forgot that it's entirely deprecated, that's 100% on me.

whitequark (approved):
Thank you!

PR #1094, opened 2024-02-09
hdl.ir: associate statements with domains.

### LINKED ISSUE #1079: Refactor `Fragment` to have a mapping from statements to domains
Currently, `Fragment` has a mapping from signals to domains that drive them. This has two issues:
- `Property` has to maintain an internal `Signal` solely so that its identity is not lost during fragment transformation.
- `Signal` can only be in one domain at a time.

Because a `Signal` can be driven from two different domains in two different modules, checking this just at module/fragment construction time is insufficient and it has to be explicitly checked at elaboration time as well. (The new IR builder takes care of this.)

Getting rid of `drivers` dictionary and only tracking assignments is necessary for `Property` to have a reasonable design.

wanda-phi:
Fixes #1079.

This does not remove `Fragment.drivers` — that will be done when the new IR lands and the remaining uses in `hdl._ir` can be killed.

This also does not remove the special signals inside `Property` — they are still used within RTLIL backend, and likewise will be removed when the new IR lands.

### DIFF
--- a/amaranth/back/rtlil.py
+++ b/amaranth/back/rtlil.py
@@ -667,6 +667,7 @@ class _StatementCompiler(_xfrm.StatementVisitor):
         self.rhs_compiler = rhs_compiler
         self.lhs_compiler = lhs_compiler
 
+        self._domain      = None
         self._case        = None
         self._test_cache  = {}
         self._has_rhs     = False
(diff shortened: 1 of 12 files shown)


whitequark (approved):
Excellent work!

PR #1095, opened 2024-02-09
hdl.ast: make it impossible to construct `*Castable` instances.

### LINKED ISSUE #1072: `ShapeCastable()` is a valid expression
This shouldn't be possible to do since it's an abstract class.

wanda-phi:
Fixes #1072.


whitequark (approved):
LGTM

PR #1109, opened 2024-02-12
hdl.ir: add IOBufferInstance.

wanda-phi:


whitequark:
We can change `back.rtlil` to escape leading `$` after this PR, right?

wanda-phi:
> We can change `back.rtlil` to escape leading `$` after this PR, right?

We can change it right now, even. The PRs that got rid of `$dff` already enabled it, as the `$tribuf`s don't reach the RTLIL backend anyway.

whitequark:
> We can change it right now, even.

Let's do that then. This may break some downstream code but I'm perfectly content with that.

amaranth/hdl/_ir.py
```diff
@@ -460,6 +460,36 @@ def __init__(self, type, *args, src_loc=None, src_loc_at=0, **kwargs):
...
+        else:
+            self.o = _ast.Value.cast(o)
+            if len(self.pad) != len(self.o):
+                raise ValueError(f"`pad` length ({len(self.pad)}) doesn't match `o` length ({len(self.o)})")
+            if oe is None:
+                self.oe = _ast.Const(1)
+            else:
+                self.oe = _ast.Value.cast(oe)
+                if len(self.oe) != 1:
+                    raise ValueError(f"`oe` length ({len(self.oe)}) has to be equal to 1")
```
whitequark, line 488:
```suggestion
                    raise ValueError(f"`oe` length ({len(self.oe)}) must be 1")
```

whitequark: (Also, I thought we were using `'single quotes'` for this?)

amaranth/hdl/_xfrm.py
```diff
@@ -277,6 +277,23 @@ def on_fragment(self, fragment):
             new_fragment = Instance(fragment.type, src_loc=fragment.src_loc)
             new_fragment.parameters = OrderedDict(fragment.parameters)
             self.map_named_ports(fragment, new_fragment)
+        elif isinstance(fragment, IOBufferInstance):
+            if hasattr(self, "on_value"):
```
whitequark, line 281: This is criminal, but really it's just more motivation to get rid of transformers.

whitequark (commented):
Looks good other than inline comments.

amaranth/hdl/_xfrm.py
wanda-phi: I mean, we do the exact same thing for `Instance` and `MemoryInstance`...

whitequark: It's criminal there too!

whitequark (approved):
LGTM, thanks!

PR #1113, opened 2024-02-13
Add reference documentation for `Value`, `ValueCastable`, `ValueLike`

whitequark:
Except `Value.matches`, `Value.__getitem__`, and `Value.implies`.

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -36,7 +36,7 @@ class DUID:
 
 
 class Shape:
-    """Bit width and signedness of a value.
+    """Bit width and signedness of a :class:`Value`.
 
     A :class:`Shape` can be obtained by:
 
@@ -326,7 +326,7 @@ class _ShapeLikeMeta(type):
 
 @final
 class ShapeLike(metaclass=_ShapeLikeMeta):
-    """An abstract class representing all objects that can be cast to a :class:`Shape`.
+    """Abstract class representing all objects that can be cast to a :class:`Shape`.
 
     :pc:`issubclass(cls, ShapeLike)` returns :pc:`True` for:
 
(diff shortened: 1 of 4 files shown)


amaranth/hdl/_ast.py
```diff
@@ -372,13 +372,54 @@ def wrapper(self, other):
...
+        * an :class:`enum.IntEnum` instance, or a :class:`enum.Enum` instance whose members are
+          all integers, where the result is a :class:`Const(obj, enum_shape)` where :pc:`enum_shape`
+          is a shape that can represent every member of the enumeration;
+        * a :class:`ValueCastable` instance, where the result is obtained by repeatedly calling
+          :meth:`obj.as_value() <ValueCastable.as_value>`.
+
+        Raises
+        ------
+        TypeError
+            If :pc:`obj` cannot be converted to a :class:`Shape`.
```
wanda-phi, line 432:
```suggestion
            If :pc:`obj` cannot be converted to a :class:`Value`.
```

amaranth/hdl/_ast.py
```diff
@@ -372,13 +372,54 @@ def wrapper(self, other):
...
+        RecursionError
+            If :pc:`obj` is a :class:`ShapeCastable` object that casts to itself.
```
wanda-phi, line 434:
```suggestion
            If :pc:`obj` is a :class:`ValueCastable` object that casts to itself.
```

amaranth/hdl/_ast.py
```diff
@@ -399,222 +440,595 @@ def __init__(self, *, src_loc_at=0):
         super().__init__()
         self.src_loc = tracer.get_src_loc(1 + src_loc_at)
 
+    @abstractmethod
+    def shape(self):
+        """Shape of :pc:`self`.
+
+        Returns
+        -------
+        :ref:`shape-like object <lang-shapelike>`
```
wanda-phi, line 461: This always returns an actual `Shape`; do we want to leave an option to return a `ShapeCastable`?

amaranth/hdl/_ast.py
```diff
@@ -399,222 +440,595 @@ def __init__(self, *, src_loc_at=0):
...
+        return Operator(">>", [other, self])
+
+    def shift_right(self, amount):
+        """Right shift by constant amount.
+
+        If :pc:`amount < 0`, performs the same operation as :pc:`self.left_right(-amount)`.
+
+        Returns
+        -------
+        :class:`Value`, :pc:`unsigned(len(self) - amount)`
```
wanda-phi, line 961:
```suggestion
        :class:`Value`, :pc:`unsigned(max(len(self) - amount, 0))`
```

amaranth/hdl/_ast.py
```diff
@@ -399,222 +440,595 @@ def __init__(self, *, src_loc_at=0):
...
+            If :pc:`self` is unsigned.
+        :class:`Value`, :pc:`signed(len(self) - amount)`
```
wanda-phi, line 963: ... uhhhh

```
>>> Signal(signed(10).shift_right(11)).shape()
AttributeError: 'Shape' object has no attribute 'shift_right'
```

wanda-phi (changes requested):

amaranth/hdl/_ast.py
whitequark: Um, `signed(10)` is not a value?

whitequark: I get

```
>>> Signal(signed(10)).shift_right(11).shape()
  File "/home/whitequark/Projects/amaranth/amaranth/hdl/_ast.py", line 1509, in shape
    return Shape(a_shape.width, True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/whitequark/Projects/amaranth/amaranth/hdl/_ast.py", line 63, in __init__
    raise TypeError(f"Width of a signed value must be a positive integer, not {width}")
TypeError: Width of a signed value must be a positive integer, not 0
```

which does seem vaguely like a bug, or underspecification at least.

amaranth/hdl/_ast.py
```diff
@@ -399,222 +452,615 @@ def __init__(self, *, src_loc_at=0):
...
+    def shift_right(self, amount):
+        """Right shift by constant amount.
+
+        If :pc:`amount < 0`, performs the same operation as :pc:`self.left_right(-amount)`.
+
+        Returns
+        -------
+        :class:`Value`, :pc:`unsigned(max(len(self) - amount, 0))`
+            If :pc:`self` is unsigned.
+        :class:`Value`, :pc:`signed(len(self) - amount)`
```
wanda-phi, line 963: After the upcoming bug fix, this will be:

```suggestion
        :class:`Value`, :pc:`signed(max(len(self) - amount, 1))`
```

amaranth/hdl/_ast.py
```diff
@@ -399,222 +452,615 @@ def __init__(self, *, src_loc_at=0):
...
-            ``1`` if an odd number of bits are set, ``0`` if an even number of bits are set.
-        """
-        return Operator("r^", [self])
+    def __contains__(self, other):
+        """Forbidden membership test operator.
 
-    def implies(premise, conclusion):
-        """Implication.
+        Python requires this operator to return a :class:`bool`. Since this is not possible
+        for Amaranth values, this operator always raises an exception.
```
wanda-phi, line 1031: We could add a note (here and/or in the error message?) that `.matches()` provides equivalent functionality (and then some) if you want to check membership in a constant set.

whitequark: Oh, good catch!

amaranth/hdl/_ast.py
```diff
@@ -399,222 +452,615 @@ def __init__(self, *, src_loc_at=0):
...
+        raise TypeError("Cannot use 'in' with an Amaranth value")
 
     def bit_select(self, offset, width):
         """Part-select with bit granularity.
 
-        Selects a constant width but variable offset part of a ``Value``, such that successive
-        parts overlap by all but 1 bit.
+        Selects a constant width, variable offset part of :pc:`self`, where parts with successive
+        offsets overlap by :pc:`width - 1` bits. When :pc:`offset` is a constant integer, this
+        operation is equivalent to :pc:`self[offset:offset + width]`.
```
wanda-phi, line 1046: Needs a note that offset must be unsigned.

amaranth/hdl/_ast.py
```diff
@@ -624,41 +1070,67 @@ def bit_select(self, offset, width):
     def word_select(self, offset, width):
         """Part-select with word granularity.
 
-        Selects a constant width but variable offset part of a ``Value``, such that successive
-        parts do not overlap.
+        Selects a constant width, variable offset part of :pc:`self`, where parts with successive
+        offsets are adjacent but do not overlap. When :pc:`offset` is a constant integer, this
+        operation is equivalent to :pc:`self[offset * width:offset * (width + 1)]`.
```
wanda-phi, line 1077:
```suggestion
        operation is equivalent to :pc:`self[offset * width:(offset + 1) * width]`.
```

amaranth/hdl/_ast.py
whitequark: I added a comment about this. We definitely should decide whether this is useful or not and commit to doing it in `Signal` or whatever or never doing it at all.

wanda-phi (approved):

PR #1121, opened 2024-02-13
vendor.lattice_ice40: use `SB_DFF` instead of `$dff`.

wanda-phi:


whitequark (approved):
This is even shorter, lol

PR #1122, opened 2024-02-13
vendor._intel: use `dff` instead of `$dff`.

### LINKED ISSUE #1046: Don't instantiate `$dff`
That's a Yosys-internal cell that we only use in `vendor/` and can replace with vendor-specific primitives. New IR will benefit from this fix.
- [x] `vendor._lattice_ice40` PR: https://github.com/amaranth-lang/amaranth/pull/1121
- [ ] `vendor._intel` PR: https://github.com/amaranth-lang/amaranth/pull/1122

wanda-phi:
Fixes #1046.


whitequark (approved):

whitequark:
This was broken before so I'll just merge it.

PR #1129, opened 2024-02-13
Remember origins of a fragment during elaboration

whitequark:
This isn't expected to result in a significant increase in memory use, so for now it's enabled by default. Elaboration chains where it is not desired to preserve origins can delete the `_origins` attribute from the fragment and nothing will be stored.

The interface `Fragment._origins` remains private, as is the rest of the `Fragment` interface (including itself), but it enables certain codebases that currently use a much more invasive technique to rely on reading a single private field.

If this interface proves to be sufficient and useful, it will graduate in a largely unmodified form into a stable interface on `Fragment`'s successor.

### DIFF
--- a/amaranth/hdl/_ir.py
+++ b/amaranth/hdl/_ir.py
@@ -31,8 +31,11 @@ class Fragment:
     @staticmethod
     def get(obj, platform):
         code = None
+        origins = []
         while True:
             if isinstance(obj, Fragment):
+                if hasattr(obj, "origins"):
+                    obj.origins = tuple(origins)
                 return obj
             elif isinstance(obj, Elaboratable):
                 code = obj.elaborate.__code__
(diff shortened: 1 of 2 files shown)


anuejn:
I really appreciate this, we also have [a monkey patched hack](https://github.com/apertus-open-source-cinema/naps/blob/main/naps/soc/tracing_elaborate.py) to have this functionality downstream. Happy to see this 🎉

whitequark:
@anuejn Good to know! Waiting on @tilk or @lekcyjna123 to give feedback before merging this now.

tilk:
A heads-up: I'm currently away and the soonest date I will be able to look at this is Monday.

lekcyjna123:
Hi,
I have checked the code and did some small experiments and in my opinion it will fit very well in our tracing code. Thanks!

PR #1142, opened 2024-02-14
Implement RFC 45: Move `hdl.Memory` to `lib.Memory`.

### LINKED ISSUE #216: Memory port transparency model is flawed
<a href="https://github.com/whitequark"></img></a> **Issue by [whitequark](https://github.com/whitequark)**
_Friday Sep 20, 2019 at 18:04 GMT_
_Originally opened as https://github.com/m-labs/nmigen/issues/216_

----

Investigating #172, it appears that there are two major issues with the way port transparency is handled.
  1. Transparent read ports should not be the default, since it is expensive. On many platforms (e.g. iCE40), fabric logic will need to be inserted. On other platforms (e.g. Xilinx), non-transparent ports have a much larger set of configurations with well-defined behavior. Transparency between different clock domains is impossible with BRAMs and will result in a (silent) failure during synthesis.
  2. Transparency should not be requested *in general*, but rather for a specific write port (in the same domain). It does not make sense to request a "fully" transparent read port for a true dual-port RAM with write ports in different domains. It does make sense to request a read port transparent only wrt one specific write port for a true dual-port RAM with write ports in the same domain (this is implementable in Xilinx) and a read port transparent wrt both write ports for a true dual-port RAM with write ports in the same domain (this is not directly implementable in any FPGA arch I know, but is useful for e.g. superscalar CPU register files).

I propose to remove `transparent` completely, replacing it with a `transparent_for=[write_ports]` argument. The default (when this argument is not specified) would be "transparent for no write ports" for synchronous read ports, and "transparent for all ports" for asynchronous read ports. It would not be possible to specify a non-default value for an asynchronous read port.

The lowering to RTLIL would be the same as today, with the selected write port ignored (beyond checking that the domain is compatible),

wanda-phi:
Fixes #216.

### DIFF
--- a/amaranth/lib/memory.py
+++ b/amaranth/lib/memory.py
@@ -1,0 +1,430 @@
+import operator
+from collections import OrderedDict
+from collections.abc import MutableSequence
+
+from ..hdl import MemoryIdentity, MemoryInstance, Shape, ShapeCastable, Const
+from ..hdl._mem import MemorySimRead
+from ..utils import ceil_log2
+from .data import ArrayLayout
+from . import wiring
+from .. import tracer
+
+
+__all__ = ["WritePort", "ReadPort", "Memory"]
+
+
+class WritePort:
+    """A memory write port.
+
+    Parameters
+    ----------
+    signature : :class:`WritePort.Signature`
+        The signature of the port.
+    memory : :class:`Memory` or ``None``
+        Memory associated with the port.
+    domain : str
+        Clock domain. Defaults to ``"sync"``. Writes have a latency of 1 clock cycle.
+
+    Attributes
+    ----------
+    signature : :class:`WritePort.Signature`
+    memory : :class:`Memory`
+    domain : str
+    """
+
+    class Signature(wiring.Signature):
+        """A signature of a write port.
+
+        Parameters
+        ----------
+        addr_width : int
+            Address width in bits. If the port is associated with a :class:`Memory`,
+            it must be equal to :py:`ceil_log2(memory.depth)`.
+        shape : :ref:`shape-like <lang-shapelike>` object
+            The shape of the port data. If the port is associated with a :class:`Memory`,
+            it must be equal to its element shape.
+        granularity : int or ``None``
+            Port granularity. If ``None``, the entire storage element is written at once.
+            Otherwise, determines the size of access covered by a single bit of ``en``.
+            One of the following must hold:
+
+            - ``granularity is None``, in which case ``en_width == 1``, or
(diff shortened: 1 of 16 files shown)


amaranth/lib/memory.py
```diff
@@ -0,0 +1,430 @@
...
+        
+        def __repr__(self):
+            return f"Memory.Init({self._items!r})"
+
+    def __init__(self, *, depth, shape, init, attrs=None, src_loc_at=0, src_loc=None):
+        # shape and depth validation performed in Memory.Init constructor.
+        self._depth = depth
+        self._shape = shape
+        self._init = Memory.Init(init, shape=shape, depth=depth)
+        self._attrs = OrderedDict(() if attrs is None else attrs)
```
whitequark, line 365: `OrderedDict` is just a normal dict since Python 3.6.

amaranth/lib/memory.py
```diff
@@ -0,0 +1,430 @@
...
+        self.src_loc = src_loc or tracer.get_src_loc(src_loc_at=1 + src_loc_at)
```
whitequark, line 366: Is this the right `src_loc_at`?

amaranth/lib/memory.py
```diff
@@ -0,0 +1,430 @@
...
+                if shape.length == 0:
+                    en_width = 0
+                elif granularity == 0:
+                    raise ValueError("Granularity must be positive")
+                elif shape.length % granularity != 0:
+                    raise ValueError("Granularity must divide data array length")
+                else:
+                    en_width = shape.length // granularity
+            else:
+                raise TypeError("Granularity can only be specified for plain `Shape` or `ArrayLayout`")
```
whitequark, line 94: `unsigned`, not just plain `Shape`, right?

whitequark (commented):
Did we decide to get rid of `MemoryIdentity` in the end?

wanda-phi:
> Did we decide to get rid of `MemoryIdentity` in the end?

I ended up including it, since this allows us to recognize memories of all kinds (old and new) as "something that has an `_identity` that is `MemoryInstance`" in the simulator, avoiding a dependency.

Still not sure if this is the way to go.

amaranth/lib/memory.py
wanda-phi: whoops, good catch

whitequark (commented):
Can you back out the docs rst file? I'd like to improve these before publishing, but I don't want to block the PR on that.

wanda-phi:
> Can you back out the docs rst file? I'd like to improve these before publishing, but I don't want to block the PR on that.

Done

whitequark (approved):
Thanks!

PR #1145, opened 2024-02-16
utils: backport `ceil_log2` and `exact_log2`. [0.4 backport]

wanda-phi:
No deprecation or other changes — just transplanting the functions raw from current main for minimal disruption.


whitequark (approved):
I'm not actually sure that we should be backporting features in general, but this case in particular seems harmless so I'll approve it as a judgement call.

PR #1146, opened 2024-02-16
hdl._ast: make `Shape` immutable and hashable.

### LINKED ISSUE #1127: Make `Shape` immutable and hashable
It has never been intended that a `Shape` could be mutated.

wanda-phi:
Fixes #1127.


whitequark (approved):
LGTM, thank you!

PR #1147, opened 2024-02-16
hdl, back.rtlil: track and emit module locations.

### LINKED ISSUE #1130: Attach `src` attributes to emitted `module`s and `cell`s
This isn't currently done in Amaranth (neither is it tracked) and with the [`$scopeinfo` support](https://github.com/YosysHQ/yosys/pull/4084) in Yosys this is newly relevant for simulation.

wanda-phi:
Fixes #1130

### DIFF
--- a/amaranth/hdl/_dsl.py
+++ b/amaranth/hdl/_dsl.py
@@ -82,15 +82,18 @@ class _ModuleBuilderSubmodules:
         object.__setattr__(self, "_builder", builder)
 
     def __iadd__(self, modules):
+        src_loc = tracer.get_src_loc()
         for module in flatten([modules]):
-            self._builder._add_submodule(module)
+            self._builder._add_submodule(module, src_loc=src_loc)
         return self
 
     def __setattr__(self, name, submodule):
-        self._builder._add_submodule(submodule, name)
+        src_loc = tracer.get_src_loc()
+        self._builder._add_submodule(submodule, name, src_loc=src_loc)
 
-    def __setitem__(self, name, value):
-        return self.__setattr__(name, value)
+    def __setitem__(self, name, submodule):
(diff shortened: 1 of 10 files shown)


amaranth/hdl/_dsl.py
```diff
@@ -546,20 +550,20 @@ def _add_statement(self, assigns, domain, depth):
...
         else:
             if name in self._named_submodules:
                 raise NameError(f"Submodule named '{name}' already exists")
-            self._named_submodules[name] = submodule
+            self._named_submodules[name] = (submodule, src_loc)
 
     def _get_submodule(self, name):
         if name in self._named_submodules:
-            return self._named_submodules[name]
+            return self._named_submodules[name][0]
```
whitequark, line 566: Please use destructuring here, never tuple indexing.

amaranth/hdl/_nir.py
```diff
@@ -211,9 +211,9 @@ def __repr__(self):
             result.append(f"(cell {cell_idx} {cell.module_idx} {cell!r})")
         return "\n".join(result)
 
-    def add_module(self, parent, name: str):
+    def add_module(self, parent, name: str, *, src_loc=None, cell_src_loc=None):
```
whitequark, line 214: I guess this works because there's a 1:1 relationship between modules and instantiations, but it's a little weird. (I don't think there is a strong need to change this.)

whitequark (commented):
Looks OK other than inline comments.

amaranth/hdl/_nir.py
wanda-phi: Well we don't really have any other place to put it, so...

whitequark (approved):
LGTM

PR #1151, opened 2024-02-19
lib.enum: honor enum.nonmember.

kivikakk:
Use `_EnumDict._member_names` to determine which members to consider. This way we don't need to redo sunder/dunder checks, and non-members are correctly excluded.

This is a defacto public API, given it remains usable from py3.8 until py3.12 inclusive.  (`_member_names` changes from a list to a keys-only dict for performance reasons in py3.11, but they iterate the same.) In current Python main (i.e. what will most likely be 3.13), a `member_names` property is added which returns those keys.

### DIFF
--- a/amaranth/lib/enum.py
+++ b/amaranth/lib/enum.py
@@ -41,9 +41,9 @@ class EnumMeta(ShapeCastable, py_enum.EnumMeta):
         # Prepare enumeration members for instantiation. This logic is unfortunately very
         # convoluted because it supports two very different code paths that need to share
         # the emitted warnings.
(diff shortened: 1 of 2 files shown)


whitequark (approved):

whitequark:
Could you squash please? Also, let me know if you need a new 0.4.x series release; I've marked this PR as a backport but I only do actual backporting/releases when there is demand.

kivikakk:
Squashed!

And thanks re: the releases, understood! I'm living on `main` so no need from me at this stage. :)

Issue #1100, opened 2024-02-10
Refactor `back.rtlil` to convert Amaranth IR structures to Yosys IR structures


whitequark:
... and then the latter to strings, instead of going to strings directly. The current design was reasonable back when Amaranth was first written, but at this point it's an impediment.

A mostly mechanical refactor of low difficulty.

**Only to be started after new IR lands.**

PR #1152, opened 2024-02-19
back.rtlil: Opportunistically trim zero and sign extension on operands.

### LINKED ISSUE #1148: New IR causes DSP inference issues with Quartus
The following design (within the Details) infers 2 DSP blocks: `Info (21062): Implemented 2 DSP elements`.

<details>

```python
from amaranth import *
from amaranth.lib import wiring
from amaranth.lib.wiring import In, Out

from amaranth_boards.de10_nano import DE10NanoPlatform

class DSPMACBlock(wiring.Component):
    # using signed 18x18->36 mode
    # https://www.intel.com/content/www/us/en/docs/programmable/683375/current/independent-multiplier-mode.html

    mul_a: In(signed(18)) # A input is 18 bits
    mul_b: In(signed(18)) # B input is 18 bits
    result: Out(signed(36)) # result is len(A)+len(B)

        m = Module()

        m.d.sync += self.result.eq(self.mul_a * self.mul_b)

        return m

class Top(Elaboratable):
        m = Module()

        m.submodules.mac = mac = DSPMACBlock()
        m.d.comb += [
            mac.mul_a.eq(Mux(platform.request("switch", 0).i, -1, 0)), # use all bits
            mac.mul_b.eq(Mux(platform.request("switch", 1).i, -1, 0)),
            platform.request("led", 0).o.eq(mac.result[-1])
        ]

        return m

if __name__ == "__main__":
    DE10NanoPlatform().build(Top())
```

</details>

Given that this is a supported DSP mode, I would expect this to infer only 1 block. This worked properly with the old IR. If I make both operands of the `*` signed by modifying line 107 of the generated Verilog (`assign \$1 = ...`) then correctly 1 block is inferred.

Not sure if there's any difference in functionality between the two scenarios.

Versions:
* amaranth 24a392887af19a9d013252759ec209d5a91a378a
* amaranth-boards: https://github.com/amaranth-lang/amaranth-boards/commit/170675812b71ee722bcf8ccdb88409a9ad97ffe2
* Yosys 0.38 (git sha1 543faed9c8c, gcc 13.2.0 -fPIC -Os)
* Quartus x86_64 Linux 20.1.1.720

wanda-phi:
Fixes #1148.

### DIFF
--- a/amaranth/back/rtlil.py
+++ b/amaranth/back/rtlil.py
@@ -744,6 +744,16 @@ class ModuleEmitter:
         emit_assignments(proc, _nir.Net.from_const(1))
         assert pos == len(cell.assignments)
 
+    def shorten_operand(self, value, *, signed):
+        value = list(value)
+        if signed:
+            while len(value) > 1 and value[-1] == value[-2]:
+                value.pop()
+        else:
+            while len(value) > 0 and value[-1] == _nir.Net.from_const(0):
+                value.pop()
+        return _nir.Value(value)
+
     def emit_operator(self, cell_idx, cell):
         UNARY_OPERATORS = {
             "-":    "$neg",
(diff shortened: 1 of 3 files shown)


whitequark (commented):
This is a little cursed and I'm worried about the potential for bugs, but we have to start doing randomized testing either way...

I would be much happier with at least a few tests, I think.

wanda-phi:
> This is a little cursed and I'm worried about the potential for bugs, but we have to start doing randomized testing either way...
> 
> I would be much happier with at least a few tests, I think.

Yeah....

I do want to add some tests for this, but unfortunately this'll have to wait until #1100 lands, we don't really have good infrastructure for that right now.

tpwrules (commented):
Works on my machine™! Thank you.

whitequark:
Since it's addressing a regresstion I'm not going to block this on #1100, @wanda-phi, your call on when to merge this.

whitequark (approved):

whitequark:
Oh, just noticed the branch name :D

whitequark (approved):
Regretfully, LFGTM.

PR #1156, opened 2024-02-27
sim.pysim: Only close VCD/GTKW files if we opened them ourselves.

### LINKED ISSUE #1107: PySim VCDWriter is useless with `io.StringIO`
It unconditionally calls `close` on the VCD and that destroys the `StringIO` object. Questionable design of the latter, perhaps, but we have to live with it.

~Not entirely clear what to do with it. I suggest just special-casing `io.StringIO` for specifically `write_vcd(vcd_file=)`. Though, there could be a similar issue with a `tempfile.TemporaryFile` or `tempfile.SpooledTemporaryFile`, so maybe a `close_vcd_file=True` argument would work better.~

It should only close IO objects that it had opened itself.

wanda-phi:
Fixes #1107.

### DIFF
--- a/amaranth/sim/pysim.py
+++ b/amaranth/sim/pysim.py
@@ -41,10 +41,14 @@ class _VCDWriter:
         # the simulator is still usable if it's not installed for some reason.
         import vcd, vcd.gtkw
 
+        self.close_vcd = False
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Thanks!

PR #1157, opened 2024-02-27
sim.pysim: Only close VCD/GTKW files if we opened them ourselves. [0.4 backport]

wanda-phi:
See #1156, #1107.


whitequark (approved):
Thanks!

PR #1161, opened 2024-02-27
back.rtlil: Remove code allowing internal yosys cells in `Instance`.

wanda-phi:
This was allowed only because Amaranth used it internally. Now that all uses are gone, let's just disallow it entirely.

### DIFF
--- a/amaranth/back/rtlil.py
+++ b/amaranth/back/rtlil.py
@@ -923,12 +923,8 @@ class ModuleEmitter:
             ports[name] = self.instance_wires[cell_idx, name]
         for name, nets in cell.ports_io.items():
             ports[name] = self.sigspec(nets)
-        if cell.type.startswith("$"):
-            type = cell.type
-        else:
-            type = "\\" + cell.type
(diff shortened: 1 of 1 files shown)


amaranth/back/rtlil.py
```diff
@@ -923,12 +923,8 @@ def emit_instance(self, cell_idx, cell):
             ports[name] = self.instance_wires[cell_idx, name]
         for name, nets in cell.ports_io.items():
             ports[name] = self.sigspec(nets)
-        if cell.type.startswith("$"):
-            type = cell.type
-        else:
-            type = "\\" + cell.type
-        self.builder.cell(type, cell.name, ports=ports, params=cell.parameters,
-                         attrs=cell.attributes, src=_src(cell.src_loc))
+        self.builder.cell("\\" + cell.type, cell.name, ports=ports, params=cell.parameters,
```
whitequark, line 926:
```suggestion
        self.builder.cell(f"\\{cell.type}", cell.name, ports=ports, params=cell.parameters,
```

whitequark (approved):
Let's go! :fire:

PR #1162, opened 2024-02-27
back.rtlil: Emit proper source location for port-signals.

wanda-phi:


amaranth/back/rtlil.py
```diff
@@ -372,6 +373,7 @@ def emit_signal_wires(self):
             # the design is flattened) will do that anyway, so it doesn't matter.
             attrs = self.value_attrs.setdefault(value, {})
             attrs.update(signal.attrs)
+            self.value_src_loc[value] = signal.src_loc
```
whitequark, line 376: Should we not do this for other values too? Most of them have a source location (all except `Const`. And I think we should just track them for `Const` too, no reason not to)

wanda-phi: we're talking NIR values here, not AST values; they don't have locations

wanda-phi: more specifically, this change is to make sure we actually emit locations for signal wires, in the case when they also become port wires; the thing we *don't* have currently covered is anonymous ports (since they don't really have 1-1 correspondence to signals); this could also be improved, but that's a more complex issue

whitequark (approved):
I confirm this fills in the missing source locations.

PR #1163, opened 2024-02-27
lib.io, build.res: Make `Pin` and related objects interfaces.

### LINKED ISSUE #1040: Make Pin an interface object
Right now it's a weird hybrid of a Record and an interface object.

wanda-phi:
Fixes #1040.

Fixes two unrelated issues:

- when `Subsignal` was involved, the resulting object had plain `Pin` objects instead of flipped
- `i_clk` has wrong direction

### DIFF
--- a/amaranth/lib/io.py
+++ b/amaranth/lib/io.py
@@ -1,74 +1,18 @@
-import warnings
-
 from .. import *
-with warnings.catch_warnings():
-    warnings.filterwarnings(action="ignore", category=DeprecationWarning)
-    from ..hdl.rec import *
-from ..lib.wiring import In, Out, Signature, flipped, FlippedInterface
+from ..lib import wiring
+from ..lib.wiring import In, Out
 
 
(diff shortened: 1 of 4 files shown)


amaranth/lib/io.py
```diff
@@ -119,23 +62,76 @@ class Pin(Record):
...
         is present.
     """
-    def __init__(self, width, dir, *, xdr=0, name=None, src_loc_at=0):
-        self.width = width
-        self.dir   = dir
-        self.xdr   = xdr
 
-        super().__init__(pin_layout(self.width, self.dir, self.xdr),
-                         name=name, src_loc_at=src_loc_at + 1)
+    class Signature(Signature):
```
whitequark, line 66: Cursed line. This is why I do

```
from amaranth.lib import wiring
from amaranth.lib.wiring import In, Out
```

whitequark (approved):
LGTM, thank you!

PR #1164, opened 2024-02-27
hdl._dsl: Change FSM codegen to avoid mutating AST nodes.

### LINKED ISSUE #1066: Improve code generation strategy for FSMs
Right now, the code in `amaranth.hdl._dsl` for FSMs does something... weird with the state signal. It should not do that. What it should do instead is:
- Make one temporary signal per state whenever `.ongoing()` is used and the FSM is inactive. (This is basically late binding for states.)
- When the FSM definition is complete, create the state signal of the appropriate size and with the right value repr.
- Whenever encountering `m.next`, use an assignment to a value-castable that resolves to the FSM signal. It won't be lowered to a value before the FSM definition is done, so it can be just added as a statement.

Then, we'll be able to make `Signal.width` immutable.

wanda-phi:
Fixes #1066.


amaranth/hdl/_dsl.py
```diff
@@ -146,16 +147,50 @@ def helper(*args, **kwds):
     return decorator
 
 
+class FSMNextStatement(_LateBoundStatement):
+    def __init__(self, ctrl_data, state, src_loc_at=0):
```
whitequark, line 151:
```suggestion
    def __init__(self, ctrl_data, state, *, src_loc_at=0):
```

amaranth/hdl/_dsl.py
```diff
@@ -146,16 +147,50 @@ def helper(*args, **kwds):
...
+                for patterns, stmts in stmt.cases.items()
+            ),
+            src_loc=stmt.src_loc,
+            case_src_locs=stmt.case_src_locs,
+        )
+    elif isinstance(stmt, (Assign, Property)):
+        return stmt
+    else:
+        assert False # :nocov:
+
```
whitequark, line 191:
```suggestion


```

whitequark (commented):
Cursed but ultimately necessary.

whitequark (approved):
LGTM. This is pretty cursed but I see no real choice. Hopefully much of this (`_top_comb_statements` for example) will be replaced when we design and implement scopes.

whitequark:
Thanks!

bl0x:
Just a note:
This change makes it impossible to refer to `FSM.state` inside of the FSM context. It used to be possible before, but now any code doing this throws the error:
`AttributeError: 'FSM' object has no attribute 'state'.`
Referring to `state` is only possible after the FSM context.

whitequark:
Yes, that's expected. It was never particularly well-defined, for example if you used `len(fsm.state)` within the FSM context, or `Signal.like(fsm.state)` your code was silently broken.

bl0x:
Very good that this is fixed now.

PR #1165, opened 2024-02-28
hdl._ast: Make AST nodes immutable.

### LINKED ISSUE #1067: Make AST nodes immutable
Right now it's actually possible to do things like "set `sig.width` on a `Signal`" or "modify `c.parts` on a `Cat`". This has never been explicitly advertised as something that is supported and is arguably a misuse of the API. Amaranth internally treats AST nodes as immutable; all fragment transformers create new ones, and it doesn't mutate them elsewhere. Any mutation would happen only in downstream code (there's one instance deep in the FSM code generation, #1066).

Now that I'm working on reference documentation I'd like to either state that `Value`s are immutable and make them so, or else document them as mutable. This could potentially break some downstream code though, so I'm nominating this for discussion.

If we decide that the nodes should be mutable, then setters must be introduced that typecheck the assignments.

wanda-phi:
Fixes #1067.

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -1462,26 +1462,38 @@ class Const(Value, metaclass=_ConstMeta):
 
     def __init__(self, value, shape=None, *, src_loc_at=0):
         # We deliberately do not call Value.__init__ here.
-        self.value = int(operator.index(value))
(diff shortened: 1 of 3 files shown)


whitequark (approved):
LGTM

PR #1168, opened 2024-02-28
lib.io: Expose `Pin` path and name as attributes.

wanda-phi:


whitequark (approved):
I think the API is questionable but since the entirety of `lib.io` is basically just for compatibility, let's merge this to fix existing breakage.

PR #1104, opened 2024-02-11
ir: kill Fragment.ports

wanda-phi:
Fixes #191.
Fixes #320.
Fixes #398.
Fixes #405.
Fixes #600.

PR #1171, opened 2024-02-28
build.plat, back.rtlil: Fix #1104 fallout.

wanda-phi:


amaranth/build/plat.py
```diff
@@ -307,8 +309,8 @@ def toolchain_prepare(self, fragment, name, *, ports, emit_src=True, **kwargs):
         # and to incorporate the Amaranth version into generated code.
         autogenerated = f"Automatically generated by Amaranth {__version__}. Do not edit."
 
-        rtlil_text, self._name_map = rtlil.convert_fragment(fragment, ports=ports, name=name,
-                                                            emit_src=emit_src)
+        rtlil_text, self._name_map = rtlil.convert_fragment(fragment, name=name,
+                                                            emit_src=emit_src, propagate_domains=False)
```
whitequark, line 313: I think this might be a stale option?

whitequark: Oh, no, it does get passed to `prepare()` down the road doesn't it

whitequark (approved):
LGTM, thank you!

amaranth/build/plat.py
wanda-phi: yup

PR #1173, opened 2024-02-28
hdl._ir: Inline `AssignmentLegalizer` into netlist building.

### LINKED ISSUE #1150: Driver-Driver conflict causes unclear error message [regression from 0.4]
The following code warns that the hierarchy will be flattened:
`DriverConflict: Signal '(sig signal)' is driven from multiple fragments: top, top.test; hierarchy will be flattened`

proceeds with Yosys, yielding a syntax error:
`ERROR: Parser error in line 126: syntax error`

The line in question:
`    connect \D None`

```python
from amaranth import *
from amaranth_boards import versa_ecp5_5g as FPGA
from amaranth.vendor import LatticeECP5Platform
import os
from amaranth.lib.fifo import AsyncFIFOBuffered

class TestElaboratable(Elaboratable):
	def __init__(self):
		self.signal = Signal()

	def elaborate(self, platform) -> Module:
		m = Module()

		m.d.comb += self.signal.eq(0)
		# Uncommenting this causes driver-driver conflict error
		#m.d.sync += self.signal.eq(0)

		return m

class MCVE(Elaboratable):
		m = Module()

		m.submodules.test = test = TestElaboratable()
		m.d.sync += test.signal.eq(0)

		return m


os.environ["AMARANTH_verbose"] = "Yes"
FPGA.VersaECP55GPlatform().build(MCVE())
```

Presumably this is supposed to cause a driver-driver conflict error (as it does when I tried to further simplify it)

wanda-phi:
Fixes #1150.


whitequark (approved):
LGTM, thank you!

PR #1180, opened 2024-03-03
test_hdl_ir: Add a bunch of tests for NIR emitter.

wanda-phi:


whitequark (approved):
I liked the original title!

PR #1196, opened 2024-03-11
Some f-strings in utils.py are broken (missing the letter "f")

mcclure:
Without this patch, `exact_log2(3)` will print `"{n} is not a power of 2"` instead of `"3 is not a power of 2"` as intended

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -16,7 +16,7 @@ class Log2TestCase(unittest.TestCase):
         self.assertEqual(ceil_log2(9), 4)
         with self.assertRaises(TypeError):
             ceil_log2(1.5)
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, r"^-1 is negative$"):
             ceil_log2(-1)
 
     def test_exact_log2(self):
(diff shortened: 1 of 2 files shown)


whitequark (approved):

whitequark:
Actually, care to add a test?

whitequark:
Also, your commit message should be formatted something like

    utils: F-strings are missing the letter "f"

mcclure:
Did a forced push with a corrected summary plus a test.

I tested with `pdm install && pdm run test` and although I am almost totally certain the test is correct and passes, I'm not sure I ran the tests right because I got 10 failures in other areas. I think this test is right though because when I changed "not" to "nfot" in the test string the failure count went up to 11 :P

wanda-phi:
> Did a forced push with a corrected summary plus a test.
> 
> I tested with `pdm install && pdm run test` and although I am almost totally certain the test is correct and passes, I'm not sure I ran the tests right because I got 10 failures in other areas. I think this test is right though because when I changed "not" to "nfot" in the test string the failure count went up to 11 :P

I don't actually see any test?

tests/test_utils.py
```diff
@@ -25,7 +25,7 @@ def test_exact_log2(self):
         self.assertEqual(exact_log2(4), 2)
         self.assertEqual(exact_log2(8), 3)
         for val in [-1, 0, 3, 5, 6, 7, 9]:
-            with self.assertRaises(ValueError):
+            with self.assertRaisesRegex(ValueError, (f"^{val} is not a power of 2$")):
```
whitequark, line 28: One for `ceil_log2` too please.

whitequark (changes requested):

mcclure:
Apologies, should have got that the first time.

whitequark (approved):

PR #1198, opened 2024-03-13
lib.memory: Allow setting `Memory.init`.

wanda-phi:
The `init` property is already mutable, so this adds no actual new functionality, just convenience.


whitequark (approved):
Thanks!

PR #1202, opened 2024-03-14
Implement RFC 51: Add `ShapeCastable.from_bits` and `amaranth.lib.data.Const`.

wanda-phi:
TODO:

- [x] tests
- [x] documentation
- [ ] ~~add custom `Const` subclasses for `Struct`/`Union`?~~ deferred

### DIFF
--- a/amaranth/lib/data.py
+++ b/amaranth/lib/data.py
@@ -2,16 +2,17 @@ from abc import ABCMeta, abstractmethod
 from enum import Enum
 from collections.abc import Mapping, Sequence
 import warnings
+import operator
 
 from amaranth._utils import final
 from amaranth.hdl import *
 from amaranth.hdl._repr import *
-from amaranth.hdl._ast import ShapeCastable, ValueCastable
+from amaranth import hdl
 
 
 __all__ = [
     "Field", "Layout", "StructLayout", "UnionLayout", "ArrayLayout", "FlexibleLayout",
-    "View", "Struct", "Union",
+    "View", "Const", "Struct", "Union",
 ]
 
 
@@ -26,7 +27,7 @@ class Field:
 
     Attributes
     ----------
-    shape : :ref:`shape-like <lang-shapelike>`
+    shape : :class:`.ShapeLike`
         Shape of the field. When initialized or assigned, the object is stored as-is.
     offset : :class:`int`, >=0
         Index of the least significant bit of the field.
@@ -55,14 +56,14 @@ class Field:
     def width(self):
         """Width of the field.
 
-        This property should be used over ``self.shape.width`` because ``self.shape`` can be
+        This property should be used over :py:`self.shape.width` because :py:`self.shape` can be
         an arbitrary :ref:`shape-like <lang-shapelike>` object, which may not have
-        a ``width`` property.
+        a :py:`width` property.
 
         Returns
         -------
         :class:`int`
-            ``Shape.cast(self.shape).width``
+            :py:`Shape.cast(self.shape).width`
         """
         return Shape.cast(self.shape).width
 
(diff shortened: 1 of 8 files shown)


amaranth/lib/data.py
```diff
@@ -590,7 +594,7 @@ class View(ValueCastable):
     an Amaranth value instead of a constant integer. The returned element is chosen dynamically
     in that case.
 
-    A view can only be compared for equality with another view of the same layout,
+    A view can only be compared for equality with another view or const of the same layout,
```
whitequark, line 597:
```suggestion
    A view can only be compared for equality with another view or constant of the same layout,
```

amaranth/lib/data.py
```diff
@@ -789,6 +799,221 @@ def __repr__(self):
...
+    """A constant value viewed through the lens of a layout.
+
+    The :class:`Const` class is a constant analog of :class:`View`. It implements largely
+    the same operations, except that it deals with compile-time constants instead of arbitrary
+    :class:`Value`.
+
+    Creating a const
+    ################
+
+    A constant can be created from a dict or list of field values using :meth:`Layout.const`,
```
whitequark, line 812: I add links to the Python stdlib just to be completely unambiguous on what we mean at all times.

```suggestion
    A constant can be created from a :class:`dict` or :class:`list` of field values using :meth:`Layout.const`,
```

amaranth/lib/data.py
```diff
@@ -789,6 +799,221 @@ def __repr__(self):
...
+    or from a bit-pattern using :meth:`Layout.from_bits`.
+
+    Accessing a const
+    #################
+
+    Slicing a const or accessing its attributes returns a part of the underlying value
+    corresponding
+    to the field with that index or name. If the shape of the field is a :class:`Layout`,
+    the accessed value will be another :class:`Const`. If it is another :ref:`shape-like <lang-shapelike>`
+    object implementing ``from_bits``, it will be the result of calling that method. Otherwise,
```
whitequark, line 822: Should be a link to ShapeCastable.from_bits

amaranth/lib/data.py
```diff
@@ -789,6 +799,221 @@ def __repr__(self):
...
+    it will be a plain :class:`int`.
+
+    Slicing a view whose layout is an :class:`ArrayLayout` can be done with an index that is
+    an Amaranth value instead of a constant integer. The returned element is chosen dynamically
+    in that case, and the resulting value will be a :class:`View` instead of :class:`Const`.
```
whitequark, line 827: Oh, that's... a really fascinating behavior I haven't thought about at all. No issue with it, just that it's kind of unusual.

amaranth/lib/data.py
```diff
@@ -789,6 +799,221 @@ def __repr__(self):
...
+
+    A const can be compared for equality with another const or view of the same layout.
+    When compared with another const, the result will be a Python :class:`bool`.
+    When compared with a view, the result will be a single-bit value. No other operators
```
whitequark, line 831: Here the disambiguation is actually important!

```suggestion
    When compared with a view, the result will be a single-bit :class:`Value`. No other operators
```

amaranth/lib/data.py
```diff
@@ -789,6 +799,221 @@ def __repr__(self):
...
+        """Convert to Amaranth value.
+
+        Returns
+        -------
+        :class:`amaranth.hdl.Const`
+            A constant Amaranth value with the underlying bit-pattern.
+        """
+        return hdl.Const(self.__target, self.__layout.size)
+
+    def as_bits(self):
```
whitequark, line 876: Maybe add a warning to `View` that using a field called `as_bits` will break `Const`?

wanda-phi: It ended up a little awkward (I think the documentation will read better when we have custom `Const` subclass support?)

whitequark (approved):
LGTM

PR #1203, opened 2024-03-15
Implement RFC 53: Low-level I/O primitives.

wanda-phi:
TODO:

- [x] tests

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -17,7 +17,7 @@ from .._unused import *
 
 __all__ = [
     "Shape", "signed", "unsigned", "ShapeCastable", "ShapeLike",
-    "Value", "Const", "C", "AnyConst", "AnySeq", "Operator", "Mux", "Part", "Slice", "Cat",
+    "Value", "Const", "C", "AnyConst", "AnySeq", "Operator", "Mux", "Part", "Slice", "Cat", "Concat",
     "Array", "ArrayProxy",
     "Signal", "ClockSignal", "ResetSignal",
     "ValueCastable", "ValueLike",
@@ -25,6 +25,7 @@ __all__ = [
     "Format",
     "Statement", "Switch",
     "Property", "Assign", "Print", "Assert", "Assume", "Cover",
+    "IOValue", "IOPort", "IOConcat", "IOSlice",
     "SignalKey", "SignalDict", "SignalSet",
 ]
 
@@ -1480,7 +1481,7 @@ class Const(Value, metaclass=_ConstMeta):
         obj = Value.cast(obj)
         if type(obj) is Const:
             return obj
-        elif type(obj) is Cat:
+        elif type(obj) is Concat:
             value = 0
             width = 0
             for part in obj.parts:
@@ -1636,9 +1637,13 @@ def Mux(sel, val1, val0):
 @final
 class Slice(Value):
     def __init__(self, value, start, stop, *, src_loc_at=0):
-        if not isinstance(start, int):
+        try:
+            start = int(operator.index(start))
+        except TypeError:
             raise TypeError(f"Slice start must be an integer, not {start!r}")
-        if not isinstance(stop, int):
+        try:
+            stop = int(operator.index(stop))
+        except TypeError:
             raise TypeError(f"Slice stop must be an integer, not {stop!r}")
 
         value = Value.cast(value)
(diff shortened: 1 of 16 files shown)


whitequark:
I think we need also guide-level documentation for instances and IO values,

amaranth/hdl/_ast.py
```diff
@@ -1733,8 +1734,17 @@ def __repr__(self):
                                            self.width, self.stride)
 
 
+class _CatMeta(ABCMeta):
```
whitequark, line 1737: I'd rather make `Cat` a function, renaming the concatenation class to `Concat` and the IO version to `IOConcat` for disambiguation.

whitequark: Actually, do you even need a metaclass here? Wouldn't `__new__` suffice?

amaranth/hdl/_ast.py
```diff
@@ -2732,6 +2742,160 @@ def case_repr(keys, stmts):
...
+    def cast(obj):
+        if isinstance(obj, IOValue):
+            return obj
+        elif isinstance(obj, Value) and len(obj) == 0:
+            return IOCat()
+        else:
+            raise TypeError(f"Object {obj!r} cannot be converted to an IO value")
+
+    def __init__(self, *, src_loc_at=0):
+        super().__init__()
```
whitequark, line 2756: There's no superclass, right?

amaranth/hdl/_ast.py
```diff
@@ -2732,6 +2742,160 @@ def case_repr(keys, stmts):
...
+
+    @property
+    def metadata(self):
+        return self._metadata
+
+    def _ioports(self):
+        return {self}
+
+    def __repr__(self):
+        return f"(ioport {self.name})"
```
whitequark, line 2817: Maybe with a dash, to visually group the `io-` variants?

```suggestion
        return f"(io-port {self.name})"
```

amaranth/hdl/_ast.py
```diff
@@ -2732,6 +2742,160 @@ def case_repr(keys, stmts):
...
+    def __init__(self, width, *, name=None, attrs=None, metadata=None, src_loc_at=0):
+        super().__init__(src_loc_at=src_loc_at)
+
+        if name is not None and not isinstance(name, str):
+            raise TypeError(f"Name must be a string, not {name!r}")
+        self.name = name or tracer.get_var_name(depth=2 + src_loc_at)
+
+        self._width = operator.index(width)
+        self._attrs = dict(() if attrs is None else attrs)
+        self._metadata = (None,) * self._width if metadata is None else tuple(metadata)
```
whitequark, line 2796: I think this needs a check for the size and type of the provided `metadata`.

amaranth/hdl/_ast.py
```diff
@@ -2732,6 +2742,160 @@ def case_repr(keys, stmts):
...
+    def _ioports(self):
+        return {self}
+
+    def __repr__(self):
+        return f"(ioport {self.name})"
+
+
+@final
+class IOCat(IOValue):
+    def __init__(self, *args, src_loc_at=0):
```
whitequark, line 2822: This class isn't directly callable, right? I'd rather accept `parts` in `__init__` and ditch `flatten`. (The same for the `class Cat:` so they match).

I don't love `flatten` due to its weird semantics and I somewhat prefer not adding more calls to it.

amaranth/hdl/_ast.py
```diff
@@ -2732,6 +2742,160 @@ def case_repr(keys, stmts):
...
+        if stop not in range(-n, n+1):
+            raise IndexError(f"Cannot stop slice {stop} bits into {n}-bit value")
+        if stop < 0:
+            stop += n
+        if start > stop:
+            raise IndexError(f"Slice start {start} must be less than slice stop {stop}")
+
+        super().__init__(src_loc_at=src_loc_at)
+        self._value = value
+        self._start = int(operator.index(start))
```
whitequark, line 2870: This seems completely redundant with `isinstance(start, int)` above.

amaranth/hdl/_ast.py
wanda-phi: Ok, going with `__new__`

whitequark (commented):
I feel that the algorithms in `Design` would be clearer if it was done in one (or two, second one computing LCA) DFS pass. However, this doesn't need to happen now.

LGTM, the only thing left is docs which I will take care of imminently.

whitequark (approved):
LGTM, great work!

PR #1214, opened 2024-03-18
lib.io: Implement `*Port` from RFC 55.

wanda-phi:


amaranth/build/res.py
```diff
@@ -18,12 +18,18 @@ class SingleEndedPort:
...
+        return len(self.io)
+
 
 class DifferentialPort:
     def __init__(self, p, n):
         self.p = p
         self.n = n
 
+    def __len__(self):
+        return len(self.p)
```
whitequark, line 31: `assert` here?

whitequark (approved):

PR #1215, opened 2024-03-19
lib.io: Implement `*Buffer` from RFC 55.

wanda-phi:


amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+    def port(self):
+        return self._port
+
+    @property
+    def direction(self):
+        return self.signature.direction
+
+    def elaborate(self, platform):
+        if hasattr(platform, "get_io_buffer"):
+            res = platform.get_io_buffer(self)
```
whitequark, line 334: Hm, we don't use this pattern elsewhere--we dispatch unconditionally based on `hasattr`.

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+            if res is not None:
+                return res
+
+        m = Module()
+
+        invert = sum(bit << idx for idx, bit in enumerate(self._port.invert))
+        if self.direction is not Direction.Input:
+            if invert:
```
whitequark, line 342: This is correct but pretty confusing--I think `if invert != 0` is clearer, just to narrow the state space.

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+        if self.direction is not Direction.Output:
+            if invert:
+                i_inv = Signal.like(self.i)
+                m.d.comb += self.i.eq(i_inv ^ invert)
+            else:
+                i_inv = self.i
+
+        if isinstance(self._port, SingleEndedPort):
+            if self.direction is Direction.Input:
+                m.submodules.iob = IOBufferInstance(self._port.io, i=i_inv)
```
whitequark, line 356: `io_buffer` or `buffer` or `inner` or something maybe? "iob" is confusing to newcomers, and just about everyone will see this in the hierarchy.

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
         return f"DifferentialPort({self._p!r}, {self._n!r}, invert={invert!r}, direction={self._direction})"
 
 
+class Buffer(wiring.Component):
+    """A combinatorial I/O buffer.
+
+    Parameters
+    ----------
+    direction : Direction
```
whitequark, line 266: These should be using `:class:`, etc. (Everywhere else in the docstrings too.)

amaranth/lib/io.py
wanda-phi: this doesn't actually show up in the hierarchy (it's lowered to an anonymous `$tribuf`)

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+        else:
+            raise TypeError("Cannot elaborate generic 'Buffer' with port {self._port!r}")
+
+        return m
+
+
+class FFBuffer(wiring.Component):
+    """A registered I/O buffer.
+
+    Equivalent to a plain :class:`Buffer` combined with reset-less registers on ``i``, ``o``, ``oe``.
```
whitequark, line 379: These should be using `:py:`. (Anything making reference to Python tokens should be using that.)

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
         return f"DifferentialPort({self._p!r}, {self._n!r}, invert={invert!r}, direction={self._direction})"
 
 
+class Buffer(wiring.Component):
+    """A combinatorial I/O buffer.
+
+    Parameters
+    ----------
+    direction : Direction
+    port : SingleEndedPort, DifferentialPort, or a custom platform-specific type
```
whitequark, line 267: Needs a comment explaining when it can be a custom platform-specific type. (I think you can specify a parameter twice and create a conditional this way?)

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+        @property
+        def direction(self):
+            return self._direction
+
+        @property
+        def width(self):
+            return self._width
+
+        def __eq__(self, other):
+            return type(self) == type(other) and self.direction == other.direction and self.width == other.width
```
whitequark, line 308: I usually use `type(self) is type(other)` in case someone is doing weird shit with metaclasses.

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+            res = platform.get_io_buffer(self)
+            if res is not None:
+                return res
+
+        m = Module()
+
+        m.submodules.iob = iob = Buffer(self.direction, self.port)
+
+        if self.direction is not Direction.Output:
+            i_tmp = Signal.like(self.i, reset_less=True)
```
whitequark, line 486: `i_reg`

amaranth/lib/io.py
```diff
@@ -253,6 +258,362 @@ def __repr__(self):
...
+    In the input direction, the port is sampled at both edges of the input clock domain.
+    The data sampled on the active clock edge of the domain appears on ``i[0]`` with a delay
+    of 1 clock cycle. The data sampled on the opposite clock edge appears on ``i[1]`` with a delay
+    of 0.5 clock cycle. Both ``i[0]`` and ``i[1]`` thus change on the active clock edge of the domain.
+
+    In the output direction, both ``o[0]`` and ``o[1]`` are sampled on the active clock edge
+    of the domain.  The value of ``o[0]`` immediately appears on the output port.  The value
+    of ``o[1]`` then appears on the output port on the opposite edge, with a delay of 0.5 clock cycle.
+
+    The degree of support for this component varies by platform.
```
whitequark, line 513: The *degree*? Isn't it either supported or not?

amaranth/lib/io.py
whitequark: Oh. Then I guess it can be just `m.submodules +=`?

amaranth/lib/io.py
```diff
@@ -253,6 +295,354 @@ def __repr__(self):
...
+        Parameters
+        ----------
+        direction : :class:`Direction`
+        width : :class:`int`
+
+        Attributes
+        ----------
+        i: :py:`unsigned(width)` (if :py:`direction in (Direction.Input, Direction.Bidir)`)
+        o: :py:`unsigned(width)` (if :py:`direction in (Direction.Output, Direction.Bidir)`)
+        oe: :py:`unsigned(1)` (if :py:`direction in (Direction.Output, Direction.Bidir)`)
```
whitequark, line 444: This should probably mention that the `init` value depends on direction too.

whitequark (approved):
LGTM, thanks!

The docstrings are intentionally not linked from `docs/stdlib/io.rst` for the time being; I will take care of that later.

PR #1218, opened 2024-03-19
hdl._ir: Remove support for non-`Elaboratable` elaboratables.

### LINKED ISSUE #1216: Make the use of `Elaboratable` for elaboratable objects required
It's always required a warning. Technically it's not a deprecation warning, but it might as well be; none of our other protocol classes (`ValueCastable` for example) are optional, `Elaboratable` shouldn't be either.

wanda-phi:
Fixes #1216.


whitequark (approved):
Thanks!

PR #1230, opened 2024-03-22
sim.core: correct deprecation warning

polymerizedsage:
Deprecation warning incorrectly referred to RFC 47 rather than RFC 27.


whitequark (approved):
Thanks!

PR #1231, opened 2024-03-23
Only preempt simulator testbenches on explicit wait points

whitequark:
Before this commit, testbenches (generators added with `add_testbench`) were not only preemptible after any `yield`, but were *guaranteed* to be preempted by another testbench after *every* yield. This is evil: if you have any race condition between testbenches, which is common, this scheduling strategy will maximize the resulting nondeterminism by interleaving your testbench with every other one as much as possible. This behavior is an outcome of the way `add_testbench` is implemented, which is by yielding `Settle()` after every command.

One can observe that:
- `yield value_like` should never preempt;
- `yield assignable.eq()` in `add_process()` should not preempt, since it only sets a `next` signal state, or appends to `write_queue` of a memory state, and never wakes up processes;
- `yield assignable.eq()` in `add_testbench()` should only preempt if changing `assignable` wakes up an RTL process. (It could potentially also preempt if that wakes up another testbench, but this has no benefit and requires `sim.set()` from RFC 36 to be awaitable, which is not desirable.)

After this commit, `PySimEngine._step()` is implemented with two nested loops instead of one. The outer loop iterates through every testbench and runs it until an explicit wait point (`Settle()`, `Delay()`, or `Tick()`), terminating when no testbenches are runnable. The inner loop is the usual eval/commit loop, running whenever a testbench changes design state.

`PySimEngine._processes` is a `set`, which doesn't have a deterministic iteration order. This does not matter for processes, where determinism is guaranteed by the eval/commit loop, but causes racy testbenches to pass or fail nondeterministically (in practice depending on the memory layout of the Python process). While it is best to not have races in the testbenches, this commit makes `PySimEngine._testbenches` a `list`, making the outcome of a race deterministic, and enabling a hacky work- around to make them work: reordering calls to `add_testbench()`.

A potential future improvement is a simulation mode that, instead, randomizes the scheduling of testbenches, exposing race conditions early.

---

Needs:
- [x] A test

### DIFF
--- a/amaranth/sim/_pycoro.py
+++ b/amaranth/sim/_pycoro.py
@@ -1,7 +1,7 @@
 import inspect
 
 from ..hdl import *
-from ..hdl._ast import Statement, SignalSet, ValueCastable
+from ..hdl._ast import Statement, Assign, SignalSet, ValueCastable
 from ..hdl._mem import MemorySimRead, MemorySimWrite
 from .core import Tick, Settle, Delay, Passive, Active
 from ._base import BaseProcess, BaseMemoryState
@@ -12,11 +12,12 @@ __all__ = ["PyCoroProcess"]
 
 
 class PyCoroProcess(BaseProcess):
-    def __init__(self, state, domains, constructor, *, default_cmd=None):
+    def __init__(self, state, domains, constructor, *, default_cmd=None, testbench=False):
         self.state = state
         self.domains = domains
         self.constructor = constructor
         self.default_cmd = default_cmd
+        self.testbench = testbench
 
         self.reset()
 
@@ -70,7 +71,7 @@ class PyCoroProcess(BaseProcess):
             except StopIteration:
                 self.passive = True
                 self.coroutine = None
-                return
+                return False # no assignment
 
             try:
                 if command is None:
@@ -88,6 +89,8 @@ class PyCoroProcess(BaseProcess):
                 elif isinstance(command, Statement):
                     exec(_StatementCompiler.compile(self.state, command),
                         self.exec_locals)
+                    if isinstance(command, Assign) and self.testbench:
+                        return True # assignment; run a delta cycle
 
                 elif type(command) is Tick:
                     domain = command.domain
@@ -102,17 +105,20 @@ class PyCoroProcess(BaseProcess):
                     self.add_trigger(domain.clk, trigger=1 if domain.clk_edge == "pos" else 0)
                     if domain.rst is not None and domain.async_reset:
                         self.add_trigger(domain.rst, trigger=1)
-                    return
+                    return False # no assignments
+
+                elif self.testbench and (command is None or isinstance(command, Settle)):
+                    raise TypeError(f"Command {command!r} is not allowed in testbenches")
 
                 elif type(command) is Settle:
                     self.state.wait_interval(self, None)
-                    return
+                    return False # no assignments
 
                 elif type(command) is Delay:
                     # Internal timeline is in 1ps integeral units, intervals are public API and in floating point
                     interval = int(command.interval * 1e12) if command.interval is not None else None
                     self.state.wait_interval(self, interval)
-                    return
+                    return False # no assignments
 
                 elif type(command) is Passive:
                     self.passive = True
@@ -140,6 +146,8 @@ class PyCoroProcess(BaseProcess):
                     state = self.state.slots[index]
                     assert isinstance(state, BaseMemoryState)
                     state.write(addr, data)
+                    if self.testbench:
+                        return True # assignment; run a delta cycle
 
                 elif command is None: # only possible if self.default_cmd is None
                     raise TypeError("Received default command from process {!r} that was added "
--- a/amaranth/sim/core.py
+++ b/amaranth/sim/core.py
@@ -115,33 +115,7 @@ class Simulator:
         self._engine.add_coroutine_process(wrapper, default_cmd=Tick(domain))
 
     def add_testbench(self, process):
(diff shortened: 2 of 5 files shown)


amaranth/sim/core.py
```diff
@@ -137,11 +134,10 @@ def wrapper():
                     try:
                         result = yield command
                         exception = None
-                        yield object.__new__(Settle)
                     except Exception as e:
                         result = None
                         exception = e
-        self._engine.add_coroutine_process(wrapper, default_cmd=None)
+        self._engine.add_testbench_process(wrapper)
```
wanda-phi, line 140: I believe this wrapper is now entirely redundant; all it does is filtering `Settle`, which can be done in the sim loop instead (which is now separate for testbenches)

amaranth/sim/pysim.py
```diff
@@ -439,6 +442,25 @@ def _step(self):
...
+
+        # Run testbenches waiting for an interval to expire, or for a signal to change state
+        converged = False
+        while not converged:
+            converged = True
+            # Schedule testbenches in a deterministic, predictable order by iterating a list
+            for testbench in self._testbenches:
+                if testbench.runnable:
+                    testbench.runnable = False
+                    while testbench.run():
```
wanda-phi, line 459: I believe `process.run()` above should also be wrapped in `while process.run(): pass` so that it doesn't terminate early

amaranth/sim/_pycoro.py
```diff
@@ -58,29 +59,38 @@ def run(self):
...
                 if command is None:
                     command = self.default_cmd
+                if isinstance(command, ValueCastable):
+                    command = Value.cast(command)
                 response = None
                 exception = None
 
-                if isinstance(command, ValueCastable):
-                    command = Value.cast(command)
+                if assigned and isinstance(command, (Value, Tick, Settle, Delay, MemorySimRead)):
```
wanda-phi, line 90: cursed: this also needs `Statement`, unless you can prove its `_rhs_signals` is empty — after all, assignments can also *read* signals

wanda-phi: same for `_MemorySimWrite` actually, its `_addr` and `_data` are currently allowed to be arbitrary expressions

amaranth/sim/pysim.py
whitequark: Early termination only applies to testbenches (it's gated on `if testbench:` in the coroutine process implementation).

amaranth/sim/_pycoro.py
whitequark: This... makes the second commit completely useless, so either a check on `_rhs_signals()` must be added, or the second commit should be rolled back.

amaranth/sim/core.py
whitequark: Huh? The main sim loop doesn't process commands (or is even aware of their presence).

amaranth/sim/_pycoro.py
whitequark: I took the 2nd commit out of the PR.

amaranth/sim/core.py
wanda-phi: er, I meant the loop in `PyCoroProcess.run`

amaranth/sim/pysim.py
wanda-phi: oops, my bad

amaranth/sim/core.py
whitequark: Oh yeah let me fix that

wanda-phi (approved):

stafverhaegen-chipflow:
> A potential future improvement is a simulation mode that, instead, randomizes the scheduling of testbenches, exposing race conditions early.

To me this is a requirement to do in the future and not optional. One of the problems with classic RTL simulation was that by trying to put determinism into situations where there is a race in the design that the problem was not seen during simulation but that the race actually shows up on a real chip.
Therefor in classical flow a post-layout verification step is normally included where simulations are done with back-annotated timing data to catch such situations before actually going to tape-out. I think we have to avoid that for a Amaranth based ASIC flow.

whitequark:
> To me this is a requirement to do in the future and not optional. One of the problems with classic RTL simulation was that by trying to put determinism into situations where there is a race in the design that the problem was not seen during simulation but that the race actually shows up on a real chip.
> Therefor in classical flow a post-layout verification step is normally included where simulations are done with back-annotated timing data to catch such situations before actually going to tape-out. I think we have to avoid that for a Amaranth based ASIC flow.

Are we talking about the same thing? The testbenches are very explicitly *not* a part of the design; they are a part of the test suite. There is nothing to back-annotate because testbenches are behavioral and can only be used to provide stimulus and simulate I/O.

stafverhaegen-chipflow:
What I had in the back of my mind is the use of behavioral models for RTL blocks or full-custom blocks for speed reasons.

whitequark:
> What I had in the back of my mind is the use of behavioral models for RTL blocks or full-custom blocks for speed reasons.

These will not be using `add_testbench` but `add_process` and `add_process` cannot race.

stafverhaegen-chipflow:
Is `await sim.tick().sample(...)` allowed in `add_testbench` behavioral models ?
One may need non-propagated signals then also in these processes.

whitequark:
> Is `await sim.tick().sample(...)` allowed in `add_testbench` behavioral models ?

Yes, `await sim.tick()` and everything you hang off it works exactly the same in `add_process` and `add_testbench`.

stafverhaegen-chipflow:
Then I agree `add_testbench` is OK for these behavioral models.

PR #1234, opened 2024-03-23
Allow Signals to be privately named using `name=""`

### LINKED ISSUE #1223: Signals with private names
In some cases it is desirable to create a signal with a name that doesn't show up in the waveform viewer, such as when a signal is created to hold the results of intermediate computations, or to change shape, and it would just add noise. I propose adding a notion of a "private signal name".

I suggest changing naming rules so that `Signal(name="")` creates a signal with a private name. Such a signal is translated to a private wire in RTLIL (with a `$\d+` name), doesn't get displayed in the waveform viewer, and so on.

---

The FSM ongoing signals (not the state signal) should be changed to have private names in the 0.5 milestone.

tpwrules:
Closes #1223

Tested a couple toy designs. The inclusion of FSM names being private should be enough to prevent regressions.

This does remove a bit of diagnostic power from the FSM tests.

### DIFF
--- a/amaranth/hdl/_ir.py
+++ b/amaranth/hdl/_ir.py
@@ -552,6 +552,8 @@ class Design:
         assigned_names = {name for name, conn, dir in self.ports if name is not None}
         for name, conn, dir in self.ports:
             if name is None:
+                if conn.name == "": # Nothing to name this port!
+                    raise TypeError("Signals with private names cannot be used in unnamed top-level ports")
                 name = _add_name(assigned_names, conn.name)
                 assigned_names.add(name)
             new_ports.append((name, conn, dir))
@@ -590,7 +592,7 @@ class Design:
                     frag_info.io_port_names[conn] = name
 
         for signal in frag_info.used_signals:
-            if signal not in frag_info.signal_names:
+            if signal not in frag_info.signal_names and signal.name != "": # Private name shouldn't be added.
                 frag_info.signal_names[signal] = _add_name(frag_info.assigned_names, signal.name)
         for port in frag_info.used_io_ports:
             if port not in frag_info.io_port_names:
--- a/amaranth/sim/core.py
+++ b/amaranth/sim/core.py
@@ -4,6 +4,7 @@ import warnings
 from .._utils import deprecated
 from ..hdl._cd import *
 from ..hdl._ir import *
+from ..hdl._ast import Value, ValueLike
 from ._base import BaseEngine
 
 
@@ -237,6 +238,16 @@ class Simulator:
                 if hasattr(file, "close"):
                     file.close()
             raise ValueError("Cannot start writing waveforms after advancing simulation time")
+
+        for trace in traces:
+            if isinstance(trace, ValueLike):
+                trace_cast = Value.cast(trace)
+                for trace_signal in trace_cast._rhs_signals():
+                    if trace_signal.name == "":
+                        if trace_signal is trace:
+                            raise TypeError("Cannot trace signal with private name")
+                        else:
+                            raise TypeError(f"Cannot trace signal with private name (within {trace!r})")
 
         return self._engine.write_vcd(vcd_file=vcd_file, gtkw_file=gtkw_file,
                                       traces=traces, fs_per_delta=fs_per_delta)
--- a/tests/test_hdl_dsl.py
+++ b/tests/test_hdl_dsl.py
@@ -603,8 +603,8 @@ class DSLTestCase(FHDLTestCase):
                 )
                 (case 1 )
             )
-            (eq (sig fsm_ongoing_FIRST) (== (sig fsm_state) (const 1'd0)))
-            (eq (sig fsm_ongoing_SECOND) (== (sig fsm_state) (const 1'd1)))
+            (eq (sig) (== (sig fsm_state) (const 1'd0)))
+            (eq (sig) (== (sig fsm_state) (const 1'd1)))
         )
         """)
         self.assertRepr(frag.statements["sync"], """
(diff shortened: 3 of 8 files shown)


amaranth/hdl/_ir.py
```diff
@@ -551,7 +551,7 @@ def _assign_port_names(self):
         new_ports = []
         assigned_names = {name for name, conn, dir in self.ports if name is not None}
         for name, conn, dir in self.ports:
-            if name is None:
+            if name is None and conn.name != "": # Private signals have no name.
```
wanda-phi, line 554: if `conn.name == ""` we need to fail here entirely — we cannot have a top-level port with no name

tests/test_hdl_dsl.py
```diff
@@ -627,8 +627,8 @@ def test_FSM_basic(self):
             "(sig a)": "comb",
             "(sig fsm_state)": "sync",
             "(sig b)": "sync",
-            "(sig fsm_ongoing_FIRST)": "comb",
-            "(sig fsm_ongoing_SECOND)": "comb",
+            "(sig)": "comb",
+            "(sig)": "comb",
```
wanda-phi, line 631: this ... does not really work this way (it's a dict, so the entries are deduplicated)

I'm not sure how to deal with this properly; perhaps this should just be converted to a NIR-based test or something...

amaranth/hdl/_ir.py
tpwrules: The port’s name here is None, we are stopping it from getting the non-name of the signal. If None it should be given a name later.

tests/test_hdl_dsl.py
tpwrules: Overlooked in a find and replace. I was more curious about the FSM tests here.

amaranth/hdl/_ir.py
wanda-phi: but there is no "later" here — when this function runs, the top-level ports are finalised, and they must have a name

this is different from non-top-level ports, where we have auto-naming. top-level ports are the *interface* of the design, and we cannot allow unnamed stuff there

tests/test_hdl_dsl.py
tpwrules: I fixed the duplicate entry.

amaranth/sim/core.py
```diff
@@ -264,4 +265,8 @@ def write_vcd(self, vcd_file, gtkw_file=None, *, traces=()):
                     file.close()
             raise ValueError("Cannot start writing waveforms after advancing simulation time")
 
+        for trace in traces:
+            if isinstance(trace, Signal) and trace.name == "":
+                raise TypeError("Cannot trace private Signal")
```
whitequark, line 270: I think this isn't sufficient due to how pysim VCD writer accepts a bunch of other stuff too, like value-castables, in `traces`.

amaranth/hdl/_ir.py
```diff
@@ -552,6 +552,8 @@ def _assign_port_names(self):
         assigned_names = {name for name, conn, dir in self.ports if name is not None}
         for name, conn, dir in self.ports:
             if name is None:
+                if conn.name == "": # Private Signal has no name so there's nothing to name this port.
+                    raise TypeError("Top-level ports with private Signals must be named")
```
whitequark, line 556:
```suggestion
                    raise TypeError("Signals used as top-level ports must have a non-empty name")
```

whitequark (changes requested):
See inline comments.

amaranth/hdl/_ir.py
tpwrules: This message is a stronger requirement than the code actually enforces, but probably easier for users to understand. Should the code be changed too?

amaranth/sim/core.py
tpwrules: I copied the logic from pysim, it seems more reasonable and I was able to test it.

amaranth/hdl/_ir.py
tpwrules: I changed the message but not the code.

whitequark: Oh, I actually misread the code.

amaranth/hdl/_ir.py
```diff
@@ -552,6 +552,8 @@ def _assign_port_names(self):
         assigned_names = {name for name, conn, dir in self.ports if name is not None}
         for name, conn, dir in self.ports:
             if name is None:
+                if conn.name == "": # Private Signal has no name so there's nothing to name this port.
+                    raise TypeError("Signals used as top-level ports must have a non-empty name")
```
whitequark, line 556: @wanda-phi Actually, do we even need this restriction?

tpwrules: It doesn't break things per se, but in my experiments you do end up with a top level name which is assigned arbitrarily and so liable to change if the design changes. This seems bad for reproducibility and interoperability.

whitequark: Oh! I think that's not entirely motivating the prohibition, since it's quite obvious on examination of the exported Verilog, and we also, as far as I recall, currently don't guarantee that a certain port will have a certain name unless it's named explicitly or there are no duplicates, but the latter is another property that you can't necessarily know about the design or enforce.

So I think we should either allow that or prohibit duplicate names too. (I think allowing this is very low friction and it reduces the annoyance if you just want to look at the Verilog but don't care about synthesizing it; like in the playground for example.)

tpwrules: I'm personally fine with allowing it, I was specifically asked to add it here: https://github.com/amaranth-lang/amaranth/pull/1234#discussion_r1536674341

wanda-phi: are you sure it doesn't break things? have you tried with *two* anonymous top-level ports? because I'm reasonably sure it'll just break, as we have no further pass that'd assign a name within NIR, and we just end up with `None` as the name and a key collision in some dict

the auto-naming you're seeing is an RTLIL backend artifact

tpwrules: You are correct, with two private top-level ports, only one shows up in the Verilog. If we want to remove the restriction, I'd say teach `_add_name` to handle a name of `""` and give it an index.

whitequark: I'm fine with a restriction, I just don't want it to be artificial. Private names are going to be rarely used anyways.

whitequark (approved):
Thanks a lot!

PR #1235, opened 2024-03-23
Enforce naming rules on core HDL

### LINKED ISSUE #1209: Zero-length submodule name breaks things
The following design:

```
from amaranth import *
from amaranth.back import verilog

class Bar(Elaboratable):
        m = Module()
        m.d.sync += Signal().eq(1)
        return m

class Foo(Elaboratable):
        m = Module()
        m.submodules[""] = Bar()
        return m

print(verilog.convert(Foo(), ports=[]))
```

gives the following exception:
```
  File "~/test.py", line 16, in <module>
    print(verilog.convert(Foo(), ports=[]))
  File "~/venv/lib/python3.10/site-packages/amaranth/back/verilog.py", line 61, in convert
    verilog_text, name_map = convert_fragment(fragment, ports, name, emit_src=emit_src, strip_internal_attrs=strip_internal_attrs, **kwargs)
  File "~/venv/lib/python3.10/site-packages/amaranth/back/verilog.py", line 39, in convert_fragment
    rtlil_text, name_map = rtlil.convert_fragment(*args, **kwargs)
  File "~/venv/lib/python3.10/site-packages/amaranth/back/rtlil.py", line 1029, in convert_fragment
    empty_checker=empty_checker).emit()
  File "~/venv/lib/python3.10/site-packages/amaranth/back/rtlil.py", line 308, in emit
    self.emit_submodules()
  File "~/venv/lib/python3.10/site-packages/amaranth/back/rtlil.py", line 516, in emit_submodules
    self.builder.cell(f"\\{dotted_name}", submodule.name[-1], ports={
  File "~/venv/lib/python3.10/site-packages/amaranth/back/rtlil.py", line 166, in cell
    name = self._make_name(name, local=False)
  File "~/venv/lib/python3.10/site-packages/amaranth/back/rtlil.py", line 67, in _make_name
    elif not local and name[0] not in "\\$":
IndexError: string index out of range
```

I suggest that `m.submodules[""] = ...` have the same behavior as `m.submodules += ...`. While it would be silly to directly write the former, it might arise from programmatic name generation, and should either work properly or immediately throw an error.

tpwrules:
The rules are quite simple and liberal:
* Can't be None or the empty string (case-specific exceptions)
* Can't contain a Unicode separator character
* Can't contain a Unicode control character, undefined character, or surrogate character

I did not bother updating the deprecated `_mem` or `_rec` modules.

Pretty much all cases I tried using a name illegal by these rules, it caused a problem somewhere else (e.g. an RTLIL syntax error).

Maybe worth adding some logic to Fragment to catch any cases that slipped through?

Closes #1209


tpwrules:
Apparently having arbitrary state names is desirable, so those will be allowed again once https://github.com/amaranth-lang/amaranth/pull/1234 is merged.

whitequark (commented):
My overall feelings about this PR:
- The general approach with the use of `unicodedata` seems good.
- I don't like having `none_ok` and `empty_ok`. There are already complex conditionals around the use of names in multiple places; adding an invisible at the call site complex conditional to this isn't making it easier to understand.
- I don't like how the error messages are no longer greppable because they are composed from multiple parts. I've tried to avoid this elsewhere because it makes it harder to figure out exactly where a message comes from if you don't have a backtrace (eg someone stringified an exception, or you just want to find every place where it's mentioned).
- I don't like that "what" is duplicated several times per entity (e.g. "Clock domain name" is independently used 3 times). I think a private class method that would encapsulate the naming logic would do better than the two boolean flags which despite having 4 states reflect 3 possible desired combinations.

I think for 0.5 I just want the linked issue fixed as a one-off because this PR has too many moving parts that I'm uncertain about.

PR #1245, opened 2024-03-25
Emit signal groups in GTKW files

### LINKED ISSUE #764: GTKW files with signal groups
For debugging complex circuits, it's useful to have the signal traces grouped according to their function. The popular _gtkwave_ tool supports this, as does the `GTKWSave` class from the `vcd` package used by Amaranth to export VCD files. All that remains is to connect this together.

I have [implemented an extension](https://github.com/kuznia-rdzeni/coreblocks/blob/master/transactron/testing/gtkw_extension.py) to Amaranth's `_VCDWriter` which does just that. The idea is that the `traces` argument can be passed other things than lists of signals. My implementation currently accepts the following as traces:

* Signals, which work as before.
* Iterables (e.g. lists) of traces, which shows them in _gtkwave_ one after the other, like before.
* Mappings (e.g. dicts) of traces, which create a new group of traces for each item in the mapping,
* Amaranth Records, which group the fields of the record.

There is no support in the current implementation for the recently added `lib.data` classes from RFC 1, but it can be easily added.

Are you interested in merging this functionality to Amaranth?

tilk:
Fixes #764.

Things to look at in the review:
* Group names for views and interfaces. The current code uses "view" and "interface". Maybe the name should reference a layout/signature in some way?
* Treating single-field view as in https://github.com/amaranth-lang/amaranth/issues/764#issuecomment-2014220740 is not implemented. It could be implemented by adding a special case.
* I've added tests that check if the code doesn't raise an exception, the resulting gtkw files are not checked. I'm not sure how to best check them; I could compare them with a reference, but this would freeze the signal naming. Is signal naming for views and interfaces specified?
* Had to add another change related to the recent private names commit. Now that there are three separate recursive functions in this PR, maybe the recursion pattern should be abstracted away?

### DIFF
--- a/amaranth/sim/core.py
+++ b/amaranth/sim/core.py
@@ -210,17 +210,25 @@ class Simulator:
                     file.close()
             raise ValueError("Cannot start writing waveforms after advancing simulation time")
 
-        for trace in traces:
-            if isinstance(trace, ValueLike):
-                trace_cast = Value.cast(trace)
+        def traverse_traces(traces):
+            if isinstance(traces, ValueLike):
+                trace_cast = Value.cast(traces)
                 if isinstance(trace_cast, MemoryData._Row):
-                    continue
+                    return
                 for trace_signal in trace_cast._rhs_signals():
                     if trace_signal.name == "":
-                        if trace_signal is trace:
+                        if trace_signal is traces:
                             raise TypeError("Cannot trace signal with private name")
                         else:
-                            raise TypeError(f"Cannot trace signal with private name (within {trace!r})")
+                            raise TypeError(f"Cannot trace signal with private name (within {traces!r})")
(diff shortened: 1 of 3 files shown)


whitequark:
@tilk Could you rebase the PR please? We've finally landed `MemoryData`, a proper replacement for `MemoryIdentity` that also improves tracing experience, but this caused a conflict here.

whitequark:
@tilk, the rework of the simulator necessary for the RFC 36 has introduced some merge conflicts here--could you fix them please?

whitequark (approved):
LGTM, thank you!

whitequark:
Could you squash the PR please? The commit message could be something like

```
sim: group signal traces according to their function.
```

whitequark:
Thanks!

PR #1247, opened 2024-03-26
hdl._ir: rename `Instance.named_ports` to `Instance.ports`.

wanda-phi:
Made possible by the new port propagation code freeing up the name.


whitequark (approved):
Thanks!

PR #1256, opened 2024-04-02
hdl._mem: add `MemoryData` class.

wanda-phi:
This implements half of RFC 62. The `MemoryData._Row` class will be implemented later, as a follow-up.

### DIFF
--- a/amaranth/hdl/_mem.py
+++ b/amaranth/hdl/_mem.py
@@ -1,33 +1,160 @@
 import operator
 from collections import OrderedDict
+from collections.abc import MutableSequence
 
 from .. import tracer
 from ._ast import *
 from ._ir import Elaboratable, Fragment
 from ..utils import ceil_log2
-from .._utils import deprecated
+from .._utils import deprecated, final
 
 
-__all__ = ["Memory", "ReadPort", "WritePort", "DummyPort"]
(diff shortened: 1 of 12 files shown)


amaranth/hdl/_mem.py
```diff
@@ -1,33 +1,158 @@
...
 
-class MemoryIdentity: pass
+@final
+class FrozenError(Exception):
+    """This exception is raised when ports are added to a :class:`Memory` or its
+    :attr:`~Memory.init` attribute is changed after it has been elaborated once.
+    """
+
+
+class MemoryData:
```
whitequark, line 22:
```suggestion
@final
class MemoryData:
```

amaranth/hdl/_mem.py
```diff
@@ -1,33 +1,158 @@
...
+    class Init(MutableSequence):
```
whitequark, line 23:
```suggestion
    @final
    class Init(MutableSequence):
```

amaranth/hdl/_mem.py
```diff
@@ -1,33 +1,158 @@
...
+        """Memory initialization data.
+
+        This is a special container used only for the :attr:`Memory.init` attribute. It is similar
```
whitequark, line 26: This should not refer to `Memory.init` since that's from `lib`, not `hdl`. Maybe "This is a special container used only for initial contents of memories."?

PR #1264, opened 2024-04-03
hdl._ast: add `SwitchValue`, reimplement `ArrayProxy` with it.

wanda-phi:


amaranth/hdl/_ast.py
```diff
@@ -1892,6 +1892,60 @@ def __repr__(self):
...
 
+@final
+class SwitchValue(Value):
+    def __init__(self, test, cases, *, src_loc=None, src_loc_at=0):
+        if src_loc is None:
+            super().__init__(src_loc_at=src_loc_at)
+        else:
+            self.src_loc = src_loc
+        self._test = Value.cast(test)
+        self._cases = []
```
whitequark, line 1903: This instance variable shouldn't be changing its type in the middle of the constructor.

amaranth/hdl/_ast.py
```diff
@@ -2356,10 +2410,17 @@ def __repr__(self):
                                        ", ".join(map(repr, self._inner)))
 
 
+def _proxy_value(name):
+    def inner(self, *args, **kwargs):
```
whitequark, line 2414: This should probably use `@functools.wraps(getattr(Value, name))`

whitequark (approved):
Thanks!

PR #1265, opened 2024-04-03
hdl._ir: raise an error when an elaboratable is duplicated in hierarchy.

### LINKED ISSUE #1194: Detect and warn about elaboratables added more than once into the hierarchy

wanda-phi:
Fixes #1194.


whitequark (approved):
Thanks!

PR #1271, opened 2024-04-03
hdl._mem: implement `MemoryData._Row` from RFC 62.

wanda-phi:


amaranth/hdl/_dsl.py
```diff
@@ -13,11 +13,90 @@
...
 from ._xfrm import *
+from ._mem import MemoryData
 
 
 __all__ = ["SyntaxError", "SyntaxWarning", "Module"]
 
 
+class _Visitor:
+    def __init__(self):
+        self.lhs_signals = SignalSet()
```
whitequark, line 24:
```suggestion
        self.driven_signals = SignalSet()
```

whitequark (approved):
Thanks!

PR #1280, opened 2024-04-04
back.rtlil: implement remaining format specifiers.

wanda-phi:
This requires a Yosys version from git. The requirement should be bumped to a proper release before Amaranth 0.5.


whitequark (approved):
Thanks!

PR #1284, opened 2024-04-05
vendor._altera: implement `lib.io` buffer primitives.

wanda-phi:


whitequark (approved):

tpwrules:
Switched my design to this branch and it works well with Quartus 20.1.1.720. It only uses basic I/O though, not even bidirectional.

amaranth/vendor/_altera.py
```diff
@@ -293,247 +445,21 @@ def create_missing_domain(self, name):
...
-        return m
+    def get_io_buffer(self, buffer):
+        if isinstance(buffer, io.Buffer):
+            res = IOBuffer(buffer.direction, buffer.port)
+        elif isinstance(buffer, io.FFBuffer):
+            res = FFBuffer(buffer.direction, buffer.port)
+        elif isinstance(buffer, io.DDRBuffer):
+            res = DDRBuffer(buffer.direction, buffer.port)
+        else:
+            raise TypeError(f"Unsupported buffer type {buffer!r}")
```
whitequark, line 456:
```suggestion
            raise TypeError(f"Unsupported buffer type {buffer!r}") # :nocov:
```

amaranth/vendor/_altera.py
```diff
@@ -1,9 +1,161 @@
...
+            if direction is not io.Direction.Input:
+                self.oe.attrs["useioff"] = 1
+
+    def elaborate(self, platform):
+        kwargs = dict(
+            p_enable_bus_hold="FALSE",
+            p_number_of_channels=len(self.port),
+        )
+        if isinstance(self.port, io.SingleEndedPort):
+            kwargs["p_use_differential_mode"] = "FALSE"
```
whitequark, line 46: You don't actually need to use this--[`Instance` accepts arguments of the form `("p", "blah", value)`](https://amaranth-lang.org/docs/amaranth/latest/guide.html#instances).

whitequark (approved):
LGTM. This is still not perfect but the Altera uses will simply share our suffering.

PR #1296, opened 2024-04-09
run: fix `execute_local_docker()` not properly catching ctrl-c

### LINKED ISSUE #1246: `execute_local_docker()` does not forward SIGINT/ctrl+c to docker subprocess
I use the `execute_local_docker()` build command almost exclusively in my current project. I have noticed that if I ctrl+c a build the docker subprocess that is spawned by Amaranth doesnt exit. This is undesirable because it can lead to mistakenly spawning two docker containers which are both actively modifying the build directory.

polymerizedsage:
fix #1246.

### DIFF
--- a/amaranth/build/run.py
+++ b/amaranth/build/run.py
@@ -9,6 +9,7 @@ import warnings
 import zipfile
 import hashlib
 import pathlib
+import random
 
 
 __all__ = ["BuildPlan", "BuildProducts", "LocalBuildProducts", "RemoteSSHBuildProducts"]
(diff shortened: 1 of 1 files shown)


amaranth/build/run.py
```diff
@@ -138,6 +138,7 @@ def execute_local_docker(self, image, *, root="build", docker_args=[]):
         subprocess.check_call([
             "docker", "run", *docker_args,
             "--rm", # remove the container after running
+           "-it" # open container with interactive terminal session, allowing it to properly catch ctrl-C exit from user.
```
whitequark, line 141: Can this not cause more problems? What happens if you run Amaranth itself from a script? I vaguely recall removing `-it` from the option list before.

polymerizedsage:
switched to using `docker stop` instead using a pre-determined container name. Kind of just chose the name arbitrarily, if you think there is a more appropriate name just lmk.

whitequark:
That will break a pair of concurrent builds, no?

whitequark (approved):
LGTM, thanks!

whitequark:
Squash the PR please.

PR #1308, opened 2024-04-10
lib.fifo: remove dependency on `lib.coding`.

wanda-phi:


whitequark (approved):
The fact that this is a negative diffstat is a perfect nail in the coffin of `lib.coding`.

PR #1312, opened 2024-04-10
lib.data: accept `data.Const` in `*.const()`.

wanda-phi:


whitequark (approved):
This makes `const` somewhat more complex but it's probably unavoidable.

PR #1318, opened 2024-04-11
CI: fix coverage submission

whitequark:


whitequark:
So after talking to Codecov folks this should be the right solution after all, but for the time being forks will need to authorize the app too.

PR #1322, opened 2024-04-11
hdl._ast: add enum name argument to `Format.Enum`.

wanda-phi:
Turns out that RTLIL enum representation requires such, so add a place to store it.


whitequark (approved):
Thanks!

PR #1330, opened 2024-04-13
hdl._nir: add combinational cycle detection.

### LINKED ISSUE #704: Detect and reject netlists with combinatorial loops

### LINKED ISSUE #1143: infinite loop in NIR builder on a trivial comb loop
The following results in an infinite loop:

```
from amaranth import *
import amaranth.back.rtlil

m = Module()
a = Signal()
m.d.comb += a.eq(a)

amaranth.back.rtlil.convert(m, ports=[])
```

wanda-phi:
Fixes #704.
Fixes #1143.

### DIFF
--- a/amaranth/hdl/_nir.py
+++ b/amaranth/hdl/_nir.py
@@ -1,4 +1,4 @@
-from typing import Iterable
+from typing import Iterable, Any
 import enum
 
 from ._ast import SignalDict
@@ -7,8 +7,9 @@ from . import _ast
 
 __all__ = [
     # Netlist core
+    "CombinationalCycle",
     "Net", "Value", "IONet", "IOValue",
-    "FormatValue", "Format",
+    "FormatValue", "Format", "SignalField",
     "Netlist", "ModuleNetFlow", "IODirection", "Module", "Cell", "Top",
     # Computation cells
     "Operator", "Part",
@@ -23,6 +24,10 @@ __all__ = [
     # Foreign interface cells
     "Instance", "IOBuffer",
 ]
+
+
+class CombinationalCycle(Exception):
+    pass
 
 
 class Net(int):
@@ -335,6 +340,7 @@ class Netlist:
     modules : list of ``Module``
     cells : list of ``Cell``
     connections : dict of (negative) int to int
+    late_to_signal : dict of (late) Net to its Signal and bit number
     io_ports : list of ``IOPort``
     signals : dict of Signal to ``Value``
     signal_fields: dict of Signal to dict of tuple[str | int] to SignalField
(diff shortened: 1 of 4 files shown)


tests/test_hdl_ir.py
```diff
@@ -3542,3 +3542,16 @@ class MyEnum(enum.Enum, shape=unsigned(2)):
...
+class CycleTestCase(FHDLTestCase):
+    def test_cycle(self):
+        a = Signal()
+        b = Signal()
+        m = Module()
+        m.d.comb += [
+            a.eq(~b),
+            b.eq(~a),
+        ]
+        with self.assertRaises(CombinationalCycle):
```
whitequark, line 3556: This should probably assert on a message using `assertRaisesRegex` to skip the variable parts like the line numbers.

amaranth/hdl/_nir.py
```diff
@@ -698,6 +804,10 @@ def __repr__(self):
         patterns = " ".join(self.patterns)
         return f"(matches {self.value} {patterns})"
 
+    def comb_edges_to(self, bit: int) -> Iterable[Net]:
```
whitequark, line 807: The annotation on the base class is probably sufficient.

amaranth/hdl/_nir.py
```diff
@@ -733,6 +843,11 @@ def resolve_nets(self, netlist: Netlist):
     def __repr__(self):
         return f"(priority_match {self.en} {self.inputs})"
 
+    def comb_edges_to(self, bit: int) -> Iterable[Net]:
```
whitequark, line 846: Likewise.

amaranth/hdl/_nir.py
```diff
@@ -405,16 +412,81 @@ def add_value_cell(self, width: int, cell):
...
+                        sig, bit = self.late_to_signal[net]
+                        e.path.append((sig, bit, sig.src_loc))
+                        raise
+                else:
+                    for src, src_loc in self.cells[net.cell].comb_edges_to(net.bit):
+                        try:
+                            traverse(src)
+                        except Cycle as e:
+                            e.path.append((self.cells[net.cell], net.bit, src_loc))
+                            raise
```
whitequark, line 462: Please use a local return instead of a nonlocal return, since using a nonlocal return adds nothing.

whitequark (changes requested):
See inline comments.

whitequark (approved):
LGTM, thanks!

widlarizer:
Sweet!

PR #1334, opened 2024-04-15
hdl._dsl: improve error message on `m.domains.cd_foo = ...`.

### LINKED ISSUE #1331: Unclear error message on `m.domains.cd_tx = ClockDomain()`
One could reasonably write the following code:

```py
m.domains.cd_tx = ClockDomain()
```

This fails with:

```
NameError: Clock domain name 'rx' must match name in `m.domains.cd_rx += ...` syntax
```

It fails because the `cd_` prefix is stripped by `ClockDomain()` but not `m.domains.x = `. The correct syntax is is:

```py
m.domains.cd_tx = ClockDomain()
```

This is a rather confusing diagnostic and should be improved.

wanda-phi:
Fixes #1331.


whitequark (approved):
Thanks!

PR #1335, opened 2024-04-15
sim: implement `Format.*` for memories in VCD.

wanda-phi:


whitequark (approved):
LGTM, thanks! There's some duplication going on but it's not catastrophic.

PR #1338, opened 2024-04-18
vendor._lattice: merge ECP5 and MachXO[23] into one platform.

wanda-phi:


whitequark (approved):
Excellent, thank you!

Issue #1339, opened 2024-04-18
Can't emit Verilog/RTLIL with Signature array members


eigenform:
When converting to Verilog or RTLIL, the port names are built from a path which is assumed to be a collection of `str`. 
However, for array members in a `Signature`, the index of an element in the path is an `int`.

```python
#!/usr/bin/env python

import amaranth
from amaranth import *
from amaranth.lib.wiring import *
from amaranth.back import verilog

class MySignature(Signature):
    def __init__(self):
        super().__init__({
            "output": Out(4),
        })

class MyComponent(Component):
    def __init__(self):
        signature = Signature({
            "foo": Out(MySignature()).array(4),
        })
        super().__init__(signature)

    def elaborate(self, platform):
        m = Module()
        m.d.comb += [
            self.foo[0].output.eq(0b0001),
            self.foo[1].output.eq(0b0010),
            self.foo[2].output.eq(0b0100),
            self.foo[3].output.eq(0b1000),
        ]
        return m

dut = MyComponent()
print(amaranth.__version__)
for (path, member, value) in dut.signature.flatten(dut):
    print(path, member, value)

v = verilog.convert(dut)
print(v)
```

```
0.5.0.dev273+20240418.201156
('foo', 0, 'output') Out(4) (sig foo__0__output)
('foo', 1, 'output') Out(4) (sig foo__1__output)
('foo', 2, 'output') Out(4) (sig foo__2__output)
('foo', 3, 'output') Out(4) (sig foo__3__output)
Traceback (most recent call last):
  File "/tmp/foo.py", line 36, in <module>
    v = verilog.convert(dut)
        ^^^^^^^^^^^^^^^^^^^^
  File "/home/meta/git/eigenform/amaranth/amaranth/back/verilog.py", line 58, in convert
    ports["__".join(path)] = (value, dir)
          ^^^^^^^^^^^^^^^
TypeError: sequence item 1: expected str instance, int found
```

whitequark:
Ah, yes. They should be mapped to `str`.

PR #1340, opened 2024-04-18
back.verilog, back.rtlil: map path elements to str

eigenform:
Easy fix for #1339 - let me know if you had something else in mind.


amaranth/back/rtlil.py
```diff
@@ -1289,7 +1289,7 @@ def convert(elaboratable, name="top", platform=None, *, ports=None, emit_src=Tru
                     dir = _ir.PortDirection.Input
                 else:
                     dir = _ir.PortDirection.Output
-                ports["__".join(path)] = (value, dir)
+                ports["__".join([str(s) for s in path])] = (value, dir)
```
whitequark, line 1292: I'd probably just do `map(str, path)`

whitequark (approved):

PR #759, opened 2023-03-02
add Lattice Nexus vendor platform (using nextpnr-nexus and prjoxide)

slagernate:
sorry, github newb. Won't do that again. Tested again with blinking all LEDs. Output:
```
python3 -m amaranth_boards.lifcl_evn
init..
IDCODE: 0x110f1043 (LIFCL-40)
NX Status Register: 0x0000150100400100
reset..
NX Status Register: 0x0000150100400e50
programming..
NX Status Register: 0x0000150100400100
Bye.
```
Notes:
--Removed mention of ECP5 in `lattice_nexus.py`
--Removed OSC

PR #1342, opened 2024-04-19
vendor._lattice: add Nexus support.

wanda-phi:
This is based on #759, but refactored to use the unified `LatticePlatform` instead of adding a new one.

Validated on LFICL-40-EVN using slightly modified amaranth-lang/amaranth-boards#217


whitequark (approved):
LGTM, thanks for the excellent PR!

PR #1343, opened 2024-04-19
vendor._lattice: Add clock constraints to oxide .pdc template.

polymerizedsage:
Use `create_clock` to set clock constraints in .pdc file generated for oxide as per: https://github.com/amaranth-lang/amaranth/pull/759#discussion_r1561582932.


wanda-phi (approved):
r+

whitequark (approved):

PR #1346, opened 2024-05-04
vendor._lattice: OFS1P3DX -> OFD1P3DX and IFS1P3DX -> IFD1P3DX for nexus

richardeoin:
Lattice renamed some IO primitives when moving from Diamond to Radiant. This is, naturally, sparsely documented but I found a reference in Table 41 of [^1]. One letter in the name of each primitives is changed, as well as one of the inputs (`i_SCLK` -> `i_CK`). This requires a different IO Buffer specialisation for Nexus (Radiant) compared to MachXO2 and ECP5 which this PR adds.

I don't quite know which name Oxide would expect, but a quick grep among the fuzzing scripts finds the updated name (same as Radiant) and not the previous one. So I assume this change is applicable to Oxide also.

Support for Nexus was recently added in #1342

[^1]: https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/RZ2/Radiant2_1_migration_guide_diamond.ashx?document_id=52926

### DIFF
--- a/amaranth/vendor/_lattice.py
+++ b/amaranth/vendor/_lattice.py
@@ -77,7 +77,7 @@ class IOBuffer(io.Buffer):
         return m
 
 
-def _make_oereg(m, domain, oe, q):
+def _make_oereg_ecp5_machxo2(m, domain, oe, q):
     for bit in range(len(q)):
         m.submodules[f"oe_ff{bit}"] = Instance("OFS1P3DX",
             i_SCLK=ClockSignal(domain),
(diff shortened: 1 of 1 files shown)


amaranth/vendor/_lattice.py
```diff
@@ -87,8 +87,18 @@ def _make_oereg(m, domain, oe, q):
             o_Q=q[bit],
         )
 
+def _make_oereg_nexus(m, domain, oe, q):
```
whitequark, line 90: Nit: missing whitespace.

amaranth/vendor/_lattice.py
```diff
@@ -118,7 +128,41 @@ def elaborate(self, platform):
                     i_D=o_inv[bit],
                     o_Q=buf.o[bit],
                 )
-            _make_oereg(m, self.o_domain, ~self.oe, buf.t)
+            _make_oereg_ecp5_machxo2(m, self.o_domain, ~self.oe, buf.t)
+
+        return m
+
+class FFBufferNexus(io.FFBuffer):
```
whitequark, line 135: Same here (two newlines between toplevel entities).

amaranth/vendor/_lattice.py
richardeoin: Fixed

wanda-phi (approved):
r+

whitequark (approved):
r=@wanda-phi

PR #1348, opened 2024-05-05
vendor.*: fix `[io]_domain` propagation when creating buffers.

wanda-phi:
Also makes `[io]_domain` attributes always 

Having conditionally-present attributes causes more problems than it's worth (see #1347). Just make them contain `None` when irrelevant.

Closes #1347.


whitequark (approved):
LGTM, thanks!

PR #1352, opened 2024-05-07
hdl._dsl: raise an error when modifying an already-elaborated `Module`.

### LINKED ISSUE #1350: Make `Module` explode if you try to modify it after it's elaborated
This protects against a beginner mistake where `m` is assigned to e.g. an attribute and used out of scope. Elaboration should consume the module and make it immutable.

wanda-phi:
Fixes #1350.

### DIFF
--- a/amaranth/hdl/_dsl.py
+++ b/amaranth/hdl/_dsl.py
@@ -288,8 +288,11 @@ class Module(_ModuleBuilderRoot, Elaboratable):
         self._domains      = {}
         self._generated    = {}
         self._src_loc      = tracer.get_src_loc()
+        self._frozen       = False
 
     def _check_context(self, construct, context):
+        if self._frozen:
+            raise AlreadyElaborated(f"Cannot modify a module that has already been elaborated")
         if self._ctrl_context != context:
             if self._ctrl_context is None:
                 raise SyntaxError("{} is not permitted outside of {}"
@@ -616,6 +619,9 @@ class Module(_ModuleBuilderRoot, Elaboratable):
                     src_loc=src_loc))
 
     def _add_statement(self, assigns, domain, depth):
+        if self._frozen:
+            raise AlreadyElaborated(f"Cannot modify a module that has already been elaborated")
+
         while len(self._ctrl_stack) > self.domain._depth:
             self._pop_ctrl()
 
@@ -643,6 +649,8 @@ class Module(_ModuleBuilderRoot, Elaboratable):
         if not hasattr(submodule, "elaborate"):
             raise TypeError("Trying to add {!r}, which does not implement .elaborate(), as "
                             "a submodule".format(submodule))
+        if self._frozen:
+            raise AlreadyElaborated(f"Cannot modify a module that has already been elaborated")
         if name == None:
             self._anon_submodules.append((submodule, src_loc))
         else:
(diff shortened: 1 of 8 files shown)


amaranth/hdl/_dsl.py
```diff
@@ -16,7 +16,12 @@
 from ._mem import MemoryData
 
 
-__all__ = ["SyntaxError", "SyntaxWarning", "Module"]
+__all__ = ["FrozenModule", "SyntaxError", "SyntaxWarning", "Module"]
+
+
+@final
+class FrozenModule(Exception):
```
whitequark, line 23: I think "frozen" doesn't make sense as a designer-facing concept. Bikeshed:
- `ElaboratedModule`?
- `AlreadyElaborated`?
- something else?

wanda-phi: yeah, I was going for consistency with `FrozenMemory`

I like `AlreadyElaborated`, do you think we should use it here and also replace `FrozenMemory` with it?

whitequark: SGTM

wanda-phi: done; had to move the exception to `_ir` to avoid a circular import between `_dsl` and `_mem`

amaranth/hdl/_dsl.py
```diff
@@ -288,8 +288,11 @@ def __init__(self):
         self._domains      = {}
         self._generated    = {}
         self._src_loc      = tracer.get_src_loc()
+        self._frozen       = False
 
     def _check_context(self, construct, context):
+        if self._frozen:
+            raise AlreadyElaborated(f"Cannot use control structures on a module that has already been elaborated")
```
whitequark, line 295: We definitely don't need unique diagnostics for all of these cases, as it's the same underlying issue and it is unimportant how the module is modified. (Otherwise you might conclude that there are some other modifications that are permitted, whereas none are.)

```suggestion
            raise AlreadyElaborated(f"Cannot modify a module that has already been elaborated")
```

amaranth/hdl/_ir.py
```diff
@@ -3,17 +3,23 @@
...
-    "Instance", "IOBufferInstance", "PortDirection", "Design", "build_netlist",
+    "AlreadyElaborated", "UnusedElaboratable", "Elaboratable", "DuplicateElaboratable",
+    "DriverConflict", "Fragment", "Instance", "IOBufferInstance", "PortDirection", "Design",
+    "build_netlist",
 ]
 
 
+@final
+class AlreadyElaborated(Exception):
+    """Exception raised when a module is being modified after elaboration."""
```
whitequark, line 20: A memory is not a module.

```suggestion
    """Exception raised when an elaboratable is being modified after elaboration."""
```

amaranth/hdl/_mem.py
```diff
@@ -101,7 +96,7 @@ def __getitem__(self, index):
 
         def __setitem__(self, index, value):
             if self._frozen:
-                raise FrozenMemory("Cannot set 'init' on a memory that has already been elaborated")
+                raise AlreadyElaborated("Cannot set 'init' on a memory that has already been elaborated")
```
whitequark, line 99: (This diagnostic and the one in the next file over is probably okay as-is, because "modify a memory" is ambiguous on whether it refers to compile-time or run-time modification.)

docs/stdlib/memory.rst
```diff
@@ -181,8 +181,6 @@ Simulation
 Memory description
 ==================
 
-.. autoexception:: amaranth.hdl.FrozenMemory
```
whitequark, line 184: This is now not documented at all, is it? How about adding `:noindex:` instead, so that you get the docstring here, but not the link target? Then some other place could be the canonical one for it.

whitequark (changes requested):
See inline comments.

whitequark (approved):
LGTM, thanks!

PR #1354, opened 2024-05-08
hdl._ir: derive `DriverConflict` from `Exception` instead of `UserWarning`.

wanda-phi:


whitequark:
We're not using it as a warning _ever_ anymore, right?

wanda-phi:
> We're not using it as a warning _ever_ anymore, right?

yup.

whitequark (approved):
LGTM, thanks!

PR #1355, opened 2024-05-08
hdl._ir, lib, vendor: add `RequirePosedge`, use it whenever required.

wanda-phi:


amaranth/hdl/__init__.py
```diff
@@ -6,8 +6,8 @@
 from ._ast import IOValue, IOPort
 from ._dsl import Module
 from ._cd import DomainError, ClockDomain
-from ._ir import AlreadyElaborated, UnusedElaboratable, Elaboratable, DriverConflict, Fragment
-from ._ir import Instance, IOBufferInstance
+from ._ir import AlreadyElaborated, UnusedElaboratable, Elaboratable, DriverConflict, DomainRequirementFailed
```
whitequark, line 9: Should this be exported? I don't want people catching this exception without importing it from a private module.

amaranth/hdl/_ir.py
```diff
@@ -385,6 +391,16 @@ def __init__(self, port, *, i=None, o=None, oe=None, src_loc_at=0, src_loc=None)
         self.src_loc = src_loc or tracer.get_src_loc(src_loc_at)
 
 
+class RequirePosedge(Fragment):
```
whitequark, line 394: This should clearly be marked as a temporary measure that can be removed at any point.

amaranth/hdl/_ir.py
```diff
@@ -662,6 +681,18 @@ def _assign_names(self, fragment: Fragment, hierarchy: "tuple[str]"):
             subfragment_name = _add_name(frag_info.assigned_names, subfragment_name)
             self._assign_names(subfragment, hierarchy=(*hierarchy, subfragment_name))
 
+    def _check_domain_requires(self):
+        for fragment, fragment_info in self.fragments.items():
+            if isinstance(fragment, RequirePosedge):
+                domain = fragment.domains[fragment._domain]
+                if domain.clk_edge != "pos":
+                    if fragment.src_loc is None:
+                        src_loc = "[unknown]"
```
whitequark, line 690:
```suggestion
                        src_loc = "<unknown>:0"
```

whitequark (changes requested):

amaranth/hdl/__init__.py
wanda-phi: Right, good point.

amaranth/hdl/_ir.py
```diff
@@ -40,6 +41,11 @@ class DuplicateElaboratable(Exception):
     pass
 
 
+class DomainRequirementFailed(Exception):
+    """Raised when a module has unsatisfied requirements about a clock domain, such us getting
```
whitequark, line 45:
```suggestion
    """Raised when a module has unsatisfied requirements about a clock domain, such as getting
```

whitequark (approved):
LGTM

PR #1357, opened 2024-05-10
Replace use of _reset with _init in Xilinx platform.

bl0x:
This looks like an oversight, when _reset was replaced with _init in FFSynchronizer.


wanda-phi:
r+

this is an oversight, but also while we're at it, shouldn't `_altera.py` like... actually look at the init value?

whitequark:
@bl0x Could you change the commit message to match our format please?

```
vendor._xilinx: ...
```

in this case.

bl0x:
Oops, fixed.

whitequark (approved):

PR #1359, opened 2024-05-12
Clarify documentation related to `Array` and `ArrayLayout`

whitequark:


docs/guide.rst
```diff
@@ -811,7 +811,11 @@ Crucially, this means that any Python object can be added to an array; the only
 
 .. warning::
 
-    Each time an array proxy object with :py:`n` elements is used in an expression, it generates a multiplexer with :py:`n` branches. However, using :py:`k` of such array proxy objects in an expression generates a multiplexer with :py:`n**k` branches. This can generate extremely large circuits that may quickly exhaust the resources of the synthesis target or even the available RAM.
```
wanda-phi, line 814: I don't think that's true anymore

whitequark: Can you elaborate?

wanda-phi: it got fixed with the new NIR-based RTLIL backend

whitequark: Oh! Right

PR #1380, opened 2024-06-10
Update install.rst to specify how to install board definitions

cHemingway:
Source for this being the correct way: https://libera.irclog.whitequark.org/amaranth-lang/2023-09-21

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -318,6 +318,10 @@ Editable development snapshot
 Installing board definitions
 =============================
 
-.. todo::
+This must be installed directly from github: 
 
-	 Explain how to install `<https://github.com/amaranth-lang/amaranth-boards>`_.
+.. code-block:: console
+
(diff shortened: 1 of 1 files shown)


whitequark:
I'm aware that this is a pain point for new users of Amaranth but I'm also not sure if the current installation page is what we should continue to present, given that it only works in a virtual environment if you're on a modern Linux system (Debian, Ubuntu, WSL 2). It's possible that the right thing to do is to recommend using PDM, but it might also not be.

I'm not going to merge this within the 0.5 release cycle but we should definitely look into improving installation instructions for the 0.6 release cycle.

cHemingway:
I found this particular step worked for me on a MacBook, using a venv but not pdm. 

Other steps didn't, (I needed to set some environmental variables to point out the toolchain and my current hack of doing this via setting os.environ is not ideal) so I can see why you would want to just rewrite the installation guide instead.

whitequark:
Take a look at https://github.com/amaranth-lang/template-fpga, it could be a useful starting point.

PR #1383, opened 2024-06-14
build: get list of used I/O ports from `Design`.

### LINKED ISSUE #1365: Constraint generation from platform resource with `DiffPairs`  of type LVDS incorrect.
When helping someone out with something involving Amaranth, I noticed that the latest main does not generate proper constraints for LVDS `DiffPairs`, causing place and route to fail due to it wanting to bind to both signals for the buffer and not just the non-inverted end of the signal.

In this case it was using the [VersaECP5Platform](https://github.com/amaranth-lang/amaranth-boards/blob/main/amaranth_boards/versa_ecp5.py) from `amaranth-boards`. Specifically that board has a differential LVDS clock as the default clock, and as it stands any design using that platform is unable to be fully placed and routed due to that.

Using the minimal blink example, the generated constraints for the LVDS signal are:
```
LOCATE COMP "clk100_0__p" SITE "P3";
IOBUF PORT "clk100_0__p" IO_TYPE=LVDS;
LOCATE COMP "clk100_0__n" SITE "P4";
IOBUF PORT "clk100_0__n" IO_TYPE=LVDS;
``` 

Which is not correct.

I ran a git bisect from when `v0.5.0.dev0` was first tagged (597b1b883924d8949061b52270eb55b97a7cfb76) up to `HEAD` from main, and it looks like commit cf534489a25d9402fb468eee011823c06f532d95 is the commit that caused the failure.

I've not done much more digging into the cause, but I figured y'all would like to know.

wanda-phi:
Fixes #1365.

### DIFF
--- a/amaranth/build/plat.py
+++ b/amaranth/build/plat.py
@@ -34,6 +34,7 @@ class Platform(ResourceManager, metaclass=ABCMeta):
         self.extra_files = OrderedDict()
 
         self._prepared   = False
+        self._design     = None
 
     @property
     def default_clk_constraint(self):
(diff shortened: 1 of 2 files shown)


whitequark (approved):
LGTM

PR #1385, opened 2024-06-14
sim: awaken all processes waiting on `changed()` at time 0.

wanda-phi:


whitequark (approved):
I'm not completely happy about the abstractions used here, but I can't come up with anything unambiguously better; LGTM.

PR #1401, opened 2024-06-16
docs/guide: remove extraneous "not only".

kivikakk:
Minor: perhaps intended initially as "not only with an int but also a value-like object".


whitequark (approved):
Thanks!

PR #1406, opened 2024-06-17
lib.data: make `len()` work on views of array layout.

### LINKED ISSUE #1403: Views with an array layout should have `len()` applicable to them
Currently this isn't possible:

```python
width = 8
sig = Signal(data.ArrayLayout(1, width))
len(sig) # => TypeError: object of type 'View' has no len()
```

wanda-phi:
Fixes #1403.

### DIFF
--- a/amaranth/lib/data.py
+++ b/amaranth/lib/data.py
@@ -850,6 +850,12 @@ class View(ValueCastable):
                 f"may only be accessed by indexing")
         return item
 
+    def __len__(self):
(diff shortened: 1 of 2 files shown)


whitequark (approved):
LGTM, thanks.

Does this enable iteration as well? Do we need a test for that?

wanda-phi:
it does not enable iteration, because iteration relies on indexing raising `IndexError` instead

PR #1411, opened 2024-06-18
lib.data: make views of `ArrayLayout` iterable. [0.5 backport]

wanda-phi:


wanda-phi:
Abandoning in favor of #1439.

PR #1420, opened 2024-06-27
Make `data.Const` comparable with `list` and `dict` objects compatible with the layout

### LINKED ISSUE #1414: Allow comparison of `data.Const` with `dict` and `list`
It would be nice, for tests, if this worked:

```python
>>> data.StructLayout({"a": 4}).const({"a": 0b0001}) == {"a": 0b0001}
  File ".../python3.12/site-packages/amaranth/lib/data.py", line 1076, in __eq__
    raise TypeError(
TypeError: Constant with layout StructLayout({'a': 4}) can only be compared to another view or constant with the same layout, not {'a': 1}
```

Currently you have to write assertions like this:

```
assert (await stream_get(ctx, stream)) == Const({"a", 0b0001}, stream.p.shape())
```

whitequark:
Before this commit, `data.Const` was only comparable to `data.Const` or `data.View`.

After this commit, in addition, it is also comparable to `dict` or `list` provided that such a value is accepted by `layout.const()` where `layout` is the layout of the `data.Const` object.

This change greatly reduces boilerplate in tests by enabling e.g.:

    assert (await stream_get(ctx, stream)) == {"value": 1}

instead of:

    assert (await stream_get(ctx, stream)) == Const({"value": 1}, stream.p.shape())

Note that, unlike `Layout.const`, which accepts arbitrary `Mapping` or `Sequence` objects, only `dict` and `list` are accepted in comparisons. Also, `data.View` continues to be comparable only to `data.View` and `data.Const`. This is to minimize the scope of the change and reduce likelihood of undesirable side effects when backported to the 0.5.x branch.

Fixes #1414.

### DIFF
--- a/amaranth/lib/data.py
+++ b/amaranth/lib/data.py
@@ -1083,9 +1083,18 @@ class Const(ValueCastable):
         elif isinstance(other, Const) and self.__layout == other.__layout:
             return self.__target == other.__target
         else:
+            cause = None
+            if isinstance(other, (dict, list)):
+                try:
+                    other_as_const = self.__layout.const(other)
+                except (TypeError, ValueError) as exc:
+                    cause = exc
+                else:
+                    return self == other_as_const
             raise TypeError(
-                f"Constant with layout {self.__layout!r} can only be compared to another view or "
-                f"constant with the same layout, not {other!r}")
+                f"Constant with layout {self.__layout!r} can only be compared to another view, "
+                f"a constant with the same layout, or a dictionary or a list that can be converted "
+                f"to a constant with the same layout, not {other!r}") from cause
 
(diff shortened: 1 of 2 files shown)


wanda-phi (approved):

whitequark:
For additional context, the main reason `View` comparisons are unimplemented are this code:

```python
end = Signal()
with m.If(stream.payload == {"data": 1, "end": end}):
    ...
```

Whether it should be allowed or not is *definitely* RFC-worthy, and @wanda-phi actually had a planned RFC that would bring the underlying functionality for composing a View out of a concatenation of parts. For the time being, I think it's simplest and safest to not touch `View` at all until we're ready.

PR #1434, opened 2024-06-29
Add `payload_init=` argument to `lib.stream.Signature()`

### LINKED ISSUE #1404: `stream.Signature` should allow setting the init value for the payload
This for example does not work:

```python
class BitSerializer(wiring.Component):
    def __init__(self, *, width, length):
        self._length = length

        super().__init__({
            "shift": In(1),
            "stream": In(stream.Signature(data.StructLayout({
                "o":  data.ArrayLayout(width, length),
                "oe": 1
            })), init={"payload": {"oe": 1}}),
            "buffer": Out(io.FFBuffer.Signature("o", width)),
        })
```

but could be valuable for the same reason `io.Buffer.Signature`'s `oe` member has `init=1`.

whitequark:
Fixes #1404.

### DIFF
--- a/amaranth/lib/stream.py
+++ b/amaranth/lib/stream.py
@@ -19,7 +19,9 @@ class Signature(wiring.Signature):
     Parameters
     ----------
     payload_shape : :class:`~.hdl.ShapeLike`
-        Shape of the payload.
+        Shape of the payload member.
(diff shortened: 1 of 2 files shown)


whitequark:
Drafted pending discussion in #1404.

whitequark:
The PR still needs a documentation update.

PR #1440, opened 2024-06-29
lib.data: accept negative int and slice for `ArrayLayout` indexing. [0.5 backport]

wanda-phi:
As a side effect, makes iterating over `ArrayLayout` views work.


whitequark (approved):
LGTM

PR #1441, opened 2024-06-29
build: add clock constraints on `IOPort` instead of buffered `Signal`.

wanda-phi:


amaranth/build/plat.py
```diff
@@ -328,7 +328,10 @@ def options(opts):
                 return " ".join(opts)
 
         def hierarchy(signal, separator):
-            return separator.join(self._name_map[signal][1:])
+            if isinstance(signal, IOPort):
```
whitequark, line 331: Perhaps this is better?

```suggestion
            if isinstance(net, IOPort):
```

whitequark (approved):
LGTM, thanks!

PR #1443, opened 2024-06-30
build, vendor: use `dir="-"` in `create_missing_domain`.

### LINKED ISSUE #1402: `create_missing_domain` uses `platform.request()` in deprecated mode in all platforms
This includes the default implementation.

It should use `request(..., dir="-")` and create an IO buffer explicitly.

wanda-phi:
Fixes #1402.


whitequark (approved):
Thanks!

PR #1448, opened 2024-07-19
Implement RFC 69

whitequark:


amaranth/lib/io.py
```diff
@@ -331,6 +314,167 @@ def __repr__(self):
...
+    @property
+    def invert(self):
+        return self._invert
+
+    @property
+    def direction(self):
+        return self._direction
+
+    def __len__(self):
+        if self._direction is Direction.Input:
```
wanda-phi, line 417: could just do `return len(self._invert)`

whitequark: That's... kind of cursed, though true.

amaranth/lib/io.py
```diff
@@ -476,6 +620,16 @@ def elaborate(self, platform):
             else:
                 m.submodules += IOBufferInstance(self._port.p, o=o_inv, oe=self.oe, i=i_inv)
                 m.submodules += IOBufferInstance(self._port.n, o=~o_inv, oe=self.oe)
+        elif isinstance(self._port, SimulationPort):
+            if self.direction in (Direction.Input, Direction.Bidir):
+                # Loop back `o` if `oe` is asserted. This frees the test harness from having to
+                # provide this functionality itself.
+                for i_inv_bit, oe_bit, o_bit, i_bit in \
+                        zip(i_inv, self._port.oe, self._port.o, self._port.i):
```
wanda-phi, line 628: this would cause an `AttributeError` on ports with `Input` direction, wouldn't it?

whitequark: Oh, hm. I wonder why are the tests not catching that?

whitequark: Fixed

docs/changes.rst
```diff
@@ -23,6 +23,14 @@ Version 0.6 (unreleased)
 ========================
 
 
+Implemented RFCs
```
wanda-phi, line 26: given that we intend to backport it, shouldn't this be in 0.5.1 section?

whitequark: moved

docs/changes.rst
```diff
@@ -59,6 +67,8 @@ Standard library changes
 * Added: views of :class:`amaranth.lib.data.ArrayLayout` can be indexed with negative integers or slices.
 * Added: :py:`len()` works on views of :class:`amaranth.lib.data.ArrayLayout`.
 * Added: views of :class:`amaranth.lib.data.ArrayLayout` are iterable.
+* Added: :class:`io.SimulationPort`. (`RFC 69`_)
+* Changed: providing :meth:`io.PortLike.__add__` is now mandatory. (`RFC 69`_)
```
wanda-phi, line 71: ... and that should actually be in 0.6?

docs/changes.rst
```diff
@@ -44,13 +44,22 @@ Standard library changes
 .. currentmodule:: amaranth.lib
 
 * Added: :py:`payload_init=` argument in :class:`amaranth.lib.stream.Signature`.
+* Changed: (deprecated in 0.5) providing :meth:`io.PortLike.__add__` is now mandatory. (`RFC 69`_)
```
wanda-phi, line 47:
```suggestion
* Changed: (deprecated in 0.5.1) providing :meth:`io.PortLike.__add__` is now mandatory. (`RFC 69`_)
```

whitequark: I've thought about this but then we should probably also edit the other version references to say .0 at the end...

whitequark: Done that.

tests/utils.py
```diff
@@ -16,15 +16,57 @@
...
+                        output_repr.append(prefix)
+                    else:
+                        output_repr.append(char)
+                elif name is None:
+                    current += char
+                    output_repr.append(char)
+                else:
+                    output_repr.append(char)
+            return "".join(output_repr)
+        # print("\n" + format_repr(squish_repr(repr(obj))))
```
wanda-phi, line 68: leftover?

tests/utils.py
```diff
@@ -16,15 +16,57 @@
...
         if isinstance(obj, list):
             obj = Statement.cast(obj)
-        def prepare_repr(repr_str):
+        def squish_repr(repr_str):
             repr_str = re.sub(r"\s+",   " ",  repr_str)
             repr_str = re.sub(r"\( (?=\()", "(", repr_str)
             repr_str = re.sub(r"\) (?=\))", ")", repr_str)
             return repr_str.strip()
-        self.assertEqual(prepare_repr(repr(obj)), prepare_repr(repr_str))
+        def format_repr(input_repr, *, indent="    "):
```
wanda-phi, line 29: horrifying.

tests/utils.py
whitequark: No, intentionally left in there for the next person who needs to update NIR tests. You can look at the diff but the diff has annoying `- ` at the front of each line.

tests/utils.py
whitequark: thank you

tests/utils.py
wanda-phi: fair enough

wanda-phi (approved):

PR #1449, opened 2024-07-19
hdl: truncate init value read from `Signal(...).init`

whitequark:
We have a warning saying that the init value will be truncated to fit the signal's shape (suppressed for common patterns of 0 and -1), but the introspected via `.init` value is, confusingly, not truncated.

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -2052,7 +2052,7 @@ class Signal(Value, DUID, metaclass=_SignalMeta):
                             .format(orig_init, shape),
                     category=SyntaxWarning,
                     stacklevel=2)
-        self._init = init.value
+        self._init = Const(init.value, shape).value
         self._reset_less = bool(reset_less)
 
         if isinstance(orig_shape, range) and orig_init is not None and orig_init not in orig_shape:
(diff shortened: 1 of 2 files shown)


whitequark:
I consider this PR a bugfix and tagged it for a backport to 0.5.1.

amaranth/hdl/_ast.py
```diff
@@ -2052,7 +2052,9 @@ def __init__(self, shape=None, *, name=None, init=None, reset=None, reset_less=F
                             .format(orig_init, shape),
                     category=SyntaxWarning,
                     stacklevel=2)
-        self._init = init.value
+        self._init = init.value & ((1 << self._width) - 1)
```
wanda-phi, line 2055: it may be a better idea to instead use `Const(..., shape).value`, and do it near the `init = Const.cast(init)` branch above?

whitequark: I'm not sure what you mean by the second part of this sentence. Can you write it as a diff?

amaranth/hdl/_ast.py
```diff
@@ -2052,7 +2052,9 @@ def __init__(self, shape=None, *, name=None, init=None, reset=None, reset_less=F
...
+        if self._signed and init.value < 0:
+            self._init |= -1 << self._width
```
wanda-phi, line 2057:
```suggestion
        init = Const(init.value, shape)
        self._init = init.value
```

whitequark: Applied, thanks!

wanda-phi (approved):

Issue #1454, opened 2024-07-20
`io.SimulationPort` invites undetected driver-driver conflicts that nondeterministically break semantics


whitequark:
Repro:

```python
from amaranth import *
from amaranth.lib import io
from amaranth.sim import Simulator


port = io.SimulationPort("io", 2)

m = Module()
m.submodules.p0 = p0 = io.FFBuffer("io", port[0])
m.submodules.p1 = p1 = io.FFBuffer("io", port[1])
m.d.comb += p0.o.eq(1)
m.d.comb += p1.o.eq(1)

async def testbench(ctx):
    await ctx.tick()
    print(f"{ctx.get(port.o):02b}")

sim = Simulator(m)
sim.add_clock(1e-6)
sim.add_testbench(testbench)
sim.run()
```

This should always output `11`, but outputs either `01` or `10`.

The cause is a driver-driver conflict that isn't detected by pysim (since nothing in pysim can do that). Running `verilog.convert(m, ports=[])` on it detects the conflict.

Although it looks like it would be a valid workaround (there is no driver-driver conflict both according to language semantics and according to `verilog.convert()`), this still fails with the same nondeterministic output:

```python
port_0 = io.SimulationPort("io", 1)
port_1 = io.SimulationPort("io", 1)
port = port_0 + port_1

m = Module()
m.submodules.p0 = p0 = io.FFBuffer("io", port[0])
m.submodules.p1 = p1 = io.FFBuffer("io", port[1])
m.d.comb += p0.o.eq(1)
m.d.comb += p1.o.eq(1)
```

This does produce the expected `11` output:

```python
port_0 = io.SimulationPort("io", 1)
port_1 = io.SimulationPort("io", 1)
port = port_0 + port_1

m = Module()
m.submodules.p0 = p0 = io.FFBuffer("io", port_0)
m.submodules.p1 = p1 = io.FFBuffer("io", port_1)
m.d.comb += p0.o.eq(1)
m.d.comb += p1.o.eq(1)
```

PR #1456, opened 2024-07-21
sim: make driving parts of a signal from distinct modules possible.

wanda-phi:
Fixes (part of) #1454.

### DIFF
--- a/amaranth/hdl/_xfrm.py
+++ b/amaranth/hdl/_xfrm.py
@@ -17,6 +17,7 @@ __all__ = ["ValueVisitor", "ValueTransformer",
            "FragmentTransformer",
            "TransformedElaboratable",
            "DomainCollector", "DomainRenamer", "DomainLowerer",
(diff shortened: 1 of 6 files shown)


amaranth/hdl/_xfrm.py
```diff
@@ -575,6 +575,71 @@ def on_fragment(self, fragment):
...
+        elif type(stmt) is Switch:
+            for (_, substmt, _) in stmt.cases:
+                self.visit_stmt(substmt)
+        elif type(stmt) in (Property, Print):
+            pass
+        elif isinstance(stmt, Iterable):
+            for substmt in stmt:
+                self.visit_stmt(substmt)
+        else:
+            assert False # nocov
```
whitequark, line 594: Needs to be `:nocov:` iirc?

```suggestion
            assert False # :nocov:
```

whitequark (approved):
LGTM

PR #1461, opened 2024-07-23
Implement RFC 66: Simulation time.

zyp:


whitequark (commented):
The implementation looks pretty good--can we have tests?

whitequark:
You should add a changelog entry.

Looks good to me on the first pass.

@wanda-phi, could you take a look as well?

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,249 @@
...
+        if not kwargs:
+            self._femtoseconds = 0
+            return
+
+        if len(kwargs) > 1:
+            raise TypeError("Period accepts at most one argument")
+
+        (unit, value), = kwargs.items()
+
+        m = re.match(r"^([munpf]?)s|([kMG]?)Hz$", unit)
```
wanda-phi, line 19: using a regex here is unnecessarily opaque, and also duplicates information already contained in the dicts below.

I'd do something like that:

```
TIME_UNITS = {
    "s": 1_000_000_000_000_000,
    "ms": 1_000_000_000_000,
    # ...
}

FREQUENCY_UNITS = {
    "Hz": 1_000_000_000_000_000,
    "kHz": 1_000_000_000_000,
    # ...
}

if unit in TIME_UNITS:
    self._femtoseconds = round(value * TIME_UNITS[unit])
elif unit in FREQUENCY_UNITS:
    # ...
    self._femtoseconds = round(FREQUENCY_UNITS[unit] / value)
else:
    raise TypeError(...)
```

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,249 @@
...
+                "f": 1,
+            }[s_unit])
+
+        else:
+            if value == 0:
+                raise ZeroDivisionError("Frequency can't be zero")
+            elif value < 0:
+                raise ValueError("Frequency can't be negative")
+
+            self._femtoseconds = round(1 / value * {
```
wanda-phi, line 45: this is an OCD-grade nit, but `(a / b)` has less rounding error than `1 / b * a`

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,249 @@
...
+        if not isinstance(other, Period):
+            return NotImplemented
+        return Period(fs = self._femtoseconds - other._femtoseconds)
+
+    def __mul__(self, other):
+        if not isinstance(other, numbers.Real):
+            return NotImplemented
+        return Period(fs = self._femtoseconds * other)
+
+    def __rmul__(self, other):
```
wanda-phi, line 162: this is subtly, but badly broken.

delegating via calling `*` will go into an infinite recursion when there are two different classes involved that have similar `NotImplemented` handling, because it will repeat the `__mul__`/`__rmul__` resolution dance. for an example, run this python program:

```py
class A:
    def __mul__(self, other):
        if not isinstance(other, A):
            return NotImplemented
        return 1

    def __rmul__(self, other):
        return self * other

class B:
    def __mul__(self, other):
        if not isinstance(other, B):
            return NotImplemented
        return 1

    def __rmul__(self, other):
        return self * other

print(A() * B())
```

what you should do instead is simply:

```
__rmul__ = __mul__
```

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,249 @@
...
+    def __mod__(self, other):
+        if not isinstance(other, Period):
+            return NotImplemented
+        return Period(fs = self._femtoseconds % other._femtoseconds)
+
+    def __str__(self):
+        return self.__format__("")
+
+    def __format__(self, format_spec):
+        m = re.match(r"^([1-9]\d*)?(.\d+)?( ?)(([munpf]?)s|([kMG]?)Hz)?$", format_spec)
```
wanda-phi, line 187: that dot in the middle of the regex needs escaping. some of the groups near the end look like they should be non-capturing groups.

also, this is horrifying. I can parse it, but aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,249 @@
...
+                value = f"{self.gigahertz:{precision or ''}f}"
+
+        str = f"{value}{space}{unit}"
+        if width:
+            str = f"{str:>{width}}"
+
+        return str
+
+    def __repr__(self):
+        return f"Period({self})"
```
wanda-phi, line 249: I'm wondering if this should be made into actually round-trippable syntax, but ... not sure about it

whitequark: `repr(x)` is supposed to be round-trippable, yes; I think that would be nice to have.

amaranth/hdl/_time.py
whitequark: I like Wanda's approach better.

amaranth/sim/_async.py
```diff
@@ -72,9 +73,18 @@ def __init__(self, interval):
...
         # `await ctx.delay(1e-15)` instead.
-        if interval < 0:
+
+        if not isinstance(interval, Period):
+            # TODO(amaranth-0.7): remove
+            warnings.warn(
+                f"Per RFC 66, the `interval` argument of `DelayTrigger()` will only accept a `Period`"
+                f" in the future.",
+                DeprecationWarning, stacklevel=1)
+            interval = Period(s = interval)
```
wanda-phi, line 83: the conventional formatting is `Period(s=interval)`

amaranth/sim/_async.py
```diff
@@ -72,9 +73,18 @@ def __init__(self, interval):
...
+
+        if interval < Period():
             raise ValueError(f"Delay cannot be negative")
-        self.interval_fs = round(float(interval) * 1e15)
+        self.interval_fs = interval.femtoseconds
```
wanda-phi, line 87: it might be better to just store the `Period` directly, now that we have that type

amaranth/sim/_async.py
```diff
@@ -681,6 +691,11 @@ def negedge(self, signal) -> TriggerCombination:
         """
         return TriggerCombination(self._engine, self._process).negedge(signal)
 
+    def elapsed_time(self) -> Period:
+        """Return the currently elapsed simulation time."""
+
+        return Period(fs = self._engine.now)
```
wanda-phi, line 697: `Period(fs=self._engine.now)`

amaranth/sim/core.py
```diff
@@ -122,12 +122,25 @@ def add_clock(self, period, *, phase=None, domain="sync", if_exists=False):
...
             raise DriverConflict(f"Domain {domain.name!r} already has a clock driving it")
 
-        period_fs = _seconds_to_femtos(period)
+        if not isinstance(period, Period):
+            # TODO(amaranth-0.7): remove
+            warnings.warn(
+                f"Per RFC 66, the `period` argument of `add_clock()` will only accept a `Period`"
+                f" in the future.",
+                DeprecationWarning, stacklevel=1)
+            period = Period(s = period)
```
wanda-phi, line 131: `Period(s=period)`

amaranth/sim/core.py
```diff
@@ -122,12 +122,25 @@ def add_clock(self, period, *, phase=None, domain="sync", if_exists=False):
...
-            phase_fs = _seconds_to_femtos(phase)
-        self._engine.add_clock_process(domain.clk, phase=phase_fs, period=period_fs)
+            phase = period / 2
+        elif not isinstance(phase, Period):
+            # TODO(amaranth-0.7): remove
+            warnings.warn(
+                f"Per RFC 66, the `phase` argument of `add_clock()` will only accept a `Period`"
+                f" (or `None`) in the future.",
+                DeprecationWarning, stacklevel=1)
+            phase = Period(s = phase)
```
wanda-phi, line 141: `Period(s=phase)`

amaranth/vendor/_siliconblue.py
```diff
@@ -382,21 +382,21 @@ def create_missing_domain(self, name):
...
+                delay = Period(us=100) // self.default_clk_period
             # Internal low-speed clock: 10 KHz.
             elif self.default_clk == "SB_LFOSC":
                 clk_i = Signal()
                 m.submodules += Instance("SB_LFOSC",
                                          i_CLKLFEN=1,
                                          i_CLKLFPU=1,
                                          o_CLKLF=clk_i)
-                delay = int(100e-6 * self.default_clk_frequency)
+                delay = Period(us=100) // self.default_clk_period
```
wanda-phi, line 393: while this is not a problem introduced by this PR, this is dangerously wrong: for safety reasons, this should be obviously rounded *up*, not down. for `SB_LFOSC` in particular, we are one ulp away from rounding this 0.9999 to a 0, instead of a 1.

... actually, I believe this code should be rounding up and also adding an extra margin of `1` cycle because there is no guarantee when the first rising edge actually comes up.

amaranth/hdl/_time.py
```diff
@@ -0,0 +1,252 @@
...
+            elif abs(self._femtoseconds) >= 1_000_000_000:
+                s_unit = "u"
+            elif abs(self._femtoseconds) >= 1_000_000:
+                s_unit = "n"
+            elif abs(self._femtoseconds) >= 1_000:
+                s_unit = "p"
+            else:
+                s_unit = "f"
+
+            unit = f"{s_unit}s"
```
whitequark, line 209: Don't we already have `_TIME_UNITS`?

whitequark: I see, `s_unit` is used below. This is probably good enough.

whitequark (approved):
LGTM

PR #1471, opened 2024-08-03
docs/stdlib/data: improved wording describing example that is shape-like

purdeaandrei:
The RGBLayout class itself is not shapelike, only instances of this class are shape-like. (i.e. `Shape.cast(RGBLayout)` would fail) It's worth being precise especially because in some cases the classes themselves are shape-like, for example in the case of Enums, the classes that inherit from Enum are shape-like.


whitequark (approved):
Thanks!

PR #1472, opened 2024-08-03
docs/stdlib/data: view value-likes can't be used _anywhere_ values can

purdeaandrei:
Clarify where View value-likes can/can't be used like plain values


purdeaandrei:
image

whitequark (commented):
Thanks for the PR! I'm going to give it a closer look when I'm better able to, since this is a more subtle topic.

PR #1475, opened 2024-08-04
docs/stdlib/fifo: document fifo stream interfaces

purdeaandrei:
Rendered document:
image

### DIFF
--- a/amaranth/lib/fifo.py
+++ b/amaranth/lib/fifo.py
@@ -60,6 +60,10 @@ class FIFOInterface:
         Does nothing if ``r_rdy`` is not asserted.
     r_level : Signal(range(depth + 1)), out
         Number of unread entries.
+    w_stream : :class:`amaranth.lib.wiring.FlippedInterface`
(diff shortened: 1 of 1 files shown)


whitequark:
These interfaces are meant to be transitionary, and the FIFOs themselves should be adapted to be stream-first. However, that would require some amount of design work and an RFC.

I realize we should probably document these things as they are in the interim, but I'm going to think for a while about *how* we document them.

purdeaandrei:
I understand, the renamed signals look not so idiomatic.

Just... my perspective as a new user was:
1) I read the Data streams page, including the Completed pipeline example where I saw SyncFIFOBuffered connected up as a stream
2) I got to the FIFO page, and I saw no sign of stream interface, so my first reaction was: something's not quite right, and I had to look in the source code to figure it out.

Are the names of w_stream/r_stream likely to change for the FIFOs?
Cause if what's gonna happen is that the old single wire non-stream signals will disappear, and the fifo will only support the stream interface, then it sounds more like we should be deprecating the single wires, and documenting w_stream/r_stream.

PR #1480, opened 2024-08-16
change gowin platform to use nextpnr-himbaechel

pepijndevos:
Over at Apicula we're migrating from nextpnr-gowin to nextpnr-himbaechel, and this updates Amaranth to use the new target.

The old target was based on nextpnr-generic and had longstanding problems such as long load times and occasional miscompiles. The new target is implemented as a microarchitecture in nextpnr-himbaechel and is significantly faster, robuster, and already has more features than the old target. Only timing-driven synthesis is still in the works, after which the new target will be pretty much a superset of features.

This PR is slightly breaking in that the end user needs to install the new nextpnr target, but should otherwise be largely compatible. The only change is that if you're passing custom Apicula-specific arguments they need to be changed from `--option value` to `--vopt option=value`.

I have confirmed that this successfully builds a blinky on Tang Nano 9K with `python -m amaranth_boards.tang_nano_9k`

### DIFF
--- a/amaranth/vendor/_gowin.py
+++ b/amaranth/vendor/_gowin.py
@@ -169,7 +169,7 @@ class GowinPlatform(TemplatedPlatform):
 
     Required tools:
         * ``yosys``
-        * ``nextpnr-gowin``
+        * ``nextpnr-himbaechel`` with Gowin uarch
         * ``gowin_pack``
 
     The environment is populated by running the script specified in the environment variable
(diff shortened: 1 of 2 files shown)


whitequark:
If you rebase this branch it should work again.

whitequark:
Also, please update the commit message so that it follows our conventions (starts with `vendor._gowin: ` in this case).

whitequark:
Please add a changelog (`docs/changes.rst`) entry since it's a breaking-ish change.

pepijndevos:
I took an educated guess at the correct changelog formatting

whitequark (approved):
LGTM, thanks a lot!

PR #1481, opened 2024-08-16
Bump sphinxcontrib-yowasp-wavedrom dependency

whitequark:
This is currently breaking builds because of an invalid escape sequence warning being elevated to error severity.


whitequark:
Oh, it looks like there are many more things like that. Did Sphinx change how warnings are treated?..

PR #1485, opened 2024-08-21
Bump sphinxcontrib-platformpicker dependency

### LINKED ISSUE #1484: docs: Install page: switching the platform selection shifts the page
Using one of these buttons shifts which part of the page the user is looking at quite substantially:

image

This is really disorienting. Ideally, the scroll would get "anchored" somehow, adjusting the scroll position so that the button being pressed is at the same position on screen before and after being pressed.

whitequark:
Fixes #1484.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ test = [
 ]
 docs = [
   "sphinx~=7.1",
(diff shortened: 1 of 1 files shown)


whitequark:
Not strictly speaking necessary as the constraint is `~=`, but it's a valuable accessibility improvement.

PR #1494, opened 2024-08-26
sim: fix memory writes with signed shapes.

wanda-phi:


whitequark (approved):
LGTM. Should we backport?

wanda-phi:
> LGTM. Should we backport?

definitely; was just waiting on review for 0.6 versions of these PRs first

PR #1499, opened 2024-08-29
build.res: allow `dir="-"` when requesting resources with subsignals.

wanda-phi:
This makes the only non-deprecated way to do I/O not be a pain when subsignals are involved.

(thanks @meithecatte for the report)

### DIFF
--- a/amaranth/build/res.py
+++ b/amaranth/build/res.py
@@ -152,7 +152,8 @@ class ResourceManager:
 
         def merge_options(subsignal, dir, xdr):
             if isinstance(subsignal.ios[0], Subsignal):
-                if dir is None:
+                orig_dir = dir
+                if dir is None or dir == "-":
                     dir = dict()
                 if xdr is None:
                     xdr = dict()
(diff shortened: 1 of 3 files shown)


whitequark (commented):
LGTM, but I propose a 0.5 backport as well as a small note in changelog. Yes, it's a deprecated API and all that, but it's important enough that we should highlight it.

wanda-phi:
> LGTM, but I propose a 0.5 backport as well as a small note in changelog. Yes, it's a deprecated API and all that, but it's important enough that we should highlight it.

added a changelog entry; 0.5 backport already done

I don't think it's a deprecated API though? I mean, unless we want to play google...

whitequark:
> I don't think it's a deprecated API though? I mean, unless we want to play google...

I mean, basically all of the functionality of `platform.request` is expressly recommended to be bypassed in the `lib.io` docs. That would make it deprecated in my view.

whitequark (approved):
LGTM, thanks!

PR #1500, opened 2024-08-29
build.res: allow `dir="-"` when requesting resources with subsignals. [0.5 backport]

wanda-phi:
This makes the only non-deprecated way to do I/O not be a pain when subsignals are involved.


whitequark (approved):
LGTM

PR #1507, opened 2024-09-07
docs: sim: fix typo async for

purdeaandrei:


whitequark (approved):
Thanks!

whitequark:
Actually, should we maybe change those snippets to be doctests? That comes with a syntax check for free.

PR #1529, opened 2024-10-07
rpc: add support for `wiring.Component`

rroohhh:
Do not infer the ports from the publicly accessible wires, but instead delegate finding the ports to the `rtlil.convert` function

### DIFF
--- a/amaranth/rpc.py
+++ b/amaranth/rpc.py
@@ -3,6 +3,8 @@ import json
 import argparse
 import importlib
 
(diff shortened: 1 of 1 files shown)


whitequark (approved):
LGTM. Have you been using the RPC functionality?

rroohhh:
Yeah I started using it for my current project, that requires a (system)verilog toplevel and also uses other preexisting non-amaranth modules.

PR #1530, opened 2024-10-09
vendor._gowin: Add 'add_preferences' override

jeanthom:
Implement 'add_preferences' override for CST files, allowing customizing physical placement of primitive.

This implementation follows what was done in `vendor._lattice`.


whitequark (approved):
LGTM

PR #1536, opened 2024-10-17
hdl._dsl: use `Enum` for fsm state signal

rroohhh:
This means the `enum_{base_type,value_*}` attributes will be emitted on rtlil export. 

(Also fixes the `src_loc_at`, which used to point at `contextlib`)


whitequark (approved):
LGTM

PR #1540, opened 2024-11-11
rpc: use rtlil instead of ilang

rroohhh:
The ilang alias to the rtlil frontend got removed in https://github.com/YosysHQ/yosys/pull/4704


whitequark (approved):
Thanks!

PR #1542, opened 2024-11-16
back/vendor: s/ilang/rtlil as a result of YosysHQ/yosys#4704

schnommus:
https://github.com/amaranth-lang/amaranth/pull/1540 did not catch all uses of this command that was deprecated upstream in https://github.com/YosysHQ/yosys/pull/4704


whitequark (approved):
Thanks!

PR #1552, opened 2025-01-18
Fix repr of Signature so it works for subclasses

robtaylor:
At the moment the `__repr__` for `Signature` doesn't work for derived classes of `Signature` (as used e.g. in Glasgow).

Simple fix here.

### DIFF
--- a/amaranth/lib/wiring.py
+++ b/amaranth/lib/wiring.py
@@ -1002,7 +1002,7 @@ class Signature(metaclass=SignatureMeta):
         return tuple()
 
     def __repr__(self):
-        if type(self) is Signature:
+        if isinstance(self, Signature):
             return f"Signature({dict(self.members.items())})"
         return super().__repr__()
 


amaranth/lib/wiring.py
```diff
@@ -1002,7 +1002,7 @@ def annotations(self, obj, /):
         return tuple()
 
     def __repr__(self):
-        if type(self) is Signature:
+        if isinstance(self, Signature):
             return f"Signature({dict(self.members.items())})"
```
whitequark, line 1006: This isn't a correct fix. There are two problems with it.

1. A repr is supposed to give information about the class name as well. Your fix would incorrectly indicate that the class name is `Signature` rather than whatever the subclass is.
2. Subclassing `Signature` is only really useful if you're going to add non-member attributes to it. But the default implementation only prints members. The check here serves as a nudge to implementers to make sure they really did cover all attributes.

whitequark (changes requested):
Apologies, it seems like I've never submitted a review.

robtaylor:
Makes sense :)

PR #1561, opened 2025-02-02
amaranth._nir: implement new `Match` cell.

wanda-phi:
Both `Matches` and `PriorityMatch` cells are obsoleted by the new cell.

This isn't really a functional change, as the validity rules on the `PriorityMatch` inputs imposed by RTLIL backend forced the netlist to conform to the same rules as the new composite cell.


whitequark (approved):
LGTM

whitequark (approved):
LGTM

PR #1575, opened 2025-04-16
backends: s/ilang/rtlil as a result of YosysHQ/yosys#4704

jeanthom:
Replace all occurrences of `read_ilang` with `read_rtlil`.

Backported from https://github.com/amaranth-lang/amaranth/commit/590cba1d6c00dffba7abb5f399680ac50e252adf into v0.4.x

PR #1576, opened 2025-04-16
Drop support for Python 3.8.

jeanthom:
Drop support for Python 3.8: documentation, pyproject.toml and GitHub Actions updates.

Required by https://github.com/amaranth-lang/amaranth/pull/1575


jeanthom:
It looks like there is some interdependency between this PR and #1575 due to the Yosys command name change.

whitequark:
Right. I guess we can combine these into one PR for that reason.

PR #1578, opened 2025-04-21
[WIP] Implement RFC 41: `lib.fixed`

vk2seb:
## Overview

- This is a staging ground for experimenting with the ergonomics of `lib.fixed` while the [associated RFC](https://github.com/amaranth-lang/rfcs/pull/41) is being worked on. It started as a fork of @zyp's early implementation of the RFC [here](https://github.com/amaranth-lang/amaranth/pull/1005), however a few things have changed since then.
  - Most importantly, this PR adds some tests, which makes it more obvious what the consequences of different design decisions will be.
  - Also, all operators required for real-world use are now implemented
- As of now, this PR adheres to the latest version of the RFC, with some minor changes (these will be discussed further in the RFC issue):
  - New methods on `fixed.Value` - `.saturate()` and `.clamp()` - these are commonly needed in my DSP codebase, but it may make sense to punt these new methods to a future RFC.
  - `.truncate()` is added as an alias for `.reshape()`, the only difference being that it verifies there was a reduction of `f_bits` requested.
  - The `numerator()` method is dropped, as I found a way to combine it with `as_value()` reliably.
- I have already integrated this implementation of `lib.fixed` in [this Tiliqua PR](https://github.com/apfaudio/tiliqua/pull/103) and tested it underneath my library of DSP cores, and will continue to use the learnings there in order to guide the RFC.
- It should be obvious that this PR needs a cleanup pass, improved diagnostics, and a lot of documentation work after the RFC is complete. However, it is already usable in real projects as all essential operations are implemented.

## Simple example

Consider the implementation of a simple [Low-pass filter](https://en.wikipedia.org/wiki/Low-pass_filter), where we wish to compute the difference equation `y = y[n-1] * beta + x * (1 - beta)` using fixed point representation:

```python
class OnePole(wiring.Component):

    def __init__(self, beta=0.9, sq=fixed.SQ(1, 15)):
        self.beta = beta
        self.sq = sq
        super().__init__({
            "x": In(sq),
            "y": Out(sq),
        })

    def elaborate():
        m = Module()
        a = fixed.Const(self.beta, shape=self.sq)
        b = fixed.Const(1-self.beta, shape=self.sq)
        # Quantization from wider to smaller fixed.Value occurs on the `.eq()`
        m.d.sync += self.y.eq(self.y*a + self.x*b)
        return m
```

### DIFF
--- a/amaranth/lib/fixed.py
+++ b/amaranth/lib/fixed.py
@@ -215,7 +215,7 @@ class Value(hdl.ValueCastable):
                 return Value.cast(self.as_value(), self.f_bits - other)
         elif not isinstance(other, hdl.Value):
             raise TypeError("Shift amount must be an integer value")
-        if other.signed:
+        if other.shape().signed:
             raise TypeError("Shift amount must be unsigned")
         return Value.cast(self.as_value() << other, self.f_bits)
 
--- a/amaranth/lib/fixed.py
+++ b/amaranth/lib/fixed.py
@@ -282,8 +282,8 @@ class Const(Value):
         if shape is None:
             signed = num < 0
             f_bits = bits_for(den) - 1
-            i_bits = max(0, bits_for(abs(num)) - f_bits)
-            shape = SQ(i_bits+1, f_bits) if signed else UQ(i_bits, f_bits)
+            i_bits = max(1 if signed else 0, bits_for(num) - f_bits)
+            shape = SQ(i_bits, f_bits) if signed else UQ(i_bits, f_bits)
 
         # Scale value to given precision.
         if 2**shape.f_bits > den:
--- a/amaranth/lib/fixed.py
+++ b/amaranth/lib/fixed.py
@@ -1,0 +1,328 @@
+# Based on latest iteration of fixed point types RFC, which
+# is an effort undertaken by the Amaranth community, as well
+# as an early (incomplete) RFC implementation by zyp@
+#
+# RFC (community): https://github.com/amaranth-lang/rfcs/pull/41
+# Early implementation (zyp@): https://github.com/amaranth-lang/amaranth/pull/1005
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+from .. import hdl, Mux
+from ..utils import bits_for
+
+__all__ = ["Shape", "SQ", "UQ", "Value", "Const"]
+
+class Shape(hdl.ShapeCastable):
+
+    def __init__(self, shape, f_bits=0):
+        self._storage_shape = shape
+        self.i_bits, self.f_bits = shape.width-f_bits, f_bits
+        if self.i_bits < 0 or self.f_bits < 0:
+            raise TypeError(f"fixed.Shape may not be created with negative bit widths (i_bits={self.i_bits}, f_bits={self.f_bits})")
+        if shape.signed and self.i_bits == 0:
+           raise TypeError(f"A signed fixed.Shape cannot be created with i_bits=0")
+        if self.i_bits + self.f_bits == 0:
+            raise TypeError(f"fixed.Shape may not be created with zero width")
+
+    @property
+    def signed(self):
+        return self._storage_shape.signed
+
+    @staticmethod
+    def cast(shape, f_bits=0):
+        if not isinstance(shape, hdl.Shape):
+            raise TypeError(f"Object {shape!r} cannot be converted to a fixed.Shape")
+        return Shape(shape, f_bits)
+
+    def const(self, value):
+        if value is None:
+            value = 0
+        return Const(value, self)._target
+
+    def as_shape(self):
+        return self._storage_shape
+
+    def __call__(self, target):
+        return Value(self, target)
+
+    def min(self):
+        c = Const(0, self)
+        c._value = c._min_value()
+        return c
+
+    def max(self):
+        c = Const(0, self)
+        c._value = c._max_value()
+        return c
+
+    def from_bits(self, raw):
+        c = Const(0, self)
+        c._value = raw
+        if self.signed and raw > c._max_value():
+            # 2s complement signed value, but `raw` was unsigned.
+            c._value = c._min_value() + c._value - c._max_value() - 1
+        if c._value < c._min_value() or c._value > c._max_value():
+            raise ValueError(
+                f"{raw} outside expected range {c._min_value()}, {c._max_value()}")
+        return c
+
+    def __repr__(self):
+        return f"fixed.Shape({self._storage_shape}, f_bits={self.f_bits})"
+
+
+class SQ(Shape):
+    def __init__(self, i_bits, f_bits):
+        super().__init__(hdl.Shape(i_bits + f_bits, signed=True), f_bits)
+
+
+class UQ(Shape):
+    def __init__(self, i_bits, f_bits):
+        super().__init__(hdl.Shape(i_bits + f_bits, signed=False), f_bits)
+
+
+class Value(hdl.ValueCastable):
+    def __init__(self, shape, target):
+        self._shape = shape
+        if self.signed and not target.shape().signed:
+            # When methods bit-pick or concatenate to
+            # the _target of a Value, and then use this
+            # to reconstruct a Value, we may lose the
+            # signedness of its underlying _target.
+            self._target = target.as_signed()
+        else:
+            self._target = target
+
+    @property
+    def signed(self):
+        return self._shape.signed
+
+    @staticmethod
+    def cast(value, f_bits=0):
+        return Shape.cast(value.shape(), f_bits)(value)
+
+    @property
+    def i_bits(self):
+        return self._shape.i_bits
+
+    @property
+    def f_bits(self):
+        return self._shape.f_bits
+
+    def shape(self):
+        return self._shape
+
+    def as_value(self):
+        return self._target
+
+    def eq(self, other):
+        if isinstance(other, hdl.Value):
+            return self.as_value().eq(other)
+        elif isinstance(other, int) or isinstance(other, float):
+            other = Const(other, self.shape())
+        elif not isinstance(other, Value):
+            raise TypeError(f"Object {other!r} cannot be converted to a fixed.Value")
+        other = other.reshape(self.f_bits)
+        return self.as_value().eq(other.as_value())
+
+    def reshape(self, f_bits):
+        # If we're increasing precision, extend with more fractional bits. If we're
+        # reducing precision, truncate bits.
+        shape = hdl.Shape(self.i_bits + f_bits, signed=self.signed)
+        if f_bits > self.f_bits:
+            result = Shape(shape, f_bits)(hdl.Cat(hdl.Const(0, f_bits - self.f_bits), self.as_value()))
+        else:
+            result = Shape(shape, f_bits)(self.as_value()[self.f_bits - f_bits:])
+        return result
+
+    def truncate(self, f_bits=0):
+        if f_bits > self.f_bits:
+            raise ValueError(
+                f"`.truncate(f_bits={f_bits}) exceeds the underlying type's f_bits={self.f_bits}. "
+                "Use `.reshape()` to instead extend `f_bits`."
+            )
+        return self.reshape(f_bits)
+
+    def clamp(self, lo, hi):
+        if not isinstance(lo, Value) or not isinstance(hi, Value):
+            raise TypeError(f"Cannot `clamp` as lo, hi are not fixed.Value")
+        lo = lo.reshape(self.f_bits)
+        hi = hi.reshape(self.f_bits)
+        return Value(self.shape(), Mux(
+            self > hi, hi,
+            Mux(self < lo, lo, self)
+        ))
+
+    def saturate(self, shape):
+        if not isinstance(shape, Shape):
+            raise TypeError(f"Cannot `saturate` to bounds of {shape!r} as it is not a fixed.Shape")
+        if not shape.i_bits <= self.i_bits:
+            raise ValueError(f"Cannot `saturate`: shape.i_bits={shape.i_bits} > self.i_bits={self.i_bits} would have no effect.")
+        clamped = self.reshape(shape.f_bits).clamp(shape.min(), shape.max())
+        return Value(shape, clamped.as_value())
+
+    def _binary_op(self, rhs, operator, callable_f_bits = lambda a, b: max(a, b), pre_reshape=True, post_cast=True):
+        if isinstance(rhs, hdl.Value):
+            rhs = Value.cast(rhs)
+        elif isinstance(rhs, int):
+            rhs = Const(rhs)
+        elif not isinstance(rhs, Value):
+            raise TypeError(f"Object {rhs!r} cannot be converted to a fixed.Value")
+        f_bits = callable_f_bits(self.f_bits, rhs.f_bits)
+        if pre_reshape:
+            lhs = self.reshape(f_bits)
+            rhs = rhs.reshape(f_bits)
+        else:
+            lhs = self
+        value = getattr(lhs.as_value(), operator)(rhs.as_value())
+        return Value.cast(value, f_bits) if post_cast else value
+
+    def __mul__(self, other):
+        return self._binary_op(other, '__mul__', lambda a, b: a + b, pre_reshape=False)
+
+    __rmul__ = __mul__
+
+    def __add__(self, other):
+        return self._binary_op(other, '__add__')
+
+    __radd__ = __add__
+
+    def __sub__(self, other):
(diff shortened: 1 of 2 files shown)


goekce:
I noticed `==` seems to work differently than standard types. Is this intended? I noticed this while writing assertions:
```py
from amaranth import *
from amaranth.lib import fixed
from amaranth.lib.wiring import Component, In, Out


class C(Component):
    o: Out(1)

        m = Module()
        self.s = Signal(2)
        self.f = Signal(fixed.SQ(2, 2))
        m.d.sync += self.o.eq(self.s)
        return m


from amaranth.sim import Period, Simulator

dut = C()


async def bench(ctx):
    for _ in range(1):
        print(ctx.get(dut.s) == 0)  # True
        print(ctx.get(dut.f) == 0)  # (== (s (slice (const 4'sd0) 0:4)) (cat (const 2'd0) (const 1'd0)))
        await ctx.tick()


sim = Simulator(dut)
sim.add_clock(Period())
sim.add_testbench(bench)
sim.run()
```
outputs:
```
True
(== (s (slice (const 4'sd0) 0:4)) (cat (const 2'd0) (const 1'd0)))
```

vk2seb:
@goekce Thanks for taking a look! I think this is happening because of the comparison outside the simulation context, rather than inside it, which means that `from_bits` is converting the type back up to a `fixed.SQ`. I see similar behaviour with other types from `amaranth.lib`, for example `enum`:

```
from amaranth import *
from amaranth.lib import fixed, enum
from amaranth.lib.wiring import Component, In, Out

class Funct4(enum.Enum, shape=unsigned(4)):
    ADD = 0
    SUB = 1
    MUL = 2

class C(Component):
    o: Out(1)

        m = Module()
        self.s = Signal(2)
        self.e = Signal(Funct4)
        self.f = Signal(fixed.SQ(2, 2))
        m.d.sync += self.o.eq(self.s)
        return m


from amaranth.sim import Period, Simulator

dut = C()


async def bench(ctx):
    for _ in range(1):
        # raw result of from_bits()
        print(ctx.get(dut.s)) # 0
        print(ctx.get(dut.e)) # Funct4.ADD
        print(ctx.get(dut.f)) # fixed.SQ(2, 2) (const 4'sd0)
        # compare inside the simulation context
        print(ctx.get(dut.s == 0))                                    # 1
        print(ctx.get(dut.e == Funct4.ADD))                           # 1
        print(ctx.get(dut.f == 0))                                    # 1
        print(ctx.get(dut.f == fixed.Const(0, shape=fixed.SQ(2, 2)))) # 1
        # compare outside the simulation context
        print(ctx.get(dut.e) == 0) # False (even though Funct4.ADD is 0)
        print(ctx.get(dut.f) == 0) # (== (s (slice (const 4'sd0) 0:4)) (cat (const 2'd0) (const 1'd0)))
        await ctx.tick()


sim = Simulator(dut)
sim.add_clock(Period())
sim.add_testbench(bench)
sim.run()
```

As an example for writing assertions, maybe take a look at the tests attached to this PR. That being said, this is still a work in progress and not quite ready for review yet!

vk2seb:
@goekce a related topic to your example's evaluation outside the simulation context is elaboration-time evaluation of constant expressions. which is something I'd like to leave out of this RFC, even if we could add it in the future. For example

```
>>> from amaranth import *
>>> Const(5) + Const(3)
(+ (const 3'd5) (const 2'd3))
>>> Const(5) == Const(3)
(== (const 3'd5) (const 2'd3))
```

goekce:
I was not aware of the possibility that I can do a simulation-time comparison. Thanks Seb, this would solve my problem.

As I understand, elaboration-time evaluation of constant expressions is extra work.👍

whitequark:
You can already do elaboration-time comparison of const expressions by doing `Const.cast(a).value == Const.cast(b).value`.

vk2seb:
For comparison with zero that makes sense but for other values, if we drop the `fixed.Shape` this kind of comparison is not so intuitive
```
>>> Const.cast(fixed.Const(0.5, shape=fixed.SQ(2, 2))).value == 2
True
```
I think this kind of use case is better covered by `as_float`
```
fixed.Const(0.5, shape=fixed.SQ(2, 2)).as_float() == 0.5
```
In the example from @goekce - if we really want to do the comparison outside the sim context for some reason
```
print(ctx.get(dut.f).as_float() == 0.0) # True
```

whitequark:
What I mean is that `Const.cast()` is the Amaranth entry point for e.g. resolving concantenations or indexing into an Amaranth constant value. It is up to the end user or library how this API is used, I only wanted to mention that it is available.

goekce:
I am confused about how `Const.cast()` would be valuable here.

I thought `Const.cast()` is actually used in the following expression so that the signal provides directly an output that can be used in elaboration-time:

```
(Pdb) ctx.get(dut.s)
0
```

But it is probably not used:
```
(Pdb) Const.cast(dut.s)
*** TypeError: Value (sig s) cannot be converted to an Amaranth constant
```

Or do you just want to say that `Const.cast()` could be used for `fixed` values whenever an expression like `with m.Case(...):` as documented in [the RFC about `Const.cast()`](https://github.com/amaranth-lang/rfcs/blob/main/text/0004-const-castable-exprs.md)?

goekce:
I noticed `__str__` should be included in RFC. Using `__str__` from the parent class could lead to confusion. In the following `fixed` signal is printed as `-4`:

```py
from amaranth import *
from amaranth.lib import fixed
from amaranth.lib.wiring import Component, In, Out


class C(Component):
    o: Out(1)

        m = Module()
        self.s = Signal(signed(2), init=-1)
        self.f = Signal(fixed.SQ(2, 2), init=-1.0)
        m.d.sync += Print('Print signed', self.s)
        m.d.sync += Print('Print fixed', self.f)
        return m


from amaranth.sim import Period, Simulator

dut = C()


async def bench(ctx):
    print('ctx.get signed', ctx.get(dut.s))
    print('ctx.get fixed', ctx.get(dut.f))
    await ctx.tick()


sim = Simulator(dut)
sim.add_clock(Period())
sim.add_testbench(bench)
sim.run()
```
Output:
```
ctx.get signed -1
ctx.get fixed fixed.SQ(2, 2) (const 4'sd-4)
Print signed -1
Print fixed -4
```

amaranth/lib/fixed.py
```diff
@@ -0,0 +1,328 @@
...
+        return self.as_value().eq(other.as_value())
+
+    def reshape(self, f_bits):
+        # If we're increasing precision, extend with more fractional bits. If we're
+        # reducing precision, truncate bits.
+        shape = hdl.Shape(self.i_bits + f_bits, signed=self.signed)
+        if f_bits > self.f_bits:
+            result = Shape(shape, f_bits)(hdl.Cat(hdl.Const(0, f_bits - self.f_bits), self.as_value()))
+        else:
+            result = Shape(shape, f_bits)(self.as_value()[self.f_bits - f_bits:])
```
goekce, line 134:
```suggestion
        elif f_bits < self.f_bits:
            result = Shape(shape, f_bits)(self.as_value()[self.f_bits - f_bits:])
        else:
            result = Shape(shape, f_bits)(self.as_value())
```
This suggestion avoids unnecessary slicing, which simplifies the representation:
- before: `fixed.SQ(3, 3) (+ (s (slice (const 5'sd-9) 0:5)) (cat (const 3'd0) (const 1'd1)))`
- after: `fixed.SQ(3, 3) (+ (const 5'sd-9) (cat (const 3'd0) (const 1'd1)))`

goekce:
If a `Mux` is an operand, the arithmetic result is wrong. I did not have the time to investigate further, but have the feeling that `Mux` cannot return that the result should be interpreted as a `fixed` shape:

`self.a` should output `0.25`, but it does not:

```py
from amaranth import *
from amaranth.lib import fixed
from amaranth.lib.wiring import Component, In, Out


class C(Component):
    y: Out(fixed.SQ(8, 4))

        m = Module()
        self.a = Signal(self.y.shape())
        self.b = Signal(self.y.shape())
        self.c = Signal(self.y.shape())
        m.d.comb += [
            self.a.eq(fixed.Const(-1.125) + Mux(1, fixed.Const(1.375), 0)),
            self.b.eq(fixed.Const(-1.125) + fixed.Const(1.375)),
            self.c.eq(Mux(1, fixed.Const(1.375), 0)),
        ]
        m.d.sync += self.y.eq(self.a)
        return m


from amaranth.sim import Period, Simulator

dut = C()


async def bench(ctx):
    await ctx.tick()
    print(ctx.get(dut.a).as_float())
    print(ctx.get(dut.b).as_float())
    print(ctx.get(dut.c).as_float())
    print(Mux(1, fixed.Const(1.375), 0).shape())


sim = Simulator(dut)
sim.add_clock(Period())
sim.add_testbench(bench)
sim.run()
```
Result:
```
9.875
0.25
0.6875
unsigned(4)
```

amaranth/lib/fixed.py
vk2seb: agree, thanks!

vk2seb:
> If a Mux is an operand, the arithmetic result is wrong. I did not have the time to investigate further, but have the feeling that Mux cannot return that the result should be interpreted as a fixed shape:

Thanks, I haven't played with `Mux` much in combination with `fixed.Value`, this makes a good candidate for some more test cases. There are some tests on the `fixed.Value(shape, Mux(a, b, c))` statement inside `fixed.Value.clamp(...)`, but in this case the 2 shapes are guaranteed to be the same, so I guess `Mux` losing the underlying type did not have an adverse effect and so I didn't catch it. In your example, forcing the incoming type to match also 'fixes' it: `self.c.eq(Mux(1, fixed.Const(1.375, self.y.shape()), 0))`

Ideally we want to attack this without touching anything outside `lib.fixed`, on a quick skim I think the information is lost on the cast to `Value` [here](https://github.com/amaranth-lang/amaranth/blob/fd41201e20592f412b15fd33fd29a9eb36899c69/amaranth/hdl/_ast.py#L1908). This also looks related to [`Shape._unify`](https://github.com/amaranth-lang/amaranth/blob/fd41201e20592f412b15fd33fd29a9eb36899c69/amaranth/hdl/_ast.py#L175) - we kind of need a similar `_unify` operation to automatically perform a `.reshape()` up to `max(f_bits)` in the same fashion as `fixed.Value._binary_op` already does.

It seems other types similarly lose the type information through `Mux`:

```
from amaranth.lib import enum, data

class Funct(enum.Enum, shape=4):
    ADD = 0
    SUB = 1
    MUL = 2

rgb565_layout = data.StructLayout({
    "red":   5,
    "green": 6,
    "blue":  5
})

print(Mux(1, Funct.ADD, Funct.MUL).shape()) # unsigned(4)
print(Mux(1, Signal(rgb565_layout), Signal(rgb565_layout)).shape()) # unsigned(16)
```

In this case it seems making `Mux` more intelligent to what is happening to `fixed.Value` would be different to how existing types behave. What do you think @whitequark ? I would like to avoid touching the infrastructure underneath `Mux` if we can here, and then we could address preserving shapes through `Mux` in a future RFC?

whitequark:
> I would like to avoid touching the infrastructure underneath `Mux` if we can here, and then we could address preserving shapes through `Mux` in a future RFC?

Yeah, that sounds reasonable. We actually discussed the option of preserving shapes through `Mux` and `SwitchValue`; @wanda-phi, do you recall what came out of those discussions?

vk2seb:
> The workaround I found was to use an m.If and create two different arithmetic assignments.

If one really wants to use a `Mux` that is `lib.Fixed` aware, an easier workaround might be to build this on top of the primitives already supplied by `lib.Fixed`, for example:

```
def FMux(test, a, b):
    if isinstance(a, fixed.Value) and isinstance(b, fixed.Value):
        f_bits = max(a.f_bits, b.f_bits)
        return fixed.Value(a.shape() if a.i_bits >= b.i_bits else b.shape(),
                     Mux(test, a.reshape(f_bits), b.reshape(f_bits)))
    elif isinstance(a, fixed.Value):
        return fixed.Value(a.shape(), Mux(test, a, b))
    elif isinstance(b, fixed.Value):
        return fixed.Value(b.shape(), Mux(test, a, b))
    else:
        raise TypeError("FMux should only be used on fixed.Value")
```

On your example:

```
self.a.eq(fixed.Const(-1.125) + FMux(1, fixed.Const(1.375), 0)),
self.b.eq(fixed.Const(-1.125) + fixed.Const(1.375)),
self.c.eq(FMux(1, fixed.Const(1.375), 0)),
self.d.eq(fixed.Const(-1.125) + FMux(1, fixed.Const(1.375, self.y.shape()), 0)),
self.e.eq(fixed.Const(-1.125) + FMux(1, fixed.Const(1.375, self.y.shape()), 0)),
self.f.eq(fixed.Const(-1.125) + FMux(1, fixed.Const(1.375), fixed.Const(0))),

# Prints
fixed.Const(0.25, SQ(8, 4))
fixed.Const(0.25, SQ(8, 4))
fixed.Const(1.375, SQ(8, 4))
fixed.Const(0.25, SQ(8, 4))
fixed.Const(0.25, SQ(8, 4))
fixed.Const(0.25, SQ(8, 4))
```

I would however not include such a workaround in this RFC. I think attacking this properly would imply modifying `Mux` to preserve shapes. Which should be a separate RFC, as it has wider consequences than just adding a module to the standard library as `lib.fixed` aims to do.

Whether `lib.fixed` should be blocked by a shape-preserving `Mux` implementation is not something I have a strong opinion on, at least in my DSP codebases I haven't found much use for `Mux` outside of `.clamp()` which is provided here and is already shape-preserving.

goekce:
Thanks for the `FMux`. I will try it.

IMHO `fixed` could attract other users, so I would not block it, if shape preserving `Mux` has not a high priority right now. `Mux` issue should then be noted though.

goekce:
1. I tried `FMux` and it works 🙂. I find it a short way of expressing statements like `z = a + (b if c else d) + e` instead of writing `if ... then z = a + b + e else z = a + d + e`.

3. I noticed:
    ```py
   In [30]: fixed.Const(-2)
   Out[30]: fixed.Const(-2.0, SQ(3, 0))
   ```
   -2 requires only two bits in two's complement. Is it intended that it generates 3 bits?

amaranth/lib/fixed.py
```diff
@@ -0,0 +1,334 @@
...
+        # If we're increasing precision, extend with more fractional bits. If we're
+        # reducing precision, truncate bits.
+        shape = hdl.Shape(self.i_bits + f_bits, signed=self.signed)
+        if f_bits > self.f_bits:
+            result = Shape(shape, f_bits)(hdl.Cat(hdl.Const(0, f_bits - self.f_bits), self.as_value()))
+        elif f_bits < self.f_bits:
+            result = Shape(shape, f_bits)(self.as_value()[self.f_bits - f_bits:])
+        else:
+            result = Shape(shape, f_bits)(self.as_value())
+        return result
```
goekce, line 137: I additionally noticed that [`reshape(shape)` from the RFC](https://github.com/zyp/amaranth-rfcs/blob/cb3d938753a6c22640bdb0befefd236300be9937/text/0041-fixed-point.md?plain=1#L63-L64) is not implemented. Is this intended?

amaranth/lib/fixed.py
```diff
@@ -0,0 +1,334 @@
...
+                raise ValueError("Shift amount cannot be negative")
+
+            if other > self.f_bits:
+                value = hdl.Cat(hdl.Const(0, other - self.f_bits), self.as_value())
+                return Value.cast(value.as_signed() if self.signed else value)
+            else:
+                return Value.cast(self.as_value(), self.f_bits - other)
+        elif not isinstance(other, hdl.Value):
+            raise TypeError("Shift amount must be an integer value")
+        if other.signed:
```
qbojj: probably should be `other.shape().signed`

amaranth/lib/fixed.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def __gt__(self, other):
+        return self._binary_compare(other, '__gt__')
+
+    def __le__(self, other):
+        return self._binary_compare(other, '__le__')
+
+    def __eq__(self, other):
+        return self._binary_compare(other, '__eq__')
+
```
qbojj: missing `__ne__`

vk2seb: thanks, added this and a simple test for it.

turbocool3r:
I've refactored my radio pet project to use this PR and it looks really good now, thanks!

The only thing I've found confusing is that `Const(a)` will raise an exception in case `a` doesn't fit in it's shape, but won't even warn you if the shape doesn't have enough fractional bits to represent `a` and it becomes zero.

Also what about complex numbers? That'll be very useful for RF stuff. Should these be at least considered in this PR?

PR #1601, opened 2025-05-14
tracer: fix on pypy-3.11

rroohhh:
pypy does not have _varname_from_oparg, so fall back to pre cpython-3.11 path 

ref: https://github.com/pypy/pypy/blob/0253c85bf5f86a45760eda7f0af8953ed8a253c7/lib-python/3/dis.py#L351

### DIFF
--- a/amaranth/tracer.py
+++ b/amaranth/tracer.py
@@ -1,4 +1,5 @@
 import sys
+import platform
 from opcode import opname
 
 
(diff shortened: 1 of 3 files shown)


amaranth/tracer.py
```diff
@@ -44,13 +45,13 @@ def get_var_name(depth=2, default=_raise_exception):
             return code.co_names[imm]
         elif opc == "STORE_FAST":
             imm |= int(code.co_code[index + 1])
-            if sys.version_info >= (3, 11):
+            if sys.version_info >= (3, 11) and platform.python_implementation() != 'PyPy':
```
whitequark, line 48: Since we are using a private method, should we check for CPython (include) rather than PyPy (exclude)?

rroohhh: Yeah makes sense.

rroohhh:
So now this hits https://github.com/pypy/pypy/issues/3183 in sby, so blocked on https://github.com/YosysHQ/sby/pull/323 I guess.

whitequark:
Let's add a `@unittest.skipIf` to those sby tests; it's not like missing them on a single PyPy version is going to impact anything.

whitequark (approved):

whitequark:
Thanks!

PR #1606, opened 2025-06-15
Update document for `ResetSynchronizer`

### LINKED ISSUE #1605: Outdated document on get_reset_sync
There isn't a direct platform override for this anymore after the change to forward it to `AsyncFFSynchronizer`.

yuyichao:
Remove reference of out-dated platform method and point to `AsyncFFSynchronizer` instead.

Fix #1605

I kept this in the Platform override section since it seems that this info is only relevant for this part.

### DIFF
--- a/amaranth/lib/cdc.py
+++ b/amaranth/lib/cdc.py
@@ -220,8 +220,8 @@ class ResetSynchronizer(Elaboratable):
 
     Platform overrides
     ------------------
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks. It looks like an upstream change broke our CI, I'll investigate when I have a moment.

whitequark:
Could you rebase please?

Issue #1382, opened 2024-06-14
`AttributeError` caused by `UnusedElaboratable` after an elaboratable has failed its initialization


jfng:
Repro:

```python3
from amaranth import *
from amaranth.lib import wiring
from amaranth.hdl import Fragment


class Foo(wiring.Component):
    def __init__(self, arg):
        if not isinstance(arg, int):
            raise TypeError
        self._arg = arg
        super().__init__({})

    def elaborate(self, platform):
        return Module()

    def __repr__(self):
        return f"Foo({self._arg!r})"


Fragment.get(Foo(1), None)
try:
    Foo("fail")
except TypeError:
    pass
```

Output:

```
# Exception ignored in: <function MustUse.__del__ at 0x7f8410a2d440>
# Traceback (most recent call last):
#   File ".../python3.12/site-packages/amaranth/_unused.py", line 37, in __del__
#     f"{self!r} created but never used", self._MustUse__warning,
#       ^^^^^^^^
#   File "/tmp/repro.py", line 17, in __repr__
#     return f"Foo({self._arg!r})"
#                   ^^^^^^^^^
# AttributeError: 'Foo' object has no attribute '_arg'
```

jfng:
This scenario is likely to only happen in unit tests, which can avoid it with a `# amaranth: UnusedElaboratable=no` modeline.

Issue #924, opened 2023-10-04
`UnusedElaboratable` message still appearing in `pytest` tests when `xfail(run=False)` marker is used in conjunction with Simulator and warning is suppressed.


cr1901:
To duplicate:

1. `git clone https://github.com/cr1901/amaranth-unused-elaboratable`
2. `pdm install`
3. `pdm run test`

Output on my Windows machine is the following:

```
William@DESKTOP-3H1DSBV MINGW64 ~/Projects/MCVE/amaranth-unused
$ pdm run test
======================================================== test session starts ========================================================= platform win32 -- Python 3.11.5, pytest-7.4.1, pluggy-1.3.0
rootdir: C:/msys64/home/William/Projects/MCVE/amaranth-unused
configfile: pyproject.toml
collected 2 items

tests/test_bar.py x                                                                                                             [ 50%] tests/test_foo.py .                                                                                                             [100%]

==================================================== 1 passed, 1 xfailed in 0.19s ==================================================== C:/msys64/home/William/Projects/MCVE/amaranth-unused/src/amaranth_template_fpga/bar.py:8: UnusedElaboratable: <amaranth_template_fpga.f

oo.Foo object at 0x000002af31f85310> created but never used
```

Test files have magic comment to suppress that warning:

```
William@DESKTOP-3H1DSBV MINGW64 ~/Projects/MCVE/amaranth-unused
$ grep Unused -R src/ tests/
tests/conftest.py:# amaranth: UnusedElaboratable=no
tests/test_bar.py:# amaranth: UnusedElaboratable=no
tests/test_bar.py:@pytest.mark.xfail(run=False)  # Comment to quash UnusedElaboratable warning.
tests/test_foo.py:# amaranth: UnusedElaboratable=no
```

Confirmed to also have the same message on Ubuntu 20.04 Linux w/ Python 3.8 as well. Commenting out line 9 of _either_ test file causes the warning to disappear.

Issue #848, opened 2023-07-23
Don't show UnusedElaboratable warning if not a single elaboratable was used


whitequark:
In this case it's probably just an erraneous exit, and the warnings are nothing but noise.

Issue #436, opened 2020-07-15
[RFC] Extending the linter interface to allow per-scope settings


whitequark:
Right now nMigen has an extremely simple linter that is also kind of hacky: it lets the first line of a file to contain a `# nmigen:` magic comment, and this comment allows disabling the `UnusedElaboratable` warning, mostly to allow creating elaboratables in tests just for their Python interface (i.e. without ever elaborating them). Issues like #404 would benefit from having a more well-defined linter interface.

This interface should allow:
  * enabling/disabling lints in file scope,
  * enabling/disabling lints in a scope delimited by the `with` statement,
  * avoid issues caused by globals in [Python's `warnings.catchwarnings`](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).

Although not required, I think that it might be nice to copy Rust's `#[allow]`/`#[warn]`/`#[deny]` system, since it seems to work well for them, and I expect that nMigen users will request a way to make lints into hard errors in the future.

programmerjake:
seems like a good idea except that it is not the standard Python way.

programmerjake:
an additional concern is that there should be an API to disable linter features without requiring the python source file of the caller to be accessible, this is needed if using nmigen from some other programming language through CPython's C API.

whitequark:
While in principle still feasible, the direct motivation for this feature is gone now that #1059 obviates the need for configuring that particular warning.

I am going to close this issue now. This isn't to imply rejection of the proposal, but simply an indication that it should go through our new RFC process now (described at https://amaranth-lang.org/rfcs/).

PR #1621, opened 2025-07-30
Modify MustUse to iterate the stack frame and collate options from all frames.

robtaylor:
Options in closer frames take precedence


amaranth/_utils.py
```diff
@@ -74,18 +74,39 @@ def decorator_like(*args, **kwargs):
...
         return decorator_like
 
+_LINT_PATTERN = re.compile(r"^#\s*amaranth:\s*((?:\w+=\w+\s*)(?:,\s*\w+=\w+\s*)*)\n$")
+
 
 def get_linter_options(filename):
     first_line = linecache.getline(filename, 1)
     if first_line:
-        match = re.match(r"^#\s*amaranth:\s*((?:\w+=\w+\s*)(?:,\s*\w+=\w+\s*)*)\n$", first_line)
+        match = _LINT_PATTERN.match(first_line)
```
whitequark, line 83: Since we're using match groups I would rather keep the expression within the function. (The Python `re` module has an LRU cache for expressions, so under any normal circumstances it will be compiled only once anyway.)

amaranth/_utils.py
```diff
@@ -74,18 +74,39 @@ def decorator_like(*args, **kwargs):
...
     return dict()
 
 
-def get_linter_option(filename, name, type, default):
-    options = get_linter_options(filename)
+def get_linter_option(frame, name, type, default):
+    """
+    Get given linter option for a given stack frame. This iterates down the frames, collating options of the form:
+    ```
+    # amaranth: {name} = value
```
whitequark, line 93: This isn't the right syntax: we don't allow spaces around `=`.

amaranth/_utils.py
```diff
@@ -74,18 +74,39 @@ def decorator_like(*args, **kwargs):
...
+    ```
+    The earliest option value in the stack takes precedence
+
+    Returns:
+        option value: bool | int
```
whitequark, line 98: The syntax should be like

```
Returns
-------
:class:`bool` or :class:`int`
    Option value
```

amaranth/_unused.py
```diff
@@ -1,5 +1,6 @@
 import sys
 import warnings
+import logging
```
whitequark, line 3: I think this isn't used anywhere?

whitequark (commented):
I'm undecided on this change overall. I agree that a better way to manage the noise from `UnusedElaboratable` is needed, but I'm not sure this is the best approach.

The main problem that gives me pause is that the semantics is so vaguely defined. Why does the outermost frame take priority, for example? Why should we not just follow the `super()` call chain in the constructor instead of going through the entire callstack?

I don't want to let this PR rot but I do want at least some discussion of the exact semantics to happen first. People will come to rely on it, and changing it later will become difficult.

amaranth/_utils.py
robtaylor: good to know! fixed

amaranth/_utils.py
robtaylor: fixed

amaranth/_unused.py
robtaylor: fixed

robtaylor:
> I'm undecided on this change overall. I agree that a better way to manage the noise from `UnusedElaboratable` is needed, but I'm not sure this is the best approach.
> 
> The main problem that gives me pause is that the semantics is so vaguely defined. Why does the outermost frame take priority, for example? 

The reasoning here is that it would be the user's code closest to the error that would control the output, which seemed the most logical choice at the time. but it may indeed make more sense that its the 'top level' that controls it.

> Why should we not just follow the `super()` call chain in the constructor instead of going through the entire callstack?
> 
> I don't want to let this PR rot but I do want at least some discussion of the exact semantics to happen first. People will come to rely on it, and changing it later will become difficult.

agreed. Its not obvious what the best policy is. That said, I think on reflection that following the call chain of the constructor could be confusing as the 'point' where a user gets the list of warnings may be quite separate in space and time from any given amaranth object.

whitequark:
By the way, regarding the "earlier/later" discussion: one caveat is that the same file can appear in the traceback multiple times, potentially in between two appearances of some other file; this is something that will be particularly acute if you use decorators.

robtaylor:
> At the time I implemented the existing code, I was concerned about the overhead of using `linecache`, since it takes a noticeable amount of time to grab the source code. This is exponentially worse with your PR since it will do this for every frame in the call stack, which is potentially a very large number. The lines do get cached at least.
> 
> A compromise could be to turn a traceback into a list of `co_filename` values, which are then turned into options later. I think this does the minimal amount of work possible.

I was thinking this too, so I measured it - surprisingly it makes little difference over the amaranth test suite, i guess as most of the sources are in local cache. 

<google-sheets-html-origin><style type="text/css"></style>
user (s) | system (s) | cpu % | total
-- | -- | -- | --
28.58 | 3.65 | 127% | 25.288
29.06 | 3.84 | 128% | 25.589
28.89 | 3.73 | 127% | 25.512
29.24 | 3.86 | 127% | 25.997
29.4 | 3.77 | 127% | 25.993
29.034 | 3.77 | 127% | 25.6758
  |   |   |  
  |   |   |  
28.28 | 3.77 | 127% | 25.137
28.64 | 3.92 | 127% | 25.511
28.56 | 3.95 | 127% | 25.446
28.84 | 3.74 | 127% | 25.465
29.11 | 3.72 | 127% | 25.822
28.686 | 3.82 | 127% | 25.4762

robtaylor:
@whitequark what's next steps here?

robtaylor:
For my reference:
Relevant open issues are:  
 * #1382
 * #924 
 
Related closed issues:
 *  #848
 *  #436

robtaylor:
Another thing that occurred to me - We currently only show these on an exception, though arguably this is when they are least useful. It seems to me that it would be more useful to default to displaying after *successful* elaboration.

whitequark:
> @whitequark what's next steps here?

Apologies for the delay, let's just merge it. I'm a little uncertain if this is the right direction to go in, but if it turns out to cause issues we can just modify the code again.

PR #1623, opened 2025-07-31
lib.data.{Field, Layout, Const}: add __hash__

rroohhh:


amaranth/lib/data.py
```diff
@@ -166,6 +169,9 @@ def as_shape(self):
         """
         return unsigned(self.size)
 
+    def __hash__(self):
+        return hash(self.size)
```
whitequark, line 173: This doesn't feel like a reasonable implementation.

whitequark (commented):
You should also add tests that these types are hashable.

amaranth/lib/data.py
rroohhh: I guess what is reasonable depends on usage, one could add in a hash of `frozenset(iter(self))`, but that is quite slow. I assumed that most uses of `__hash__` would be having a `dict` with `Layout`s of different sizes, but maybe that is a bad assumption?

whitequark: Yes, precomputing the hash on creation.

zyp: What about just putting `@functools.cache` on it?

whitequark: Also works, yeah!

whitequark (approved):

whitequark:
Could you update the changelog?

whitequark:
Thanks!

whitequark:
Sorry, didn't notice it was waiting on me still.

PR #1627, opened 2025-09-04
vendor._gowin: use braces around net names in SDC

agrif:
Gowin's SDC parser doesn't like quotes, so use braces instead.

This is a follow-up to PR #1501, merged as 0f27429. That PR just missed this one last use of `tcl_quote`.

Actually *using* signal constraints is somewhat fragile, as Gowin's tools don't seem to be listening to the `(* keep = "true" *)` attributes, but at least now there's no syntax error in the SDC files.


whitequark (approved):
LGTM, thanks!

PR #1642, opened 2025-12-03
docs: add a version switch.

miyukoc:
This patch makes the theme include a JavaScript file that replaces the static version HTML with a `<select>` element. The list of versions is fetched from `<content root>/../versions.json` and is expected to adhere to the following format:

```json
[
  { "name": "1.2.3", "root_url": "/docs/amaranth/v1.2.3/" },
  { "name": "1.2.2", "root_url": "/docs/amaranth/v1.2.2/" }
]
```

This file is expected to be updated manually similarly to how updates are done for https://amaranth-lang.org/index.html. However, an automation could be built at a maintainer's request: for example, this file could be assembled anew when pushing from this repo's GitHub Actions workflow to https://github.com/amaranth-lang/amaranth-lang.github.io.

The appearance of the version switch matches that of the Read the Docs theme's version switch.

image

Once this PR is merged, a follow-up PR should be created for https://github.com/amaranth-lang/amaranth-lang.github.io that adds this JavaScript file to all past versions of the documentation, together with the initial version of the `versions.json` file.

### DIFF
--- a/docs/_static/version-switch.js
+++ b/docs/_static/version-switch.js
@@ -1,0 +1,111 @@
+document.addEventListener('DOMContentLoaded', () => {
+  let contentRoot = new URL(
+    document.documentElement.dataset.content_root ?? DOCUMENTATION_OPTIONS.URL_ROOT,
+    window.location.href,
+  );
+
+  function insertVersionSwitch(versions) {
+    let versionElement = document.querySelector('.wy-side-nav-search > .version');
+    if (!versionElement) return;
+
+    let root = document.createElement('div');
+    root.innerHTML = `
+      <div class="switch-menus">
+        <div class="version-switch"><select></select></div>
+      </div>
+    `;
+
+    let switchMenus = root.firstElementChild;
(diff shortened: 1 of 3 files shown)


whitequark:
> However, an automation could be built at a maintainer's request: for example, this file could be assembled anew when pushing from this repo's GitHub Actions workflow to https://github.com/amaranth-lang/amaranth-lang.github.io.

Yes please.

docs/_static/version-switch.js
```diff
@@ -0,0 +1,111 @@
...
+    versionSwitchSelect.addEventListener('change', (event) => {
+      let option = event.target.selectedIndex;
+      let item = event.target.options[option];
+      window.location.href = item.dataset.url;
+    });
+
+    versionElement.replaceWith(switchMenus);
+  }
+
+  fetch(new URL('../versions.json', contentRoot)).then(async (response) => {
```
whitequark: Does this work when you're in a subdirectory, like https://amaranth-lang.org/docs/amaranth/latest/stdlib/memory.html ?

PR #1661, opened 2026-04-09
build.plat: Skip ports with empty metadata

jeanthom:
Skip yielding ports with None metadata. This change fixes a bug which prevents building a design using IOPort instances with no metadata (required by some Gowin IP cores).


whitequark (approved):
Looks good, let me keep the PR open a little to make sure I'm not missing anything.

Will you want a backport?

jeanthom:
@whitequark I don't need it at the moment, but let me know if you'd like me to backport it to a previous version.

PR #1675, opened 2026-08-12
vendor._gowin: (Apicula) add -family option to synth_gowin call

### LINKED ISSUE #1673: (Gowin) Apicula template generates call to `synth_gowin` without `-family` option, which defaults to gw1a family
Was looking at build logs for curiosity, and noticed it was using GW1N family extra cells during synthesis:

```
2.2. Executing Verilog-2005 frontend: /Users/joel/oss-cad-suite-20260805/libexec/../share/yosys/gowin/cells_xtra_gw1n.v
```

Figured I'd poke around in there see what else is available, and there's `cells_xtra_gw2a.v` which targets the GW2A-LV18PG256C8/I7 FPGA I'm using. I'm not using any specific functionality out of any of these extra family-specific cells (yet) so it hasn't bothered me.

Looking at yosys docs (ref: https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_techlibs_gowin.html), the `-family <family>` switch determines which family-specific cells get used. Docs state:

```
sets the gowin family to the specified value. The default is 'gw1n'.
The following families are supported:
'gw1n', 'gw2a', 'gw5a'.
```

So this should probably be set correctly for Gowin parts in `amaranth/vendor/_gowin.py`

In theory it's an easy fix:

```
--- a/amaranth/vendor/_gowin.py
+++ b/amaranth/vendor/_gowin.py
@@ -387,7 +387,7 @@ class GowinPlatform(TemplatedPlatform):
             {% endfor %}
             read_rtlil {{name}}.il
             {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
-            synth_gowin {{get_override("synth_opts")|options}} -top {{name}} -json {{name}}.syn.json
+            synth_gowin {{get_override("synth_opts")|options}} -family {{platform.series[:4].lower()}} -top {{name}} -json {{name}}.syn.json
             {{get_override("script_after_synth")|default("# (script_after_synth placeholder)")}}
         """,
     }
```

Because `platform.series` is always a part extracted out of the part number, it should always be available and be set to something vaguely sensible that Apicula supports.

This may be an issue for yosys synth_gowin old enough to not have the `-family` switch or separate gw2a libraries, but given the major changes in Amaranth 0.6 for Gowin parts anyway I don't think this will be a big deal because everyone will need a reasonably up-to-date yosys to even have `nextpnr-himbaechel`

Happy to put together a PR for this.

joelpmichael:
Fixes #1673. Per observations in issue report, this may require an updated yosys which understands the `-family` option to `synth_gowin`, but Amaranth 0.6.0 will require a yosys that's new enough to contain `nextpnr-himbaechel` anyway. Given this concern, this PR should not be back-ported to 0.5.x or earlier.

This will only affect code that targets Gowin family-specific cells that differ across Gowin families (GW1N, GW2A, GW5A), of which there are none within Amaranth-lang itself, or in the Gowin-based boards in Amaranth-boards that I have access to.

Tested on Sipeed Tang Primer 20k + Dock hardware, based on GW2A-LV18PG256C8/I7

### DIFF
--- a/amaranth/vendor/_gowin.py
+++ b/amaranth/vendor/_gowin.py
@@ -387,7 +387,7 @@ class GowinPlatform(TemplatedPlatform):
             {% endfor %}
             read_rtlil {{name}}.il
             {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
-            synth_gowin {{get_override("synth_opts")|options}} -top {{name}} -json {{name}}.syn.json
+            synth_gowin {{get_override("synth_opts")|options}} -family {{platform.series[:4].lower()}} -top {{name}} -json {{name}}.syn.json
             {{get_override("script_after_synth")|default("# (script_after_synth placeholder)")}}
         """,
     }


whitequark (approved):
Thank you!

PR #1676, opened 2026-08-12
Add Value.__iter__

### LINKED ISSUE #1674: Iterable Values
Values are technically iterable:

```python
>>> x = Signal(5)
>>> list(x)
[(slice (sig x) 0:1), (slice (sig x) 1:2), (slice (sig x) 2:3), (slice (sig x) 3:4), (slice (sig x) 4:5)]
```

And this is pretty useful, for example you can zip signals together, or signals with related metadata:

```python
>>> list(zip(x, range(5)))
[((slice (sig x) 0:1), 0), ((slice (sig x) 1:2), 1), ((slice (sig x) 2:3), 2), ((slice (sig x) 3:4), 3), ((slice (sig x) 4:5), 4)]
```

Unfortunately, this is just conventional, and [values are in principle not Iterables](https://docs.python.org/3/library/collections.abc.html#id5). One consequence is, among others, that such uses of `zip` do not pass checking with `pyright` and other type checkers. It is possible to add a "virtual" `__iter__` method just to the type stub just to make the static checks pass - but this does not help with dynamic checks via `isinstance`, and also accessing `__iter__` directly will lead to runtime errors.

tilk:
Closes #1674.

### DIFF
--- a/amaranth/_utils.py
+++ b/amaranth/_utils.py
@@ -23,8 +23,9 @@ def to_binary(n: int, width: int) -> str:
 
 
 def flatten(i):
+    from amaranth.hdl._ast import Value
     for e in i:
-        if isinstance(e, str) or not isinstance(e, Iterable):
+        if isinstance(e, (str, Value)) or not isinstance(e, Iterable):
             yield e
         else:
             yield from flatten(e)
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -1182,6 +1182,17 @@ class Value(metaclass=ABCMeta):
         else:
             raise TypeError(f"Cannot index value with {key!r}")
 
+    def __iter__(self):
+        """Iteration over bits, from least to most significant.
+
+        Yields
+        -------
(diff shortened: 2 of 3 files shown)


whitequark:
Now that I see that this fails a lot of tests, I recall that @wanda-phi wrote basically this exact patch and had to backtrack on it because of how many tests it failed.

Here are some thoughts:
- Accessing `__iter__` directly shouldn't really be a big issue since `iter(x)` works fine and that's the "right" way to do it per Python docs.
- How about just inheriting it from `Iterable`? Would that work?
- I'm not sure what the root cause is.

tilk:
This is weird, will investigate later.

whitequark:
Thank you!

tilk:
Turns out making `Value` an `Iterable` makes it another of these weird `Iterable` instances (like `str`) for which iteration is self-referential, which broke `amaranth._utils.flatten`.

whitequark:
Thanks. If you squash the commits I'll hit merge!

whitequark (approved):

whitequark:
Thanks!

PR #1678, opened 2026-08-21
Add super().__init_subclass__ call to ShapeCastable and ValueCastable

### LINKED ISSUE #1677: __init_subclass__ in ValueCastable and ShapeCastable does not call super().__init_subclass__
If `super().__init_subclass__` is not called, then `__init_subclass__` is not called on other base classes when doing multiple inheritance:

```python
class Foo(ShapeCastable, OtherClassWithInitSubclass):
    ...
```

This happens naturally when using `Generic` from `typing` - `__init_subclass__` is used internally. The following does not work because of the issue:

```python
class Foo[T](ShapeCastable):
    ...
```

tilk:
Closes #1677.

### DIFF
--- a/amaranth/hdl/_ast.py
+++ b/amaranth/hdl/_ast.py
@@ -255,6 +255,7 @@ class ShapeCastable:
         if cls.from_bits is ShapeCastable.from_bits:
             raise TypeError(f"Class '{cls.__qualname__}' deriving from 'ShapeCastable' must override "
                             f"the 'from_bits' method")
+        super().__init_subclass__(**kwargs)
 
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Thank you!

tilk:
Is this fix going to be backported to 0.5.x?

whitequark:
Would you like it to be? (Backports are done purely on demand.)

tilk:
Yes.

whitequark:
Could you send a PR to the v0.5.x branch please? If you do that, I can quickly cut a release.

PR #1680, opened 2026-09-04
docs: stdlib.stream: Fix bit values in serial receiver example

neuschaefer:
The example code uses the bit pattern 0b1010'0111.


whitequark (approved):
Thanks!

PR #1681, opened 2026-09-09
Add super().__init_subclass__ call to ShapeCastable and ValueCastable

tilk:
Backport of #1678 to v0.5.x.


whitequark (approved):
Thank you!

whitequark:
It looks like we have to drop the Python 3.8 CI job first.

whitequark:
#1682 will do it, hopefully

PR #1685, opened 2026-09-17
Fix typo in guide

diondokter:
Had some trouble parsing what was going on.
But I think `addr` should be the `bus_addr` that was created just above it


whitequark (approved):
Thank you!

